PR #129, opened 2014-07-11
Add basic inbound/outbound firewall, and a customised one for web hosts

GregSutcliffe:
New firewall, for review. Don't apply this to hosts yet :)

Current setup:

default DENY in and out
DNS & NTP allowed NEW outbound
SSH allowed NEW/ESTABLISHED inbound
SSH allowed ESTABLISHED only oubound
Puppet(8140) allowed NEW outbound

The web class adds https/rsync ports for yum and rsyncd

We can extend this further to the other hosts as time allows for testing what ports are required (eg I can imagine ci.theforeman.org will need outbound SSH state:NEW for Jenkins)

Thoughts?

### DIFF
--- a/puppet/modules/rbot/templates/redmine_urls.rb.erb
+++ b/puppet/modules/rbot/templates/redmine_urls.rb.erb
@@ -155,11 +155,11 @@ class RedmineUrlsPlugin < Plugin
   # if the channel isn't in the channelmap.
   #
   def base_url(target)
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
This has been open forever, and is cluttering up my shiny new review dashboard. Happy to re-open if there's future interest, but clsoing for now.

PR #151, opened 2015-02-04
adding QT to jenkins server

unorthodoxgeek:
in order to run a headless browser, we need libQT on the CI.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -114,6 +114,12 @@ class slave($github_user = undef,
       name => $osfamily ? {
         Debian => "libaugeas-dev",
         default => "augeas-devel"
+      };
+    "lib-qt":
+      ensure => present,
+      name => $osfamily ? {
+        Debian => "libqt4-dev",
+        default => "libqt4-devel"
       }
   }
 


puppet/modules/slave/manifests/init.pp
```diff
@@ -114,6 +114,12 @@
       name => $osfamily ? {
         Debian => "libaugeas-dev",
         default => "augeas-devel"
+      };
+    "lib-qt":
+      ensure => present,
+      name => $osfamily ? {
+        Debian => "libqt4-dev",
+        default => "libqt4-devel"
```
domcleal: This package doesn't exist on EL6.

domcleal: "capybara-webkit requires Qt version 4.8 or greater" according to the project's README too, so I don't think that EL6 satisfies this, it appears to have 4.6.

unorthodoxgeek: the CI runs on EL6?

if so, how can we solve this? compile from source?

domcleal: It's a mix of EL6 and Debian hosts.  I don't particularly want to have software installed from source on our servers, and it'd take a lot of effort to implement.

I don't think you can rely on that gem until we can rebuild the environment onto EL7.

domcleal: Or is there an older version that has support for 4.6?

PR #152, opened 2015-02-05
Initial setup/use of jenkins-job-builder

GregSutcliffe:
Installs JJB from PiP
Deploys config file
Deploys Job YAML from puppet
Calls 'jjb test' and redirects to /tmp for evaluation

Currently just implements deploy_web as a simple case, and the deb half of the proxy pipeline. You'll need to add the view yourself once the jobs are deployed (if you build a test jenkins which doesn't have the view defined). There is an analogous jenkins-view-builder, but it seems buggy, so I'll leave that for a future PR.

You can apply it via
```
class { 'jenkins_job_builder':
  url      => 'http://192.168.122.2:8080',
  username => 'user',
  password => 'changeme',
  config_dir => 'puppet:///modules/jenkins_job_builder/config'
}
```
Clearly the user/pw are a user with write access to create/update jobs. Installing 1.1.0 from PiP seems sufficient for now, so I dropped the VCS install from the Openstack version of the module. Chance are we'll need to revert to VCS mode and fork JJB so we can add the modules we need to configure (eg, the scheduled rebuild plugin) but this gets us started.

Likely to be a lot of non-DRY yaml in here, as I was just implementing jobs one at a time. Fixes welcome, or we can do that in another PR.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
+++ b/puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+ruby=2.0.0
+# Clean gemset and database
+. /etc/profile.d/rvm.sh
+gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
+rvm use ruby-${ruby}@${gemset}
+rvm gemset delete ${gemset} --force
+exit 0
--- a/puppet/modules/jenkins_job_builder/files/config/yaml/jobs/test_proxy_develop.yaml
+++ b/puppet/modules/jenkins_job_builder/files/config/yaml/jobs/test_proxy_develop.yaml
@@ -1,0 +1,45 @@
+- job:
+    name: test_proxy_develop
+    project-type: matrix
+    properties:
+      - github:
+          url: https://github.com/theforeman/smart-proxy
+    scm:
+      - foreman-proxy
+    triggers:
+      - github
+      - pollscm: '*/5 * * * *'
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+          - 1.8.7
+          - 1.9.2
+          - 1.9.3
+          - 2.0.0
+          - 2.1
+      - axis:
+          type: user-defined
+          name: puppet
+          values:
+          - 2.6.0
+          - 2.7.0
+          - 3.0.0
+          - 3.4.0
+          - 3.7.0
+    execution-strategy:
+      combination-filter: '!( (ruby ==~ /(1\.9|2).*/ && puppet ==~ /2\.6.*/) || (ruby ==~ /2.*/ && puppet ==~ /(2|3\.[01]).*/) || (ruby ==~ /2.[^0]*/ && puppet ==~ /(2|3\.[0-4]).*/) )'
+    builders:
+      - test_proxy
+    publishers:
+    - gemset_cleanup_proxy
+    - archive:
+        artifacts: 'pkg/*'
+    - trigger-parameterized-builds:
+      - project: release_nightly_build_rpm, release_nightly_build_deb
+        condition: SUCCESS
+        predefined-parameters: |
+          project=foreman-proxy
+          jenkins_job=${JOB_NAME}
+          jenkins_job_id=${BUILD_NUMBER}
--- a/puppet/modules/jenkins_job_builder/files/config/yaml/publishers/ircbot_freenode.yaml
+++ b/puppet/modules/jenkins_job_builder/files/config/yaml/publishers/ircbot_freenode.yaml
@@ -1,0 +1,15 @@
+- publisher:
+    name: ircbot_freenode
+    publishers:
+    - ircbot:
+        strategy: failure-and-fixed
+        notify-start: false
+        notify-committers: false
+        notify-culprits: false
+        notify-upstream: false
+        notify-fixers: false
+        message-type: summary
+        channels:
+            - name: '#theforeman-dev'
+              notify-only: true
+        matrix-notifier: only-configurations
--- a/puppet/modules/jenkins_job_builder/manifests/init.pp
+++ b/puppet/modules/jenkins_job_builder/manifests/init.pp
@@ -1,0 +1,68 @@
+# Class to install, configure, and maintain JJB, as well
+# as to deploy the actual jobs to jenkins
+#
+# Loosely based on
+# https://git.openstack.org/cgit/openstack-infra/puppet-jenkins/tree/manifests/job_builder.pp
+# and should probably be kept up to date from there
+#
+# $run is our addition and is so named becuase noop is a bit too magic in puppet syntax...
+#
+class jenkins_job_builder (
+  $url = '',
+  $username = '',
+  $password = '',
+  $config_dir = '',
+  $run = 'false',
+) {
+
+  if ! defined(Package['jenkins-job-builder']) {
+    package { 'jenkins-job-builder':
+      ensure   => present,
+      provider => 'pip',
+    }
+  }
+
+  file { '/etc/jenkins_jobs':
+    ensure => directory,
+  }
+
+  file { '/etc/jenkins_jobs/config':
+    ensure  => directory,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    recurse => true,
+    purge   => true,
+    force   => true,
+    source  => $config_dir,
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  # test for a string here since it's annoyingly hard to pass a boolean from foreman via yaml
+  if $run == 'false' {
+    $cmd = "jenkins-jobs test /etc/jenkins_jobs/config > /var/cache/jjb.xml"
+  }else{
+    $cmd = "jenkins-jobs update /etc/jenkins_jobs/config > /var/cache/jjb.xml"
+    # eventually we may wish to nuke unmanaged jobs:
+    #$cmd = "jenkins-jobs update --delete-old /etc/jenkins_jobs/config > /var/cache/jjb.xml"
+  }
+
+  exec { 'jenkins_jobs_update':
+    command     => $cmd,
+    timeout     => '600',
+    path        => '/bin:/usr/bin:/usr/local/bin',
+    refreshonly => true,
+    require     => [
+      File['/etc/jenkins_jobs/jenkins_jobs.ini'],
+      Package['jenkins-job-builder'],
+    ],
+  }
+
+# TODO: We should put in  notify Exec['jenkins_jobs_update']
+#       at some point, but that still has some problems.
+  file { '/etc/jenkins_jobs/jenkins_jobs.ini':
+    ensure  => present,
+    mode    => '0400',
+    content => template('jenkins_job_builder/jenkins_jobs.ini.erb'),
+  }
+}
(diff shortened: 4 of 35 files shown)


puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+    force   => true,
+    source  => $config_dir,
+    require => File['/etc/jenkins_jobs'],
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  exec { 'jenkins_jobs_update':
+    #command    => 'jenkins-jobs update --delete-old /etc/jenkins_jobs/config', #danger!
+    #command    => 'jenkins-jobs update /etc/jenkins_jobs/config',              #op
+    command     => 'jenkins-jobs test /etc/jenkins_jobs/config > /tmp/jjb.xml', #no-op
```
domcleal: Maybe use logoutput instead, or as well as?

Using > will probably need provider => shell.

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+  file { '/etc/jenkins_jobs/config':
+    ensure  => directory,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    recurse => true,
+    purge   => true,
+    force   => true,
+    source  => $config_dir,
+    require => File['/etc/jenkins_jobs'],
```
domcleal: superfluous

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+    notify  => Exec['jenkins_jobs_update'],
```
domcleal: `~>`

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+    ],
+  }
+
+# TODO: We should put in  notify Exec['jenkins_jobs_update']
+#       at some point, but that still has some problems.
+  file { '/etc/jenkins_jobs/jenkins_jobs.ini':
+    ensure  => present,
+    mode    => '0400',
+    content => template('jenkins_job_builder/jenkins_jobs.ini.erb'),
+    require => File['/etc/jenkins_jobs'],
```
domcleal: superfluous

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+    path        => '/bin:/usr/bin:/usr/local/bin',
+    refreshonly => true,
+    require     => [
+      File['/etc/jenkins_jobs/jenkins_jobs.ini'],
+      Package['jenkins-job-builder'],
+    ],
+  }
+
+# TODO: We should put in  notify Exec['jenkins_jobs_update']
+#       at some point, but that still has some problems.
```
domcleal: What do you mean exactly?

puppet/modules/jenkins_job_builder/files/config/yaml/jobs/test_proxy_develop.yaml
```diff
@@ -0,0 +1,44 @@
...
+          - 2.6.0
+          - 2.7.0
+          - 3.0.0
+          - 3.4.0
+          - 3.7.0
+    execution-strategy:
+      combination-filter: '!( (ruby ==~ /(1\.9|2).*/ && puppet ==~ /2\.6.*/) || (ruby ==~ /2.*/ && puppet ==~ /(2|3\.[01]).*/) || (ruby ==~ /2.[^0]*/ && puppet ==~ /(2|3\.[0-4]).*/) )'
+    builders:
+      - test_proxy
+    publishers:
```
domcleal: missing gemset cleanup?

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+ruby=2.0.0
```
domcleal: Be a little careful here, the jobs are subtly different.  Can it be parameterised, or do we create multiple publishers for each variant?

1. Some jobs that use Ruby, e.g. deploy_web, just have a hardcoded version of Ruby here.  Most use Ruby version as an axis so `${ruby}` is already set.
2. test_develop etc will also call rake db:drop
3. Plugin jobs will also cd to the Foreman checkout, then db:drop
4. I've forgotten the fourth.
2.

puppet/modules/jenkins_job_builder/files/config/yaml/publishers/ircbot_freenode.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - ircbot:
+        strategy: failure-and-fixed
+        notify-start: false
+        notify-committers: false
+        notify-culprits: false
+        notify-upstream: false
+        notify-fixers: false
+        message-type: summary
+        channels:
+            - name: '##gwmngilfen'
```
domcleal: update please

domcleal:
Does JJB restart Jenkins in some way, or will Jenkins pick up on the config file changes?

I think I've changed some of these jobs slightly since you wrote this, but not to worry, we'll pick that up from no-op.

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,60 @@
...
+    purge   => true,
+    force   => true,
+    source  => $config_dir,
+    require => File['/etc/jenkins_jobs'],
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  exec { 'jenkins_jobs_update':
+    #command    => 'jenkins-jobs update --delete-old /etc/jenkins_jobs/config', #danger!
+    #command    => 'jenkins-jobs update /etc/jenkins_jobs/config',              #op
```
domcleal: How about parameterising this?  Then we could enable/disable it from Foreman, or even have a staging-jenkins server which is always up to date or something.

puppet/modules/jenkins_job_builder/manifests/init.pp
GregSutcliffe: Thats copied from the openstack code. I left it in so I could see when they fixed it, if we ever update our code from theirs...

puppet/modules/jenkins_job_builder/files/config/yaml/jobs/test_proxy_develop.yaml
GregSutcliffe: Hmm, must have missed that. Will add.

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
GregSutcliffe: I think it *can* be made sufficiently flexible, but for our first pass we may just want to use multiple publishers. We can then verify the xml doesn't change (or changes suitably) when we refactor.

puppet/modules/jenkins_job_builder/files/config/yaml/publishers/ircbot_freenode.yaml
GregSutcliffe: Hah, oops. Good catch :)

puppet/modules/jenkins_job_builder/manifests/init.pp
GregSutcliffe: Good idea, will implement that today.

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
domcleal: Yeah, multiple's fine, so long as they're all there.

puppet/modules/jenkins_job_builder/manifests/init.pp
GregSutcliffe: There's some extra stuff in between this and the exec now, so I'll leave this here...

puppet/modules/jenkins_job_builder/files/config/yaml/jobs/test_proxy_develop.yaml
GregSutcliffe: [x] Done

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
GregSutcliffe: I've added a proxy one for now. The ruby thing we can handle with something like "localruby=${ruby:=2.0.0}". We could rationalize the way we call rvm which seems different between the two, but possibly not for good reasons. The rake tasks could be conditional'ed in some way....

It seems doable anyway :)

GregSutcliffe:
@domcleal updated. As to your comment, yeah, it's API driven. JJB generates *all* the required XML on each run, and fires it into the API, regardless of the current state of those jobs. Since it's API driven, Jenkins should properly handle not breaking jobs which are currently running, and then switching over for new executions. Unmanaged jobs will be unaffected unless we switch on the nuke option ;)

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
```diff
@@ -0,0 +1,6 @@
...
+# Clean gemset and database
+. /etc/profile.d/rvm.sh
+gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
+rvm use ruby-${ruby}@${gemset}
```
domcleal: missing the gemset delete?

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
domcleal: Yeah, just setting 'ruby' if it's unset would work.  I only see a difference in using `@gemset`, which could be safely removed.

puppet/modules/jenkins_job_builder/files/config/scripts/gemset_cleanup.sh
GregSutcliffe: Hmm, i must have only copied the visible lines in the textarea. Fixed.

puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -0,0 +1,68 @@
...
+    mode    => '0755',
+    recurse => true,
+    purge   => true,
+    force   => true,
+    source  => $config_dir,
+    notify  => Exec['jenkins_jobs_update'],
+  }
+
+  # test for a string here since it's annoyingly hard to pass a boolean from foreman via yaml
+  if $norun == 'false' {
```
domcleal: Aren't these (test/update) the wrong way around?  Or am I mis-reading the double negative?

GregSutcliffe: ugh, i should know better than to use double negatives like that. fixed.

domcleal:
Thanks @GregSutcliffe!

Note that I have changed some of the jobs in the Jenkins UI since you originally wrote this, so you may have to go through the no-op mode a little and update, sorry...

GregSutcliffe:
That's fine. My initial task is to get JJB installed and running on Jenkins, which this should achieve, with some luck.

PR #154, opened 2015-02-17
Updated mysql, postgresql + deps to latest versions for EL7 support

domcleal:
Also fixed an implicit dependency on wget which isn't in the base install.

----

No idea if this is everything yet, but this is the easy part causing lots of red.


GregSutcliffe:
Good call on moving apt to the submodules, no idea why we didn't do that earlier. Merged, thanks!

PR #156, opened 2015-03-01
Add foreman-docker to cron task repos

daviddavis:


domcleal:
https://github.com/theforeman/redmine/blob/master/.openshift/cron/hourly/git_repos is the current file that's used.  However Docker is listed, and it's already in Redmine: http://projects.theforeman.org/projects/docker/repository

PR #157, opened 2015-03-12
Switch JJB to a git clone, for ease of coding new plugin xml changes

GregSutcliffe:
This is so we can use things like http://www.larrycaiyu.com/blog/2014/01/14/extend-jenkins-job-builder/ on our fork of JJB to add plugins we use in Jenkins.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -76,3 +76,6 @@
 [submodule "puppet/modules/apt"]
 	path = puppet/modules/apt
 	url = https://github.com/puppetlabs/puppetlabs-apt.git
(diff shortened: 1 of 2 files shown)


domcleal:
I take it some of the Puppet manifest styling comes from OpenStack's manifest, do you have a link to it still?

GregSutcliffe:
@domcleal yep, https://git.openstack.org/cgit/openstack-infra/puppet-jenkins/tree/manifests/job_builder.pp. I've tried to clean up a bit though.

domcleal:
Thanks!

PR #158, opened 2015-03-12
Update initial JJB job set from XML review

GregSutcliffe:
I left one concurrent-throttle example in there in case we decide we want it. Everything else should be expected.

@domcleal how does this relate to the [json in the slave class](https://github.com/theforeman/foreman-infra/commit/8b9c3ee0aaa7db1d70f6b37973a099e19e4439a6) interact with this stuff? Do we need to add a task to move that at some point?

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/config/yaml/publishers/systest_foreman.yaml
+++ b/puppet/modules/jenkins_job_builder/files/config/yaml/publishers/systest_foreman.yaml
@@ -2,8 +2,9 @@
     name: systest_foreman
     publishers:
     - archive:
-        artifacts: '*.tar.gz'
+        artifacts: 'debug/*,*.bats.out,*.tap'
         allow-empty: true
+        only-successful: true
     - tap:
         results: '*.bats.out'
         fail-if-no-results: true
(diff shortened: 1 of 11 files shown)


puppet/modules/jenkins_job_builder/files/config/yaml/publishers/systest_foreman.yaml
```diff
@@ -2,8 +2,9 @@
     name: systest_foreman
     publishers:
     - archive:
-        artifacts: '*.tar.gz'
+        artifacts: 'debug/*,*.bats.out,*.tap'
         allow-empty: true
+        only-successful: true
```
domcleal: false

domcleal:
@GregSutcliffe the JSON in that class is configuration for this script: https://github.com/theforeman/test-pull-requests

The script runs inside the pull_request_scanner Jenkins job and looks for PRs, submits jobs to Jenkins and updates PR results.  It just has to refer to the jobs we create here, so there isn't really any duplication.

(Eventually I will try and replace the script with Jenkins' own PR tools, but it requires a few job changes, which'll be much easier once we have JJB job templates.)

domcleal:
Merged as 85004ff with that fixup, thanks @GregSutcliffe.  Once deployed, I'll flip the parameter in Foreman.

PR #160, opened 2015-03-18
fixes #9813 - ensure slaves have latest qpid-cpp-client-devel

dustints:


puppet/modules/slave/manifests/init.pp
```diff
@@ -151,12 +151,12 @@
   if $osfamily == 'RedHat' {
     yumrepo { 'katello-pulp':
       descr    => "Katello Pulp Repo",
-      baseurl  => "http://fedorapeople.org/groups/katello/releases/yum/nightly/pulp/RHEL/\$releasever/\$basearch",
+      baseurl  => "http://koji.katello.org/releases/yum/katello-nightly/pulp/RHEL/\$releasever/\$basearch"
```
domcleal: syntax error, keep the comma

dustints: @domcleal updated

domcleal:
Thanks, it should be deployed within about 35 mins.

PR #162, opened 2015-03-27
[jjb] Fix formatting in katello.org SCM yaml

GregSutcliffe:
JJB doesn't like the bare / it seems:
```
yaml.scanner.ScannerError: while scanning an alias
in "/etc/jenkins_jobs/config/yaml/scm/katello.org.yaml", line 9, column 13
expected alphabetic or numeric character, but found '/'
in "/etc/jenkins_jobs/config/yaml/scm/katello.org.yaml", line 9, column 14
```


domcleal:
Thanks @GregSutcliffe!

PR #164, opened 2015-04-01
add Debian/jessie build support

mmoll:
And also some minor other things, please check esp. the change to wheezy in puppet/modules/freight/manifests/install.pp

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/config/yaml/jobs/packaging_build_deb_coreproject.yaml
+++ b/puppet/modules/jenkins_job_builder/files/config/yaml/jobs/packaging_build_deb_coreproject.yaml
@@ -20,8 +20,8 @@
           name: os
           values:
           - wheezy
+          - jessie
           - precise
-          - squeeze
           - trusty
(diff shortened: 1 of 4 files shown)


puppet/modules/jenkins_job_builder/files/config/yaml/jobs/packaging_build_deb_coreproject.yaml
```diff
@@ -20,8 +20,8 @@
           name: os
           values:
           - wheezy
+          - jessie
```
domcleal: Could you either revert this part until we have the sources in foreman-packaging, or change the combination filter to not build jessie by default?  (replace squeeze with jessie)

mmoll: sure, done

domcleal:
Thanks @mmoll!  It should roll out within 40 mins or so, I'll try and remember to check.

PR #169, opened 2015-04-23
don't use PL repo for Debian/jessie

mmoll:
The "jessie" repo was empty anyway (besides puppetlabs-release), and "testing"
repos were filled. However, now even puppetlabs-release was pulled from
apt.puppetlabs.com index, so fb-install-plpuppet.bats would always fail.


domcleal:
Thanks!

PR #170, opened 2015-05-31
Add openscap repo data and install openscap

shlomizadok:
What I'd like to achieve: __Installing OpenSCAP on the slave servers.__
How: Adding openscap repo data to the slave puppet.

~~Please note related PR: https://github.com/theforeman/puppet-foreman/pull/326
Which will require a version bump for foreman-infra/modules/foreman submodule.~~

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -162,6 +162,19 @@ class slave($github_user = undef,
     }
   }
 
+  # Needed by foreman_openscap gem dependency OpenSCAP
+  if $osfamily == 'RedHat' {
+    yumrepo { 'isimluk-openscap':
+      enabled     => 1,
+      gpgcheck    => 0,
+      baseurl     => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/epel-${::operatingsystemmajrelease}-\$basearch/",
+      includepkgs => ['openscap'],
+    } ->
+    package { 'openscap':
+      ensure => latest,
+    }
+  }
+
   # specs-from-koji
   if $osfamily == 'RedHat' {
     package {


domcleal:
I'm not sure that depending on one class from that module is a good idea, we don't intend it to be reused like that.  e.g. it contains a reference to Foreman::Plugin in the middle.  Probably better just to copy the repo definition here.

shlomizadok:
I am so happy you say that :)
I wanted not to DRY myself, though it seemed like a perfect situation were DRYing is acceptable :smile:

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
     }
   }
 
+  # Needed by foreman_openscap gem dependency OpenSCAP
+  if $osfamily == 'RedHat' {
+    $repo = $::operatingsystem ? {
+        'Fedora' => 'fedora',
```
domcleal: don't worry about Fedora, we don't deploy it

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
...
+        default  => 'epel',
+      }
+
+      if $::operatingsystemmajrelease == undef {
```
domcleal: don't worry about this, we have 1.7+ and can rely on it

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
...
+        $versions_array = split($::operatingsystemrelease, '\.') # facter 1.6
+        $major = $versions_array[0]
+      } else {
+        $major = $::operatingsystemmajrelease # facter 1.7+
+      }
+
+      yumrepo { 'isimluk-openscap':
```
domcleal: indentation's off

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
...
+        enabled  => 1,
+        gpgcheck => 0,
+        baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/${repo}-${major}-\$basearch/",
```
domcleal: could you also set `includepkgs => ['openscap']` so we don't pick up other stuff from this repo?

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
...
+      } ->
+    package { 'openscap':
```
domcleal: do we need -devel for compiling against?

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,30 @@
...
+      ensure => '1.2.3'
```
domcleal: trailing comma

domcleal:
Yeah, DRY is good, but this isn't part of a shipped project so it's less important, especially crossing the module boundary...

shlomizadok:
@domcleal, thank you so much for your comments. I hope it looks better now

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,19 @@
...
 
+  # Needed by foreman_openscap gem dependency OpenSCAP
+  if $osfamily == 'RedHat' {
+    yumrepo { 'isimluk-openscap':
+      enabled  => 1,
+      gpgcheck => 0,
+      baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/epel-${::operatingsystemmajrelease}-\$basearch/",
+      includepkgs => ['openscap'],
+    } ->
+    package { 'openscap':
```
domcleal: should this be openscap-devel if you're compiling against it (e.g. https://github.com/theforeman/foreman-packaging/blob/rpm/develop/rubygem-openscap/rubygem-openscap.spec#L35)?  I assume you're adding this in order to build the openscap gem?

ohadlevy: AFAIK no, as it needs to use FFI C functions.

domcleal: Ah yes, thanks.

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,19 @@
...
+      ensure => '1.2.3',
```
domcleal: This needs to be the full VR, so `1.2.3-0.1.el6` etc or just use `latest` (preferred).

puppet/modules/slave/manifests/init.pp
```diff
@@ -162,6 +162,19 @@
     }
   }
 
+  # Needed by foreman_openscap gem dependency OpenSCAP
+  if $osfamily == 'RedHat' {
+    yumrepo { 'isimluk-openscap':
+      enabled  => 1,
+      gpgcheck => 0,
+      baseurl  => "http://copr-be.cloud.fedoraproject.org/results/isimluk/OpenSCAP/epel-${::operatingsystemmajrelease}-\$basearch/",
```
domcleal: indentation's off on lines 168-170

domcleal:
Thanks @shlomizadok, should happen within the next 40 mins or so if successful.

By the way, please leave a comment when you update a PR to say what's fixed, else I'll only find out it changed by luck or persistence.

PR #171, opened 2015-06-04
Create jenkins job for testing fusor plugin

sdherr:
This pull request should (hopefully, I don't have any way of testing it...) create the infrastructure / job definition needed for us to run tests automatically in jenkins whenever people submit pull requests to the fusor project. Fusor is a plugin to foreman that allows you to easily deploy several Red Hat products.


domcleal:
We have standard jobs for testing plugins, it would be better if you could use those rather than create new variants.

I'd also suggest that you follow the notes here about how we support plugins: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Plugin#Making-your-plugin-official, since for us to provide CI tooling would require that the plugin is moved to our GitHub organisation and standardised in its layout and distribution.

PR #173, opened 2015-06-29
Add foreman_remote_execution plugin

stbenjam:
Just guessing how to do this based on how other plugins are setup. :sheep:


domcleal:
Thanks @stbenjam

domcleal:
See http://ci.theforeman.org/job/test_plugin_foreman_remote_execution_master/1/.

PR #175, opened 2015-07-15
Test foreman_remote_execution against 1.9-stable

stbenjam:


domcleal:
Do I understand this correctly, that you only want to test pull requests against 1.9-stable but keep the master tests against develop?

stbenjam:
No, sorry, I guess like this.

domcleal:
Thanks, it should take effect within 45 mins.

PR #176, opened 2015-07-21
Add puppet lookaside option to repoclosure script.

ehelms:


ehelms:
Forgot to add the parameter.

ehelms:
Updated with the parameter.

puppet/modules/jenkins_job_builder/files/config/scripts/packaging_repoclosure.sh
```diff
@@ -41,5 +41,15 @@ if [ -n "${copr_lookasides}" ]; then
   done
 fi
 
+puppet_lookaside=""
+if [ -n "${puppet_lookasides}" ]; then
+  lookaside_repos=$(echo $puppet_lookasides | tr "," "\n")
+
+  for lookaside_repo in $lookaside_repos
+  do
+    puppet_lookaside+="--repofrompath=${lookaside_repo},http://yum.puppetlabs.com/el/${osver}/products/x86_64/ -l ${lookaside_repo} "
```
domcleal: Could you perhaps put the el/ inside the string?  We use this repoclosure script for Fedora too, so would be good to have the option to use that repo.

ehelms:
Added the ${os} flag into the string.

domcleal:
${os} would be el6 or el7 etc, not "el" or "fedora"?

ehelms:
I updated the script to also include a dry run mode for local verification of the output which helped in testing via:

```
 puppet_lookasides=puppet/f20 os=f20 sh packaging_repoclosure.sh --dry-run
 puppet_lookasides=puppet/el6 os=el6 sh packaging_repoclosure.sh --dry-run
```

puppet/modules/jenkins_job_builder/files/config/scripts/packaging_repoclosure.sh
```diff
@@ -41,5 +41,25 @@ if [ -n "${copr_lookasides}" ]; then
...
-cd repoclosure
-./repoclosure.sh yum_${os}.conf http://koji.katello.org/releases/yum/${repo}/${osname}/${osver}/x86_64/ -l ${os}-base -l ${os}-updates -l ${os}-epel -l ${os}-scl -l ${os}-scl-ruby -l ${os}-scl-v8 -l ${os}-openscap ${koji_lookaside} ${foreman_lookaside} ${copr_lookaside}
+puppet_lookaside=""
+if [ -n "${puppet_lookasides}" ]; then
+  lookaside_repos=$(echo $puppet_lookasides | tr "," "\n")
+
+  for lookaside_repo in $lookaside_repos
+  do
+    if [ ${osname} == 'Fedora' ]; then
+      puppet_lookaside+="--repofrompath=${lookaside_repo},http://yum.puppetlabs.com/${osname}/${os}/products/x86_64/ -l ${lookaside_repo} "
```
domcleal: Lower case 'Fedora' here, perhaps hard code it instead of using osname?

ehelms:
Knew I missed something, updated to go the hard coded route.

domcleal:
Thanks!

PR #178, opened 2015-07-23
Add katello-packaging to pull request scanner

ehelms:


puppet/modules/slave/manifests/init.pp
```diff
@@ -34,6 +34,7 @@
     "hammer_cli_foreman_discovery",
     "katello",
     "katello_installer",
+    "katello-packaging",
```
domcleal: Use an underscore, it has to match the filename below.

puppet/modules/slave/templates/test_pull_requests_katello_packaging.json.erb
```diff
@@ -0,0 +1,43 @@
+{
+    "proxy_host": null,
+    "proxy_port": null,
+    "jenkins_host": "http://ci.theforeman.org/",
+    "jenkins_url": "http://ci.theforeman.org/",
+    "hub_config": "~/.config/hub",
+    "github_user": "katello",
```
domcleal: This *may* need to be uppercase to match the org name, but I'm not certain.

ehelms:
Updated - changed to uppercase and the dash to underscore

domcleal:
Thanks @ehelms.

PR #179, opened 2015-07-23
Add foreman_openscap plugin

shlomizadok:


domcleal:
Thanks @shlomizadok, let me know if this doesn't start working, should happen in a few hours.

PR #180, opened 2015-08-09
bump Ruby 2.1 and 2.2 versions

mmoll:
2.1.5 is coming with Debian/jessie and 2.2.2 is the lastest 2.2.

The rest is just to appease puppet-lint.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -4,127 +4,127 @@ class slave($github_user = undef,
             $koji_certificate = undef,
             $rackspace_username = undef,
             $rackspace_api_key = undef) {
-  file { "/var/lib/workspace":
+  file { '/var/lib/workspace':
     ensure => directory,
-    owner => "jenkins",
-    group => "jenkins"
+    owner  => 'jenkins',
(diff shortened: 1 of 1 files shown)


puppet/modules/slave/manifests/init.pp
```diff
@@ -193,33 +193,33 @@
...
     }
-    slave::rvm_config { "ruby-2.1":
-      version => "ruby-2.1.2",
+    slave::rvm_config { 'ruby-2.1':
+      version => 'ruby-2.1.5',
     }
-    slave::rvm_config { "ruby-2.2":
-      version => "ruby-2.2.1",
+    slave::rvm_config { 'ruby-2.2':
+      version => 'ruby-2.2.2',
```
domcleal: We're on 1.26.11 (the latest) at the moment, but it doesn't have 2.2.2 available.  (I suspect this might help https://github.com/theforeman/foreman/pull/2517 too, but we're a bit stuck.)

mmoll: I see, that's unexpected :astonished:
left 2.2 at 2.2.1 now

domcleal:
Thanks @mmoll.

PR #184, opened 2015-09-11
Change Debian mirrors to US for proximity to builders

domcleal:
This past week we've been seeing build failures on Debian OSes connecting to the .de mirror from our slaves which are all in the US.  142df89 already changed Ubuntu, so this changes Debian too.


mmoll:
:+1:

PR #186, opened 2015-09-23
Remove projects redirect and fix nudnik to handle it

GregSutcliffe:
I tried to test this with a VM, but the rbot code just will not run for me. I *think* it'll be fine, and we can fail-forward if need be :)

### DIFF
--- a/puppet/modules/web/templates/web.conf.erb
+++ b/puppet/modules/web/templates/web.conf.erb
@@ -5,7 +5,6 @@
 
   DocumentRoot /var/www/vhosts/web/htdocs
   RewriteEngine on
-  RewriteRule   ^/projects(.*)  http://projects.theforeman.org/projects$1  [R,L]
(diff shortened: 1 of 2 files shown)


puppet/modules/web/templates/web.conf.erb
```diff
@@ -5,7 +5,6 @@
 
   DocumentRoot /var/www/vhosts/web/htdocs
   RewriteEngine on
-  RewriteRule   ^/projects(.*)  http://projects.theforeman.org/projects$1  [R,L]
```
domcleal: I'm not sure we can remove this redirect, sorry.  I just checked the web server's access log and see we're still getting a lot of requests for these URLs by users from search engines.  I think the reason is that we're only serving a temporary, not permanent redirect.

(`grep "GET /projects" /var/log/httpd/web_access.log`)

GregSutcliffe:
@domcleal ok, i'll just renanme projects.html then

PR #187, opened 2015-09-24
ARM slave support

domcleal:
Some fixes from running our slave modules on two new ARM builders.

### DIFF
--- a/puppet/modules/utility/manifests/init.pp
+++ b/puppet/modules/utility/manifests/init.pp
@@ -28,10 +28,7 @@ class utility($sysadmins = ['/dev/null']) {
 
   package { "ruby-shadow":
     ensure => installed,
-    name => $osfamily ? {
-      Debian => "libshadow-ruby1.8",
-      default => "ruby-shadow"
-    }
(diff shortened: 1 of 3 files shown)


puppet/modules/utility/manifests/init.pp
```diff
@@ -28,10 +28,7 @@
 
   package { "ruby-shadow":
     ensure => installed,
-    name => $osfamily ? {
-      Debian => "libshadow-ruby1.8",
```
mmoll: This might still be needed on Ubuntu/precise

mmoll:
one note, :+1: otherwise

puppet/modules/utility/manifests/init.pp
domcleal: We've only got Wheezy slaves, and now one Ubuntu 14.10 (which I hope we can rebuild to 16.04 when it's out).

mmoll: perfect then :)

domcleal:
Thanks for the review!

PR #192, opened 2015-10-13
refs #12152: add ruby 2.2.2 to smart proxy test matrix

witlessbird:


domcleal:
You'll also need to update rvm on the slaves to make 2.2 available.  Do that under puppet/modules/slave/manifests/init.pp.

domcleal:
(And preferably 2.2.3, since it's out.)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_develop.yaml
```diff
@@ -23,6 +23,7 @@
           - 1.9.3
           - 2.0.0
           - 2.1
+          - 2.2.2
```
domcleal: We define aliases in slave/manifests/init.pp, so you should be able to use "2.2" to mean whatever 2.2 release we aliased (currently 2.2.1).

witlessbird:
@domcleal: updated 2.2.2 alias for slaves.

domcleal:
Did you want to try 2.2.3 while we're here, if it works for you with rvm stable?

witlessbird:
lemme try installing 2.2.3.

witlessbird:
@domcleal: switched to 2.2.3

domcleal:
Thanks @witlessbird, I'll check that it installs correctly.

PR #193, opened 2015-10-21
Ensure libvirt-dev(el) is present on the slave

ehelms:
When trying to apply the slave module to a clean Centos 7 and then run the test_develop script, bundle install failed without libvirt-devel installed.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -141,6 +141,12 @@ class slave($github_user = undef,
       name   => $::osfamily ? {
         Debian  => 'libaugeas-dev',
         default => 'augeas-devel'
+      };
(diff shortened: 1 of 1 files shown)


mmoll:
:+1:

domcleal:
That resource is already provided on our slaves via the libvirt class?

ehelms:
So would doing an include of the libvirt class be acceptable here? I feel
applying the slave class to a box should handle configuring a box to act as
a slave that can run the test suites.

domcleal:
Not really sure we should include it again, it's already in our Builder host group.

ehelms:
I get what your point is; however the result is that the slave module cant
configure a complete slave that the unit test job can run on which seems
broken.

mmoll:
what's the status here?

PR #188, opened 2015-09-24
Add test_develop configuration

ehelms:

PR #194, opened 2015-10-23
Add test_katello_core

ehelms:
Currently relies on #188

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/clone_foreman.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/clone_foreman.sh
@@ -1,0 +1,4 @@
+#!/bin/bash -ex
+
+rm -rf foreman/
+git clone https://github.com/theforeman/foreman --branch "${foreman_branch}"
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello_core.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello_core.yaml
@@ -1,0 +1,36 @@
+- job:
+    name: test_katello_core
+    project-type: matrix
+    logrotate:
+      daysToKeep: -1
+      numToKeep: 32
+    properties:
+      - github_katello
+    scm:
+      - katello
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+            - 1.9.3
+      - axis:
+          type: user-defined
+          name: database
+          values:
+            - postgresql
+      - axis:
+          type: label-expression
+          name: slave
+          values:
+            - fast
+    builders:
+      - clone_foreman
+      - test_katello
+    publishers:
+      - archive:
+          artifacts: 'plugin/pkg/*,Gemfile.lock'
+          only-if-success: false
+      - gemset_cleanup
+      - junit:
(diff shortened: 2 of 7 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello_core.yaml
```diff
@@ -0,0 +1,36 @@
...
+    name: test_katello_core
+    project-type: matrix
+    logrotate:
+      daysToKeep: -1
+      numToKeep: 32
+    properties:
+      - github_katello
+    scm:
+      - katello
+      - foreman
```
ehelms: If we enable the MultipleSCMs plugin then for plugins we can use the SCM configuration to manage the base foreman and plugin repositories that are needed for running tests like this example here. This would require a small change proposed here (and not in #188) to clone foreman to a foreman directory. See:

https://github.com/theforeman/foreman-infra/pull/194/files#diff-cb6e769e57f35b144c8a27a6b3508b53R7
https://github.com/theforeman/foreman-infra/pull/194/files#diff-e1dc954418569c69c1e176c941c38c0fR3

ehelms:
Rebased and updated

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman.yaml
```diff
@@ -2,7 +2,8 @@
     name: foreman
     scm:
       - git:
-         url: git://github.com/theforeman/foreman.git
-         wipe-workspace: true
-         branches:
-          - 'develop'
+          url: git://github.com/theforeman/foreman.git
+          wipe-workspace: true
+          basedir: foreman
```
domcleal: Changing this will mean you also need to update the paths in publishers in test_develop.

ehelms: Reverted that change for now, as it applied for the case of using multiple scm and that will be done as a re-factor after this.

domcleal:
Thanks @ehelms.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/clone_foreman.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -ex
+
+rm -rf foreman/
+git clone https://github.com/theforeman/foreman --branch "${foreman_branch}"
```
domcleal: The `foreman_branch` parameter seems to be undefined: http://ci.theforeman.org/job/test_katello_core/16490/database=postgresql,ruby=1.9.3,slave=fast/console.  Some slaves (EL6?) are using a default branch instead, some aren't.

ehelms: Thanks for pointing that out - I have pushed a fix to add that parameter to the job with a default.

domcleal: Are you sure you did?

ehelms: Blast - I pushed it to a branch instead. So I just opened a PR with it instead -- https://github.com/theforeman/foreman-infra/pull/203

PR #195, opened 2015-10-27
Test plugin matrix

ehelms:
Requires #188

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/gemset_cleanup.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/gemset_cleanup.sh
@@ -1,8 +1,14 @@
 #!/bin/bash
 [ -z "$ruby" ] && ruby=2.0.0
+
+if [ -d foreman ];then
+  cd foreman/
+fi
+
 # Clean gemset and database
 . /etc/profile.d/rvm.sh
 gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
 rvm use ruby-${ruby}@${gemset}
+bundle exec rake db:drop
 rvm gemset delete ${gemset} --force
 exit 0
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
@@ -1,0 +1,67 @@
+#!/bin/bash -ex
+
+TOP_ROOT=`pwd`
+APP_ROOT=$TOP_ROOT/foreman
+PLUGIN_ROOT=$TOP_ROOT/plugin
+
+cd $APP_ROOT
+
+### START test_develop ###
+# This section is from test_develop, please keep it in sync
+
+# setup basic settings file
+sed -e 's/:locations_enabled: false/:locations_enabled: true/' $APP_ROOT/config/settings.yaml.example > $APP_ROOT/config/settings.yaml
+sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' $APP_ROOT/config/settings.yaml
+
+# RVM Ruby environment
+. /etc/profile.d/rvm.sh
+# Use a gemset unique to each executor to enable parallel builds
+gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
+rvm use ruby-${ruby}@${gemset} --create
+rvm gemset empty --force
+#gem update --no-ri --no-rdoc
+gem install bundler --no-ri --no-rdoc
+
+# Retry as rubygems (being external to us) can be intermittent
+while ! bundle install --without=development -j5; do
+  bundle clean --force || true
+  (( c += 1 ))
+  if [ $c -ge 5 ]; then
+    echo "bundle install continually failed" >&2
+    exit 1
+  fi
+done
+
+# Database environment
+(
+  sed "s/^test:/development:/; s/database:.*/database: ${gemset}-dev/" $HOME/${database}.db.yaml
+  echo
+  sed "s/database:.*/database: ${gemset}-test/" $HOME/${database}.db.yaml
+) > $APP_ROOT/config/database.yml
+
+# Create DB first in development as migrate behaviour can change
+bundle exec rake db:drop db:create db:migrate
+
+### END test_develop ###
+
+# Now let's introduce the plugin
+echo "gem '${plugin_name}', :path => '${PLUGIN_ROOT}'" >> bundler.d/Gemfile.local.rb
+
+# Plugin specifics..
+[ -e ${PLUGIN_ROOT}/script/ci/katello.yml ] && cp ${PLUGIN_ROOT}/script/ci/katello.yml ${PLUGIN_ROOT}/config/katello.yml
+
+# Update dependencies
+while ! bundle update; do
+  (( c += 1 ))
+  if [ $c -ge 5 ]; then
+    echo "bundle update continually failed" >&2
+    exit 1
+  fi
+done
+
+# Now let's add the plugin migrations
+bundle exec rake db:migrate
+
+tasks="jenkins:unit"
+[ ${database} = postgresql ] && tasks="$tasks jenkins:integration"
+bundle exec rake $tasks TESTOPTS="-v"
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
@@ -1,0 +1,71 @@
+- job:
+    name: test_plugin_matrix
+    project-type: matrix
+    logrotate:
+      daysToKeep: 45
+      numToKeep: -1
+    parameters:
+      - string:
+          name: plugin_repo
+          default:
+          description: "Git URL containing the Foreman plugin, e.g. <pre>https://github.com/theforeman/foreman_example</pre>"
+      - string:
+          name: plugin_branch
+          default: master
+          description: "Git branch name of the Foreman plugin, e.g. <pre>master</pre>"
+      - string:
+          name: plugin_name
+          default:
+          description: "Name of the Foreman plugin/gem, in the format that should be in the Gemfile, e.g. <pre>foreman_test</pre>"
+      - string:
+          name: foreman_repo
+          default: https://github.com/theforeman/foreman
+          description: "Git URL containing Foreman itself, which the plugin will be added to,
+          e.g. <pre>https://github.com/theforeman/foreman</pre>"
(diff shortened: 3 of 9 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash -ex
+
+[ -e plugin ] && rm -rf plugin/
+git clone -b ${plugin_branch} ${plugin_repo} plugin
```
ehelms: Similar to another PR, I could remove the need to do this by using the MultipleSCM plugin (https://wiki.jenkins-ci.org/display/JENKINS/Multiple+SCMs+Plugin) to handle parameterized cloning. Or could do that as an enhancement if we simply wanted to replace the job as-is in Jenkins with the JJB version.

ehelms:
Once test_develop configuration is in, I can also re-factor this to re-use the common script pieces across them.

ehelms:
Before I go further on this, what do you think about adding the MultipleSCM plugin to handle cloning foreman and the plugin repository instead of using a script to do it?

domcleal:
I don't think you need the plugin for that, isn't it just for handling multiple *types* of SCMs?  It looks like you can add multiple git repos to a single job already.

I'd suggest though that we do this in two stages though - convert the job, then refactor it.  It'll be much easier to review.

ehelms:
Fair enough - I'll tidy this up to represent as is configuration first.

For MultipleSCM, on the surface it looks like multiple types of SCMS, but you can add multiple "git" repositories with full configuration. A single instance of the git SCM will allow you to add multiple repository URLS, but not configure properties specific to it. For example, what directory to check it out to, what branch(es) to target, etc. It would allow us to specify clone foreman and the plugin. And for the plugin clone to the 'plugin' directory for branch 'master'. For Foreman, clone it to develop or $parameter branch, to directory foreman .

ehelms:
This is now re-based and updated.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/wrappers/abort_build.yaml
```diff
@@ -0,0 +1,8 @@
+- wrapper:
+    name: abort_build
```
domcleal: We've some other uses of this in systest_foreman and deploy_deb, did you want to refactor that in at the same time?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+      daysToKeep: 45
+      numToKeep: -1
+    parameters:
+      - string:
+          name: plugin_repo
+          default:
+          description: "Git URL containing the Foreman plugin, e.g. <pre>https://github.com/theforeman/foreman_example</pre>"
+      - string:
+          name: plugin_branch
+          default:
```
domcleal: default: master

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+          description: "Git branch name of the Foreman plugin, e.g. <pre>master</pre>"
+      - string:
+          name: plugin_name
+          default:
+          description: "Name of the Foreman plugin/gem, in the format that should be in the Gemfile, e.g. <pre>foreman_test</pre>"
+      - string:
+          name: foreman_repo
+          default:
```
domcleal: default is https://github.com/theforeman/foreman

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+          description: "Git URL containing Foreman itself, which the plugin will be added to,
+          e.g. <pre>https://github.com/theforeman/foreman</pre>"
+      - string:
+          name: foreman_branch
+          default:
```
domcleal: develop

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+          description: "Git branch name of Foreman itself, e.g. <pre>develop</pre>"
+    concurrent: true
+    scm:
+      - git:
+          url: ${foreman_repo}
+          branches:
+            - upstream/${foreman_branch}
+          basedir: 'foreman'
+          prune: true
+          wipe-workspace: false
```
domcleal: true

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+      combination-filter: "!(foreman_branch == '1.7-stable' && ruby == '2.1')"
+    wrappers:
+      - default_xvfb
+      - abort_build
+      - timestamps
+    builders:
+      - clone_repository
+      - test_plugin
+    publishers:
+      - gemset_cleanup
```
domcleal: This is missing the db:drop from the original, and the same for test_develop actually.  The plugin version also needs to cd to foreman/.

domcleal:
> For MultipleSCM, on the surface it looks like multiple types of SCMS, but you can add multiple "git" repositories with full configuration. A single instance of the git SCM will allow you to add multiple repository URLS, but not configure properties specific to it. For example, what directory to check it out to

Ah yes, I see now.  Thanks.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/wrappers/abort_build.yaml
ehelms: Looks like those jobs have 60 and 30 minute timeouts -- do you want me to keep that as is or use 120 for all things?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
ehelms: Updated -- https://github.com/theforeman/foreman-infra/pull/195/files#diff-fd2ec544c14fd176a5ae4d65c093af34R7

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/gemset_cleanup.sh
```diff
@@ -4,5 +4,6 @@
 . /etc/profile.d/rvm.sh
 gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
 rvm use ruby-${ruby}@${gemset}
+bundle exec rake db:drop
```
domcleal: In the plugin job you'll also need to `cd foreman/` as IIRC each different script executes from the job workspace rather than the PWD of the last script.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/wrappers/abort_build.yaml
domcleal: Keeping it would be good I think, those jobs shouldn't take that long.  Either perhaps hardcode it using the percentage type settings in this plugin or simply parameterise this wrapper macro: http://docs.openstack.org/infra/jenkins-job-builder/definition.html#macro

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/gemset_cleanup.sh
ehelms: Updated to check if foreman/ exists -- https://github.com/theforeman/foreman-infra/pull/195/files#diff-fd2ec544c14fd176a5ae4d65c093af34R4

ehelms:
Parameterized the abort_build job and updated those jobs back to their original timeouts

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,71 @@
...
+      daysToKeep: 45
+      numToKeep: -1
+    parameters:
+      - string:
+          name: plugin_repo
+          default: master
+          description: "Git URL containing the Foreman plugin, e.g. <pre>https://github.com/theforeman/foreman_example</pre>"
+      - string:
+          name: plugin_branch
+          default:
```
domcleal: master

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -0,0 +1,71 @@
+- job:
+    name: test_plugin_matrix
+    project-type: matrix
+    logrotate:
+      daysToKeep: 45
+      numToKeep: -1
+    parameters:
+      - string:
+          name: plugin_repo
+          default: master
```
domcleal: not master :)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/wrappers/abort_build.yaml
```diff
@@ -0,0 +1,8 @@
...
+    wrappers:
+      - timeout:
+          type: absolute
+          timeout: {timeout}
```
domcleal: I'm not sure what's happening here precisely, perhaps a lack of quoting?  In the XML output I see:

    &lt;timeoutMinutes>OrderedDict([('timeout', None)])</timeoutMinutes>

ehelms: I have tried a number of things and so far this just isn't populating. Going to continue investigating but I am bit stumped based upon the documentation.

domcleal: https://github.com/domcleal/foreman-infra/commit/028914b8e4df27728dac4ebe20fb76650d102fd0 seems to work for me.

ehelms: Turns out, I was getting a tuple index error when the timeout parameter was indented 4 spaces and I interpreted that incorrectly. The real culprit was the `{0}` not being escaped after introducing the parameter. Updated the other jobs based on your suggestion.

domcleal:
Thanks @ehelms!

PR #196, opened 2015-11-06
Add foreman_pipeline to jjb

xprazak2:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_pipeline.yaml
```diff
@@ -0,0 +1,6 @@
+- project:
+    name: foreman_pipeline
+    repo: foreman_pipeline
```
domcleal: This is the default (taken from the name)

domcleal:
Are you following up with PR testing after?

xprazak2:
@domcleal: I would very much like that

domcleal:
Sure, you can either do it now via the second set of instructions on http://projects.theforeman.org/projects/foreman/wiki/Jenkins#Adding-a-new-Foreman-plugin if your tests are passing OK, or we can merge this to see if it works (once that comment's fixed) and enable PR tests after.

xprazak2:
Everything I need to test PRs shlould be configured now, except for the  `value of GITHUB_SECRET_TOKEN from rhc env list -a prprocessor -n theforeman` in repo webhook, may I be so bold and ask for help, @domcleal ?

puppet/modules/slave/templates/test_pull_requests_foreman_pipeline.json.erb
```diff
@@ -0,0 +1,49 @@
...
+    "proxy_port": null,
+    "jenkins_host": "http://ci.theforeman.org/",
+    "jenkins_url": "http://ci.theforeman.org/",
+    "hub_config": "~/.config/hub",
+    "github_user": "theforeman",
+    "bot_github_user": "<%= github_user -%>",
+    "irc_channel": "#theforeman-dev",
+    "not_mergable_cache": "~/pr_tests/cache/test_pull_request_foreman_pipeline_not_mergable",
+    "repo_groups": {
+        "foreman_pipeline": "foreman_pipeline",
```
domcleal: The trailing comma here isn't valid JSON.

domcleal:
Sure, I'll add the webhook on merge, don't worry about that.

domcleal:
Oh sorry, I didn't know you're updated this.  In light of https://github.com/theforeman/foreman-packaging/pull/892 I'm unsure if it'll work as-is because our jobs only ever install Foreman and add the plugin in on top.  If you can specify Katello as a dependency (Gemfile, gemspec) and a regular db:migrate/seed works then it might be OK.

xprazak2:
@domcleal: I updated the foreman_pipeline gemspec with Katello dependency, if there is anything else, let me know...

domcleal:
Thanks, it should show up in an hour or two under the Plugins tab on Jenkins.

PR #197, opened 2015-11-13
Created nightly symlink in releases dir for easier shell scripts

lzap:
SSIA

It's this one: http://yum.theforeman.org/plugins/nightly/


domcleal:
Cheers!

PR #198, opened 2015-11-23
Set branch parameter by default for test_katello_core

ehelms:


ehelms:
Test runs fail without this as it has no revision to checkout

domcleal:
Thanks, merged.  I've also granted you commit to this and other infra repos, so if something's obviously broken, please just go ahead and fix it.

PR #199, opened 2015-11-25
Enable Travis CI to test Puppet syntax and JJB

domcleal:


ehelms:
Great idea, can even see the output to inspect the XML - ACK

PR #201, opened 2015-11-25
Add hammer_cli_foreman_remote_execution

stbenjam:
I was trying to use travis but it's too slow against theforeman org, I'm fairly certain we're being severely throttled.  Anything else I need to do? Seems like the parent tasks aren't in foreman-infra, so `test_hammer_cli_foreman_remote_execution` would need creating manually?


ehelms:
Do those test jobs exist already? Guessing not, so you'd need add them as JJB configurations.

stbenjam:
How do I do that?

stbenjam:
All I see in this repo for the other hammer plugins is the changes I've made here

ehelms:
Here is an example PR of adding a job -- https://github.com/theforeman/foreman-infra/commit/6183146731c1db2f1fbee0a032068f4a2240c4d2

I'm not sure we have a strict policy of new jobs need to go through JJB process at the moment, but better IMO to push new jobs in that direction so we aren't continually playing catch up.

domcleal:
I don't mind making the jobs manually for now, we only have the Hammer discovery project existing in Jenkins right now which could be copied over.  It needs refactoring really into a general Hammer plugin test (like Foreman and the smart proxy jobs) so I'm not too bothered about putting it into JJB before it's been refactored.

stbenjam:
I'm happy with the current Travis setup, so gonna close this

PR #202, opened 2015-12-08
Add test_hammer_cli_foreman job configuration

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli_foreman.yaml
```diff
@@ -0,0 +1,40 @@
+- job:
+    name: test_hammer_cli_foreman
+    project-type: matrix
+    concurrent: true
+    logrotate:
+      daysToKeep: 14
+      numToKeep: -1
+    properties:
+      - github:
+          url: https://github.com/theforeman/foreman
```
domcleal: hammer-cli-foreman

ehelms: Updated

domcleal:
Thanks!

PR #203, opened 2015-12-09
Add foreman_branch parameter to test_katello_core

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello_core.yaml
```diff
@@ -11,6 +11,10 @@
           name: branch
           default: master
           description: "Branch to run test suite on"
+      - string:
+          name: foreman_branch
+          default: develop
+          description: "Foreman branch to run test suite against "
```
domcleal: Nitpick, extra space here?

ehelms:
Updated

domcleal:
Thanks!

PR #204, opened 2015-12-10
Create dedicated gemset_create builder

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/deploy_web.sh
```diff
@@ -3,14 +3,6 @@
 # Build the site on the slave 
 
 ruby=2.0.0
```
domcleal: unused?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello_core.yaml
```diff
@@ -37,6 +37,7 @@
             - el7
     builders:
       - clone_foreman
+      - create_gemset
```
domcleal: gemset_create

ehelms:
Updated

domcleal:
Would you mind rebasing this to factor out the gemset creation in the hammer_cli_foreman test?

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/deploy_katello_site.sh
```diff
@@ -1,16 +1,5 @@
 #!/bin/bash -xe
 
-# Build the site on the slave 
-ruby=2.0.0
-# RVM Ruby environment
-. /etc/profile.d/rvm.sh
```
domcleal: Will moving this out of the shell script prevent the rest of the test from working?  Each build step is going to spawn a new shell, so the environment won't be inherited from one to the next.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/builders/gemset_create.yaml
```diff
@@ -0,0 +1,4 @@
+- builder:
+    name: gemset_create
+    builders:
+      - shell: !include-raw scripts/gemset_create.sh
```
domcleal: Rather than a separate shell builder step, you might need to use a list of files in *other* builders and add gemset_create.sh first.  (Bottom examples at http://docs.openstack.org/infra/jenkins-job-builder/definition.html#module-jenkins_jobs.local_yaml)

ehelms: Good point and I think that could work, before I try that, what are your thoughts on using https://wiki.jenkins-ci.org/display/JENKINS/RVM+Plugin ? Looks like JJB has a macro for it (http://docs.openstack.org/infra/jenkins-job-builder/wrappers.html?highlight=rvm#wrappers.rvm-env).

domcleal: I'm not sure how/where the gemset name would be generated, since currently it's taking part of the job name (I'm guessing the `cut` means there was a slash in the job name, but I can't remember for sure).

I guess a parameter could be written straight into the job YAML and passed into gemset_create?

Another thought I've had is to get rid of per-job+executor gemsets and leave gems installed permanently in a per-executor gemset.  This would mean we just use ${EXECUTOR_NUMBER} in the rvm plugin's gemset name, which ought to expand.

ehelms: What about a more unique environment variable that should be present such as BUILD_ID or BUILD_TAG ?

domcleal: The only problem with that is when the job fails or is aborted, and something goes wrong with the cleanup (e.g. abort was pressed twice), the gemset will never get cleaned up and we may slowly run out of disk space on slaves.  If there's a chance the gemset will get reused then it might get cleaned up properly later.

The reason I've been thinking of using just build executor is that we might be able to remove the cleanup and leave gems installed, making subsequent builds faster (Gemfile.lock gets recreated, but the gem is already installed).

I think using just build executor without any job name would be fine, irrespective of removing cleanup or not.

ehelms: Fair enough - do you (or me but you have the power to install plugins) want to try that with the rvm plugin on a job or stick with this refactoring?

domcleal: Sure, I'll install the plugin when the instance is quiet (probably over the weekend) and we can check if it'll work next week.  It's probably not worth refactoring this twice.

ehelms: Great - thanks!

domcleal: Rvm plugin's installed, it's under the Build Environment section of the job config.

ehelms: First test of this, we'll need bundler present which means I can:

1) Create a builder that installs bundler into the RVM environment for job configurations to use
2) Update the slave manifest to ensure bundler is automatically installed into the base RVM that the gemsets use

ehelms: Test run where I removed all but the bundler install and set the rvm environment to ${ruby}@${EXECUTOR_NUMBER}

http://ci.theforeman.org/job/test_katello_core/database=postgresql,ruby=1.9.3,slave=fast/16669/console

domcleal: > http://ci.theforeman.org/job/test_katello_core/database=postgresql,ruby=1.9.3,slave=fast/16669/console

This looks like it's using Ruby 1.8.7, given the number of errors.  I guess it didn't inherit the rvm environment.

ehelms: I am not sure what's going on, if you look at the output of http://ci.theforeman.org/job/test_katello_core/database=postgresql,ruby=1.9.3,slave=fast/16683/console it appears to have the correct gemset enabled and in use, but the bundle install doesn't indicate that it is.

domcleal: It could be the shebang line in the scripts creating a new shell, the plugin may rely on using the default shell (I've not checked).

ehelms: I tried removing the shebangs, no dice. I also tried adding activation of the rvm environment, using rvm gemset use explicitly and still hitting the same issue without any insight into. If you have other ideas, I am happy to try otherwise I may have to abandon this strategy.

PR #206, opened 2016-02-01
Adds Katello as optional step for plugin testing

xprazak2:
foreman_pipeline plugin requires Katello to run but our Jenkins is not prepared for it. When job on PR is triggered, Jenkins clones Foreman, then clones Foreman Pipeline and tries to build it. To force presence of Katello, Foreman Pipeline has it listed in dependencies to make sure it is available. Unfortunately, it is not possible to specify git source in gemspec, so Jenkins gets Katello from rubygems which causes gem clashes. This should introduce optional step into test_plugin_matrix job which clones Katello's repo

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
@@ -49,6 +49,7 @@ echo "gem '${plugin_name}', :path => '${PLUGIN_ROOT}'" >> bundler.d/Gemfile.loca
 
 # Plugin specifics..
 [ -e ${PLUGIN_ROOT}/script/ci/katello.yml ] && cp ${PLUGIN_ROOT}/script/ci/katello.yml ${PLUGIN_ROOT}/config/katello.yml
+[ -e ${PLUGIN_ROOT}/gemfile.d/${plugin_name}.rb ] && cat ${PLUGIN_ROOT}/gemfile.d/${plugin_name}.rb >> bundler.d/Gemfile.local.rb
 
 # Update dependencies
 while ! bundle update; do


domcleal:
Use a Gemfile.

xprazak2:
I tried [this](https://github.com/theforeman/foreman_pipeline/blob/master/Gemfile) and several other variations, it is completely ignored.

domcleal:
Oh, I guess because it's installed from Foreman's gemfile rather than yours.  It's probably preferable to copy a bundler.d/-esque file from the plugin then, as many ship one.  Either that or simply depend on the rubygem from the gemspec.

xprazak2:
Depending on Katello from gemspec simply does not work, because head of Foreman's develop branch and latest released Katello gem are not always compatible. Copying from bundler.d could  work, so I added one "plugin specifics" into test_plugin script, which shloud copy content of [this](https://github.com/xprazak2/foreman_pipeline/commit/ff61f24508f20484262c0a49ef8bd2d067defa59) file.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
```diff
@@ -49,6 +49,7 @@ echo "gem '${plugin_name}', :path => '${PLUGIN_ROOT}'" >> bundler.d/Gemfile.loca
 
 # Plugin specifics..
 [ -e ${PLUGIN_ROOT}/script/ci/katello.yml ] && cp ${PLUGIN_ROOT}/script/ci/katello.yml ${PLUGIN_ROOT}/config/katello.yml
+[ -e ${PLUGIN_ROOT}/bundler.d/foreman_pipeline.rb ] && cat ${PLUGIN_ROOT}/bundler.d/foreman_pipeline.rb >> bundler.d/Gemfile.local.rb
```
domcleal: Should use the plugin name from the parameter, not be hardcoded to one plugin.

domcleal:
The naming convention you're suggesting seems confusing - isn't that typically going to suggest a runtime bundler.d file, rather than development/test-only?  It won't be any use at runtime if, I assume, the contents point to a git branch - and it's common for plugins (esp smart proxy ones) to include the *runtime* file at bundler.d/.  It might be better to call it something clearer (i.e. closer to Gemfile in the root).

The commit message also doesn't describe the patch.

xprazak2:
I changed it to [gemfile.d](https://github.com/xprazak2/foreman_pipeline/commit/89cb76866ac158124166d28fc3897c2f54b1a243), for the lack of better name. I also updated the message

domcleal:
Thanks, merged and will be active in an hour or so.

xprazak2:
Great, thanks.

PR #208, opened 2016-02-02
adding predefined lookaside support for repclosure

jlsherrill:
and removing openscap from being used by default


domcleal:
I think you need to add the parameter to the job YAML too, under puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_repoclosure.yaml.

jlsherrill:
@domcleal thanks for the pointer!

domcleal:
Thanks @jlsherrill, should be running within an hour or so.

PR #210, opened 2016-02-09
Manage SELinux configs for freight and rsync access

domcleal:
Content stored in the web root should also be accessible via rsync, so
use a shared context.  Freight staing content should all be under the
public content context so it remains correct when copied to the web
roots.

---

This just implements the manually configured SELinux policy changes on web01 in Puppet.  Replaces @lzap's #142.


GregSutcliffe:
Looks sane. As stated elsewhere, strugling to test locally but happy to merge in a day or two if no other concerns raised.

PR #213, opened 2016-02-17
Drop 1.9.3 and 2.1 testing from katello core testing

ehelms:


daviddavis:
LGTM

jlsherrill:
:+1: from me.  Katello does not support these versions of ruby in production and I think the likelihood that we will soon is very low and not worth the extra testing

PR #214, opened 2016-02-20
Add test_bastion_javascript

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/pr_checkout.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -xe
+
+cd plugin/
```
domcleal: This shouldn't be in a generic "pr_checkout", perhaps "plugin_pr_checkout" or use the root directory.

ehelms: Tried a multi-purpose solution for plugin and non-plugin pr checkouts -- let me know what you think. If you would still prefer they be separated logic for plugins and non-plugins I'll generate two scripts.

domcleal: I think separate's probably going to be best, or it needs to take a variable.. it's probably best not to assume a certain layout, or what type of PR we're testing.  (Perhaps it's testing a non-plugin PR against the plugin.)

ehelms:
Updated

domcleal:
Merged, thanks @ehelms.

PR #215, opened 2016-02-26
Switch to manual-only tests for forman bug day

GregSutcliffe:
As per -dev discussion, this should switch the Jenkins tests to manual-only. We can merge before Monday morn and revert after the bugday is done.


domcleal:
Looks fine to me, I'll try to merge it on Sunday night/Monday morning.  If you could send a warning to the dev list, it'd be appreciated - people might not have seen the in-thread discussion or this PR.

GregSutcliffe:
Good idea, will do by the end of the day.

PR #216, opened 2016-03-11
Test all branches on katello-packaging

jlsherrill:


domcleal:
Thanks, should be active within about 90 minutes.

PR #219, opened 2016-04-14
Creates db axis dynamically for test_plugin_matrix

xprazak2:
Allows plugins to specify, which databases to use for test_plugin_matrix. Current behaviour (testing against postgres, mysql and sqlite) is left as a default. The reason behind this is that foreman_pipeline needs Katello, but Katello does not officially support mysql. I have fixed a bunch of migration errors related to foreign keys that appear only on mysql, but I do not see how I could work around [Katello 's index being too long](https://github.com/Katello/katello/blob/master/db/migrate/20150613134559_add_rpm.rb#L20). It Requires Dynamic axis plugin for Jenkins.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
@@ -1,6 +1,6 @@
 - project:
     name: foreman_discovery
-    repo: foreman_discovery
+    defaults: plugin
     branch:
       - develop:
           foreman_branch: develop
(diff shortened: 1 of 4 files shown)


domcleal:
The tests are failing.

xprazak2:
@domcleal, tests are now passing

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
```diff
@@ -1,6 +1,7 @@
 - project:
     name: foreman_discovery
     repo: foreman_discovery
```
domcleal: repo is superfluous if applying the plugin defaults

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -44,12 +48,10 @@
             - 2.0.0
             - 2.1
       - axis:
-          type: user-defined
+          type: dynamic
```
domcleal: This looks like it requires an additional plugin, so it'll have to wait until I get an opportunity to install it.

domcleal:
Thanks, plugin installed and it seems to work OK.

PR #221, opened 2016-05-09
remove wheezy and precise from nightly builds

mmoll:
I don't know if that's sufficient...


domcleal:
Thanks, I'll fix up the coreproject job too.

PR #224, opened 2016-05-16
Change 'postgres' to 'postgresql'

xprazak2:
I found out foreman_pipeline has [builds from other plugins](http://ci.theforeman.org/job/test_plugin_matrix/1445/) in the matrix. This was hopefully caused by specifying 'postgres' instead of 'postgresql' as one of the databases. ~~We should also be ok without defaults in test_plugin_matrix since they are supplied by test_plugin.~~

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_pipeline.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_pipeline.yaml
@@ -1,6 +1,6 @@
 - project:
     name: foreman_pipeline
     defaults: plugin
-    dbs: 'postgres sqlite3'
+    dbs: 'postgresql sqlite3'
     jobs:
       - 'test_plugin_{name}_{branch}'


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -28,7 +28,7 @@
           description: "Git branch name of Foreman itself, e.g. <pre>develop</pre>"
       - string:
           name: dbs
-          default: 'postgresql mysql sqlite3'
+          default:
```
domcleal: Why remove the default value?

domcleal: Oh, I see your comment in the PR description now, but I don't understand why this is an improvement.

xprazak2: It just seemed unnecessary, but I will leave it alone since it does not change anything.

domcleal: It's useful when running this job directly, it saves users having to enter a value.

domcleal:
Thanks, it'll be rolled out within 40 mins or so.

PR #226, opened 2016-06-08
run npm install if package.json is present

tbrisker:
This will allow running tests on PRs that need npm modules.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
@@ -24,6 +24,9 @@ while ! bundle install --without=development -j 5; do
   fi
 done
 
+# run npm install if package.json exists and we are running integration tests (which only run on postgresql)
+[ ${database} = postgresql ] && [ -e "$APP_ROOT/package.json"] && npm install
+
 # Database environment
 (
   sed "s/^test:/development:/; s/database:.*/database: ${gemset}-dev/" $HOME/${database}.db.yaml
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
@@ -32,6 +32,9 @@ while ! bundle install --without=development -j5; do
   fi
 done
 
+# run npm install if package.json exists
+[ -e "$APP_ROOT/package.json"] && npm install
+
 # Database environment
 (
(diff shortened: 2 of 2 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,9 @@ while ! bundle install --without=development -j 5; do
   fi
 done
 
+# run npm install if package.json exists
+[ -e "package.json"] && npm install
```
ohadlevy: nit: are you sure its the $CWD ? maybe full path would be better?

tbrisker: fixed to use $APP_ROOT

dLobatog:
:+1: works well here

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,9 @@ while ! bundle install --without=development -j 5; do
   fi
 done
 
+# run npm install if package.json exists
+[ -e "$APP_ROOT/package.json"] && npm install
```
ehelms: Since integration tests only run on postgres, I used this for the pull request test:

```
[ ${database} = postgresql ] && [ -e "$APP_ROOT/package.json" ] && npm install
```

domcleal:
There's an unresolved comment above, plus test_plugin.sh should also be updated to match. The plugin one will need to run `npm install` on all databases, not only PostgreSQL, as plugins don't have separate integration/unit test tasks generally (they hook jenkins:unit for all).

dLobatog:
Would https://github.com/appcelerator/npm-fast-install  be possible to use here?

ehelms:
Since npm-fast-install appears to be a globally installed module, I believe it wold follow this model [1]. Could we do that as a follow up? It would be nice to unblock test_develop

[1] https://github.com/theforeman/foreman-infra/blob/9df925271bbd407d40227d0b0252d02cb14576ca/puppet/modules/slave/manifests/init.pp#L165

tbrisker:
@domcleal Addressed the comment and added the line to test_plugin.sh as well. Would it also be needed for the other scripts there? I wasn't sure.
@ehelms @dLobatog not quite sure I understand your comments, is this a suggestion for a future improvement or something that should go into this PR?

dLobatog:
@tbrisker Not for this PR, I don't know even if the jobs are isolated or not in a way that makes npm-fast-install useless, it was just a question

domcleal:
Merged, thanks @tbrisker.

PR #227, opened 2016-06-21
use Ruby 2.3 to build the website

mmoll:
recent jekyll versions aren't tested with Ruby 2.0 anymore


mmoll:
I don't get why still only version 59 of github-pages is getting installed...

domcleal:
Thanks @mmoll. I've changed the workspace to be wiped between builds in 801f4ba which will cause the Gemfile.lock to be discarded and the latest gems to be installed.

PR #228, opened 2016-07-06
Fixes #15595 - Do not test with Jenkins on 1.9.3

akofink:
Dependent on https://github.com/theforeman/foreman-packaging/pull/1212

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli_foreman.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli_foreman.yaml
@@ -19,10 +19,10 @@
           type: user-defined
           name: ruby
           values:
-            - 1.9.3
             - 2.0.0
             - 2.1
             - 2.2
+            - 2.3
     builders:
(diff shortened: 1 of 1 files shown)


domcleal:
Fine by me, up to you @mbacovsky or @tstrachota.

mmoll:
maybe also add 2.3?

akofink:
@mmoll I like that idea, but should it be associated with this issue or a separate one?

mmoll:
I'd say just do it. We do run hammer on Ubuntu/xenial under 2.3, so we did miss that a few months ago...

akofink:
@domcleal Is this blocked by anything else?

mmoll:
Personally I'd maybe open a PR against at least hammer-cli to update the gemspec accordingly. And otherwise the two people mentioned are the hammer wizards... But if the failing tests are real pain for you, I'm also not against merging this.

domcleal:
@akofink it's waiting for one of the two maintainers to confirm they want to drop support.

mbacovsky:
@domcleal, I'm fine with dropping 1.9.3 testing.

domcleal:
Merged, thanks.

PR #230, opened 2016-07-26
ci: add pr testing for expire_hosts & host_extra_validator

timogoebel:
@domcleal : Would you mind setting up hooks for the repos as well? I don't have sufficient permissions to to it myself.


domcleal:
Thanks @timogoebel. I'll set up the hooks now.

PR #231, opened 2016-08-04
Add PR testing to Ansible

dLobatog:


puppet/modules/slave/templates/test_pull_requests_foreman_ansible.json.erb
```diff
@@ -0,0 +1,49 @@
...
+    "settings": {
+        "test_settings": {
+            "name": "test",
+            "test_prefix": "Test Results:",
+            "allow_multiple": true,
+            "run_without_trigger": true,
+            "branches": {
+                "*": {
+                    "jenkins_job_name": "test_plugin_pull_request",
+                    "downstream_job_name": "test_plugin_foreman_ansible_master",
```
domcleal: This job testing master doesn't currently exist, you'll need to create it first (step 2 at http://projects.theforeman.org/projects/foreman/wiki/Jenkins#Adding-a-new-Foreman-plugin) or change to `true` to not test master at all.

domcleal: Sorry, I see now that's what the first file does. Could you also update the slave::pr_test_config list in puppet/modules/slave/manifests/init.pp? (Step 2 of the second part.)

dLobatog:
Updated - I don't have permissions to set the hook in the foreman_ansible project, so that's missing

domcleal:
Thanks, it'll be active within the next hour or so.

PR #233, opened 2016-08-11
Add PR testing to DigitalOcean

dLobatog:
I think this is all that's needed - can't set the hooks in the repo myself because I don't have permissions, would be glad to do it if I can get them


puppet/modules/slave/templates/test_pull_requests_foreman_digitalocean.json.erb
```diff
@@ -0,0 +1,49 @@
...
+    "settings": {
+        "test_settings": {
+            "name": "test",
+            "test_prefix": "Test Results:",
+            "allow_multiple": true,
+            "run_without_trigger": true,
+            "branches": {
+                "*": {
+                    "jenkins_job_name": "test_plugin_pull_request",
+                    "downstream_job_name": "test_plugin_foreman-digitalocean_master",
```
domcleal: This is foreman_digitalocean (underscore).

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman-digitalocean.yaml
```diff
@@ -0,0 +1,6 @@
+- project:
+    name: foreman-digitalocean
```
domcleal: This already exists with an underscore, leave it out.

puppet/modules/slave/templates/test_pull_requests_foreman_digitalocean.json.erb
```diff
@@ -0,0 +1,49 @@
...
+                    "build_token": "<%= jenkins_build_token -%>",
+                    "addtl_jenkins_params": {
+                        "foreman-digitalocean": {
+                            "plugin_repo": "https://github.com/theforeman/foreman-digitalocean.git",
+                            "plugin_name": "foreman-digitalocean"
```
domcleal: I think this needs to be underscore, it's the gem name that will be added to the Gemfile to load it.

dLobatog:
Updated to fix the underscores & removed the JJB yaml

domcleal:
Thanks @dLobatog!

PR #234, opened 2016-08-30
Reduce plugin matrix test job count by limiting ruby/database combo

ehelms:
Reduce the number of test jobs by only testing postgres and mysql
once and testing all supported rubies against sqlite.


ehelms:
The idea is this should test:

1.9.3 - sqlite
2.0.0 - sqlite
2.1 - sqlite
2.2 - sqlite
2.3 - sqlite

2.2 - postgres
2.2 - mysql

Seems redundant to test every ruby in combination with every database as the version of ruby shouldn't affect the database aspect enough to matter.

ehelms:
If this seems reasonable - I'd propose we do this same thing for test_develop and test_develop_pull_request

domcleal:
I'm not keen on removing tests, however:

1. Tests on all DBs should be using Ruby 2.3, not 2.2.
2. Tests on all Rubies should be on PostgreSQL, as that's where integration tests run and FKs/constraints are present.

ehelms:
Updated to reflect this break down:

1.9.3 - postgres
2.0.0 - postgres
2.1 - postgres
2.2 - postgres
2.3 - postgres

2.3 - mysql
2.3 - sqlite

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,16 @@
           values:
             - fast
     execution-strategy:
-      combination-filter: "!(foreman_branch == 'develop' && ruby == '1.9.3')"
+      combination-filter: >
+        "!(foreman_branch == 'develop' && ruby == '1.9.3') ||
+        !(ruby == '1.9.3' && database == 'sqlite') ||
```
domcleal: This is very hard to read, why use negations? Running combinations where database == 'postgresql' or ruby == '2.3' would simplify it considerably.

ehelms:
Fair point - updated to be in the positive instead of negative

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,15 @@
           values:
             - fast
     execution-strategy:
-      combination-filter: "!(foreman_branch == 'develop' && ruby == '1.9.3')"
+      combination-filter: >
+        "!(foreman_branch == 'develop' && ruby == '1.9.3') ||
+        (ruby == '1.9.3' && database == 'postgresql') ||
+        (ruby == '2.0.0' && 'postgresql') ||
```
domcleal: database ==

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,15 @@
...
+        (ruby == '2.1' && database == 'postgresql') ||
+        (ruby == '2.2' && database == 'postgresql') ||
+        (ruby == '2.3' && database == 'postgresql') ||
+        (ruby == '2.3' && database == 'mysql') ||
+        (ruby == '2.3' && database == 'sqlite')"
```
domcleal: No need to list each DB if you just check for ruby == 2.3

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,15 @@
           values:
             - fast
     execution-strategy:
-      combination-filter: "!(foreman_branch == 'develop' && ruby == '1.9.3')"
+      combination-filter: >
+        "!(foreman_branch == 'develop' && ruby == '1.9.3') ||
```
domcleal: This condition (line) will likely be true for say, Ruby 2.0.0 on MySQL, causing it to run every sub-job apart from Ruby 1.9.3 on develop, I think.

ehelms:
Updated

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,12 @@
           values:
             - fast
     execution-strategy:
-      combination-filter: "!(foreman_branch == 'develop' && ruby == '1.9.3')"
+      combination-filter: >
+        "(ruby == '1.9.3' && database == 'postgresql') ||
```
domcleal: The foreman_branch check is required here, it should only be run when the branch is older than develop (or remove this capability if no plugin is using it) as develop doesn't support 1.9.3.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,12 @@
...
+        (ruby == '2.0.0' && database == 'postgresql') ||
+        (ruby == '2.1' && database == 'postgresql') ||
+        (ruby == '2.2' && database == 'postgresql') ||
```
domcleal: Optimisation - lines 64-67 can simply be written as `database == 'postgresql'`, it'll match any Ruby version.

ehelms:
Updated based on comments

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_plugin_matrix.yaml
```diff
@@ -60,7 +60,10 @@
           values:
             - fast
     execution-strategy:
-      combination-filter: "!(foreman_branch == 'develop' && ruby == '1.9.3')"
+      combination-filter: >
+        "(foreman_branch != 'develop' && ruby == '1.9.3' && database == 'postgresql') ||
+        (database == 'postgresql') ||
```
domcleal: This will run the Ruby 1.9.3 on PostgreSQL test on develop, but Ruby 1.9.3 should be excluded.

ehelms:
Updated to exclude ruby 1.9.3 -- https://paste.fedoraproject.org/428006/38674201/

domcleal:
Merged, thanks @ehelms.

PR #237, opened 2016-09-09
update puppetlabs/postgresql to 4.8.0

mmoll:
also update puppetlabs/apt to 1.8.0 because pl/pgsql depends on that version.

This is needed to avoid this bogus "start" on Debian/jessie on every Puppet run:
```
Sep  9 20:55:32 banana puppet-agent[15235]: (/Stage[main]/Puppet::Agent::Service/Service[puppet]/ensure) ensure changed 'running' to 'stopped'
```


mmoll:
after switching to systemd (the default for Debian/jessie, but not Bananian Linux), this resolved itself.

PR #239, opened 2016-09-21
Upgrade npm before npm install

tbrisker:
This allows two major benefits: shrinkwrap can be used to only pin
certain dependencies rather then entire trees, and download size is
greatly reduced as the module tree is flattened to reduce duplications.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
@@ -24,8 +24,11 @@ while ! bundle install --without=development -j 5; do
   fi
 done
 
-# run npm install if package.json exists and we are running integration tests (which only run on postgresql)
-[ ${database} = postgresql ] && [ -e "$APP_ROOT/package.json" ] && npm install
+# we need to install node modules for integration tests (which only run on postgresql)
+if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
+  npm install npm # first upgrade to newer npm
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -25,7 +25,8 @@ while ! bundle install --without=development -j 5; do
 done
 
 # run npm install if package.json exists and we are running integration tests (which only run on postgresql)
-[ ${database} = postgresql ] && [ -e "$APP_ROOT/package.json" ] && npm install
+# and upgrade npm to the newest version first
+[ ${database} = postgresql ] && [ -e "$APP_ROOT/package.json" ] && npm install npm && $APP_ROOT/node_modules/.bin/npm install
```
domcleal: Nitpick, can you rewrite this into an `if` block please? (`if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then` with the two commands within the block? More than one command in a single line gets difficult to follow.

tbrisker: will do.

tbrisker:
I first want to unblock tests on jenkins, and for that just upgrading npm will do. The next step is to upgrade node as well so we can get rid of the version pinning with shrinkwrap.

tbrisker:
@domcleal fixed your comment.

domcleal:
Thanks @tbrisker, it ought to deploy within an hour or so.

PR #240, opened 2016-09-22
Remove ruby 1.8.7 from smart-proxy test matrix

witlessbird:
Seeing that puppet 2.6.0 was only being tested under ruby 1.8.7, I removed it from the matrix and the filter too.


domcleal:
Thanks @witlessbird! I'll remove it from the PR job manually.

PR #242, opened 2016-10-04
Change f19 to 24 on yum repo index

dLobatog:


puppet/modules/web/files/yum-HEADER.html
```diff
@@ -8,15 +8,15 @@
...
+  <li>/releases/1.4/f24/x86_64</li>
 </ul>
 
 <p>foreman-release RPMs containing an appropriate .repo file are available with fixed URLs:</p>
 
 <ul>
   <li><a href="http://yum.theforeman.org/releases/latest/el6/x86_64/foreman-release.rpm">http://yum.theforeman.org/releases/latest/el6/x86_64/foreman-release.rpm</a></li>
   <li><a href="http://yum.theforeman.org/releases/latest/el7/x86_64/foreman-release.rpm">http://yum.theforeman.org/releases/latest/el7/x86_64/foreman-release.rpm</a></li>
-  <li><a href="http://yum.theforeman.org/releases/latest/f19/x86_64/foreman-release.rpm">http://yum.theforeman.org/releases/latest/f19/x86_64/foreman-release.rpm</a></li>
+  <li><a href="http://yum.theforeman.org/releases/latest/f24/x86_64/foreman-release.rpm">http://yum.theforeman.org/releases/latest/f19/x86_64/foreman-release.rpm</a></li>
```
domcleal: Update the link text as well, please.

puppet/modules/web/files/yum-HEADER.html
```diff
@@ -8,15 +8,15 @@
 
 <ul>
   <li>/releases/1.7/el7/x86_64</li>
-  <li>/releases/1.4/f19/x86_64</li>
+  <li>/releases/1.4/f24/x86_64</li>
```
dLobatog: Actually let me revert this, it doesn't make sense (there's no 1.4/f24 version)

domcleal: Might be best to change the release number so it's a valid path, but it is only an example.

dLobatog:
Latest push updates the link and not just the text (sorry)

domcleal:
Thanks @dLobatog.

PR #243, opened 2016-10-11
Migrate test db while testing plugins

adamruzicka:
By default rake db:create creates databases for development and test
environments and rake db:migrate migrates the development database. When
we run tests the test environment is used by default and thanks to

`config.active_record.maintain_test_schema = true`

the test db gets migrated automatically. However this somehow doesn't
satisfy the access permissions tests for plugins which fail. Running the
tests against an already migrated database solves the issue.


domcleal:
This should be unnecessary, the test database should be created with the schema only. Foreman's initialisers should _not_ be run against the test database. This looks like http://projects.theforeman.org/issues/16821 which can be fixed in Foreman without changing the test environment.

adamruzicka:
Closing in favor of https://github.com/theforeman/foreman/pull/3933

PR #244, opened 2016-10-13
install unzip on jenkins slaves

mmoll:
see http://ci.theforeman.org/job/packaging_build_deb_coreproject/arch=armv7,label=debian,os=jessie/3727/console


domcleal:
Thanks @mmoll.

PR #246, opened 2016-10-15
update rubygems version to 2.5.2

mmoll:
As RVM with Ruby 2.3 is already bringing 2.5.1 by default, the current default leads to a change at every puppet run, as a downgrade 2.5.1 -> 2.4.8 is not done.


domcleal:
Thanks!

PR #247, opened 2016-10-16
use correct JDK on newer Ubuntus/Debians

mmoll:


puppet/modules/java/manifests/init.pp
```diff
@@ -1,9 +1,16 @@
 class java {
-  package { "java":
+  if $::osfamily == 'RedHat' {
+    $java_package = 'java-1.7.0-openjdk'
+  } else {
+    if  ($::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '16.04') >= 0) or
+        ($::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9') >= 0) {
+      $java_package = 'openjdk-8-jdk'
+  } else {
```
domcleal: Indentation's off here, it looks like another `if` clause. Or change the above `} else { if (..) {` to `elsif`

mmoll:
as it turns out, there's an easier solution for this

domcleal:
Thanks @mmoll.

PR #251, opened 2016-10-18
updated ruby 2.x.x versions to current stable ones

witlessbird:
This updates ruby 2.x.x versions to current stable ones. This should also resolve https://bugs.ruby-lang.org/issues/11910 that is occurring in tests on ruby 2.3.0.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -283,17 +283,17 @@ class slave($github_user = undef,
         version => 'ruby-1.9.3-p392',
       }
       slave::rvm_config { 'ruby-2.0.0':
-        version => 'ruby-2.0.0-p353',
+        version => 'ruby-2.0.0-p484',
       }
     }
     slave::rvm_config { 'ruby-2.1':
       version => 'ruby-2.1.5',
     }
     slave::rvm_config { 'ruby-2.2':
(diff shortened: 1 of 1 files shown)


mmoll:
I'd rather like to see some alignment to ruby versions that are used with Foreman packages. these would be 2.0.0p484 (Ubuntu/trusty), 2.1.5p273 (Debian/jessie) and 2.3.1p112 (Ubuntu/xenial).

witlessbird:
What would be the benefit of this in a build environment? All that's going to happen is we are going to see build failures for bugs that have been fixed...

mmoll:
@witlessbird exectly, we can catch problems before users are experiencing these in their environments.

witlessbird:
I can live with older versions for rubies < 2.2, or even <= 2.2.x. I'd prefer not to pin 2.3.1 to a particular patch-level, as it's the version that changes the most atm.

domcleal:
Since 2.1.x, there is no separate patch level release - 2.3.1 is always the same 2.3.1, there are no different "p" versions released.

witlessbird:
Changed ruby versions to the ones used by foreman.

mmoll:
:+1:

domcleal:
Merged, thanks @witlessbird. It'll take a while to update, give it an hour or two.

PR #252, opened 2016-10-22
replace homegrown ntp module with puppetlabs-ntp

mmoll:
- use version 4.2.0 for Puppet 3.x compatibility


domcleal:
Thanks @mmoll!

PR #253, opened 2016-10-24
enable testing for omaha plugins

timogoebel:


domcleal:
Thanks @timogoebel, it should start running within an hour or two. For PRs, update https://github.com/theforeman/prprocessor/blob/master/config/repos.yaml with pr_scanner: true (this will require a Redmine project).

timogoebel:
@domcleal : Thanks a lot. Are you saying, that i basically need a redmine project for pr testing?

domcleal:
It will work as-is, but it'll only check for PRs to test every 30 minutes. If that's fine, leave it.

The prprocessor uses a GitHub webhook to accelerate that, but yes, it requires a Redmine project to be configured for each repo right now. You could use the generic Plugins sub-project if you wish.

timogoebel:
@domcleal : Thanks for the explanation. I think, I'll wait for the first pr to come in and decide than if it's worth the effort. :-)

domcleal:
I've gone back over prprocessor's code, it should be fine to add a repo without a Redmine project now.

PR #254, opened 2016-10-27
extend PATH for pbuilder

mmoll:
otherwise the rake binstub in /usr/local/bin is not found on Debian 8 or newer

### DIFF
--- a/puppet/modules/debian/manifests/init.pp
+++ b/puppet/modules/debian/manifests/init.pp
@@ -131,6 +131,13 @@ class debian {
     }
   }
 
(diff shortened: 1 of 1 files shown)


domcleal:
Does PATH need to be exported rather than just set in the current shell? The original in /usr/share exports it.

mmoll:
It seems these variables get exported globally in the pbuilder chroot environment. I had these line added on my Ubuntu 16.04 builders to fix the builds since the begnning and the sole addition fixed it.

domcleal:
Ok, thanks @mmoll!

PR #255, opened 2016-10-27
disable armhf builds on Ubuntu/trusty

mmoll:
As http://projects.theforeman.org/issues/15827 seems not to be solveable. Once this is merged, I'll open a PR to the docs and inform foreman-users.


domcleal:
Thanks @mmoll.

PR #256, opened 2016-11-08
removed ruby 1.9.3 and puppet 2.7.0 and 3.0.0 from test matrix

witlessbird:


witlessbird:
I only just realised I don't know how test_proxy_develop_pr_core is being generated. Does it need a similar change?

domcleal:
Can you also remove 1.9.3, Puppet 2 and 3.[0-1] from `combination-filter` too please?

> I only just realised I don't know how test_proxy_develop_pr_core is being generated. Does it need a similar change?

Yeah, it's manually edited at the moment - I can copy this change on merge.

witlessbird:
Removed ruby 1.9.3 and older puppet versions from the filter.

domcleal:
Thanks @witlessbird, it should update within an hour or two.

PR #259, opened 2016-11-23
Allow core dependency plugin testing

iNecas:
Scan first thought the Gemfile in case the plugin gem is already
mentioned somewhere, as bundler doesn't like when something is mentioned twice.

Preparation for https://github.com/theforeman/foreman/pull/3670


iNecas:
Anyone for review? Without this, we are blocked on foreman-tasks PRs

domcleal:
Merged, thanks @iNecas.

PR #262, opened 2016-12-11
lower the jenkins slave disk cleanup threshold

mmoll:


domcleal:
Why reduce it?

mmoll:
Two nodes were taken offline over the weekend because of this job (and somehow failed to come online again?), on one I did extend the filesystem, the other is just a DEB package build slave and already has allocated all disk space included the extremly cheap cloud instance offer. A look into the monitoring reveals that are no disk space problems, despite only 10GB being free:

disk

Maybe the script should be modified to use some percentage as threshold, but this is a quick fix for now.

domcleal:
OK - merged, thanks.

PR #264, opened 2016-12-21
enable testing for monitoring plugins

timogoebel:


domcleal:
Merged, thanks @timogoebel. It should be there in around an hour.

timogoebel:
@domcleal : Thanks. IIRC one also needs to add hooks to the repos. Could you do that for me, please? :-)
While you're at it, please also disable merge commits. Thanks!

domcleal:
Sure, should all be done.

PR #265, opened 2016-12-27
Update the branch we upgrade from to 1.13-stable

dLobatog:
1.11 bundle install has many incompatibilities, such as:
 gettext_i18n_rails-1.8.0, nokogiri 1.7.0
which do not work with Ruby 2.0.0.


dLobatog:
Alternatively we can pin deps on 1.11-stable directly

dLobatog:
Closed, pinned nokogiri on 1.11-stable instead to  < 1.7

dLobatog:
f

PR #266, opened 2017-01-02
Updated stable discovery version to 8.0

lzap:
I am shooting in the dark here.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
@@ -4,7 +4,7 @@
     branch:
       - develop:
           foreman_branch: develop
-      - 4.0-stable:
-          foreman_branch: 1.9-stable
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
```diff
@@ -4,7 +4,7 @@
     branch:
       - develop:
           foreman_branch: develop
-      - 4.0-stable:
-          foreman_branch: 1.9-stable
+      - 7.0-stable:
+          foreman_branch: 7.0-stable
```
mmoll: that should be 1.13(or 1.14?)-stable, I guess

lzap: Ah yeah fixing.

lzap:
Updated to 8.0 as I am releasing this today.

domcleal:
Merged, a new 8.0-stable testing job should be created within an hour or so.

PR #267, opened 2017-01-02
Send failed develop and stable branches to list

lzap:
Here is my proposal: Let's create an email list and configure Jenkins to
send failure emails for all core + plugins develop + stable branches
there. Developers can pick from there and filter what they want.

I am new to this YAML thing, I haven't tested this yet, so this needs
review and also I did not create any mailing list yet (the email there
is just a suggested name we could use).

I hope this will not include failures from PRs, that would totally ruin
the experience and that is not my intention at all.


ekohl:
Given we no longer have mailing lists but now discourse this exact implementation will no longer work. 
In https://community.theforeman.org/t/sending-jenkins-build-failures-to-discourse/9976 we agreed that at this point we're not ready yet so I'm closing this for now.

PR #269, opened 2017-01-12
Bump mysql to 3.10.0 to silence facter

GregSutcliffe:
This should fix the huge number of cron emails from the slaves, see https://github.com/puppetlabs/puppetlabs-mysql/pull/883 for details.


domcleal:
Merged, thanks @GregSutcliffe.

PR #270, opened 2017-01-16
Add Puppet PC1 repository to pbuilder

domcleal:
1. https://github.com/theforeman/foreman-packaging/issues/1491
1. this
1. https://github.com/theforeman/foreman-packaging/issues/1492


mmoll:
The PC1 and non-PC1 repos are enabled after this, I guess the selection of 3 vs. 4 will be puppet vs. puppet-agent?

domcleal:
Yes, that's right. Both will be needed as older branches will build against `puppet`.

mmoll:
allright, as puppet-agent doesn't seem to provide puppet, this should work...

mmoll:
merged, thanks @domcleal!

PR #271, opened 2017-01-19
Add name param to yum repo to make yum-cron happy

GregSutcliffe:
Hopefully prevents emails like this:

```
/etc/cron.hourly/0yum-hourly.cron:

Repository 'isimluk-openscap' is missing name in configuration, using id
```


mmoll:
Merged, thanks @GregSutcliffe!
Please report back, if that completely resolves the issue.

GregSutcliffe:
Sadly not. Looks like I misread the options for the `yumrepo` resource - according to https://docs.puppet.com/puppet/latest/types/yumrepo.html#yumrepo-attribute-descr we want to set `desc` not `name`. I'll send a new PR.

PR #273, opened 2017-01-26
Ensure Ansible is available on slaves

ehelms:


ehelms:
2.2 and i know they push out new each release to EPEL.

mmoll:
merged, thanks @ehelms!

If there should any Ansible version problems arise with Debian or Ubuntu slaves, please let me know.

PR #279, opened 2017-02-27
Create foreman_virt_who_configure master jenkins job

ares:


domcleal:
Merged, thanks @ares. It should be created within an hour or so.

PR #280, opened 2017-03-07
Update the foreman_virt_who_configure test job

ares:
It turned out that installing Katello breaks Foreman tests, so I can't use the normal plugin matrix. The test for this plugin can only run Katello test suite and it's own tests :-(

I hope I got it right, help would be appreciated since I'm not sure how jenkins job builder works exactly.


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_virt_who_configure.yaml
```diff
@@ -1,5 +1,28 @@
...
+      - string:
+          name: foreman_pr_git_ref
+          default:
+          description: 'Foreman Git ref for PR to test, usually the branch name'
     jobs:
       - 'test_plugin_{name}_{branch}'
+    builders:
+      - clone_foreman
+      - test_katello
+      - test_foreman_virt_who_configure
```
ehelms: Does foreman_virt_who link in to Katello enough to make running test_katello worth while?

ares: so far it does not modify Katello code directly (and I want to avoid it) but it requires Katello to live (e.g. organization#label) and introduces new SSO method that is only activated for [Katello::Api::Rhsm::CandlepinProxiesController](https://github.com/theforeman/foreman_virt_who_configure/blob/master/app/services/sso/basic_with_hidden.rb#L4)

I think it would be nice to know that this plugin does not break Katello. But if it's acceptable to add more code (e.g. monitoring incoming reports) to Katello code base and only use it if this plugin is installed (or make this plugin Katello dependency), maybe it's not worth it.

ehelms:
@ares OK - looks good to start with this.

ares:
Should this rebuild the job or do I need to delete it from jenkins first? Anyone knows? It seems it didn't change.

ehelms:
It should get auto updated when puppet runs.

ehelms:
@ares If needed, I can run update manually for you.

ares:
so something is wrong (probably on my side) since it still runs test_plugin_matrix with all 3 databases

PR #286, opened 2017-04-07
Added discovery 9.0 job

lzap:
Is it okay to keep N-1 stable version? @dLobato


domcleal:
Merged, thanks @lzap. If you no longer use the other job, please open a new PR and remove it.

lzap:
Thanks I would rather keep it until 10.0, we have some auto discovery issues I need to solve, they might be in 1.14 as well.

PR #288, opened 2017-04-10
Update release_plugins_test for 1.15

dLobatog:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_plugins_test.yaml
```diff
@@ -14,20 +14,19 @@
...
           name: os
           values:
-          - el6
           - el7
           - f19
           - f21
           - f24
     execution-strategy:
-      combination-filter: 'os == "el7" || os == "f24" || (os == "el6" && (version == "1.12"))'
+      combination-filter: 'os == "el7" || os == "f24"
```
domcleal: Syntax error, missing quotation mark.

domcleal:
Made obsolete by d621426.

PR #291, opened 2017-04-20
enable testing for puppetdb plugin

timogoebel:


domcleal:
Merged, thanks @timogoebel. Webhook added to the repository too.

PR #292, opened 2017-04-21
use updated npm for Debian/stretch

mmoll:


mmoll:
I use a lauchpad PPA and a Ubuntu/zesty builder, which is a gross hack, but works fine.

domcleal:
Is this intended to replace the `npm install npm` that's usually in the `rules`? And/or in case it's removed from stretch?

mmoll:
The problem is that the npm coming with stretch is too old for the npm modules used in Foreman.

domcleal:
Yeah, though other distros use the main distro version to install a newer one locally (https://github.com/theforeman/foreman-packaging/blob/deb/1.15/debian/jessie/foreman/rules#L15). So this PR is intending to replace that?

mmoll:
oh... hm... let me check

mmoll:
I totally forgot about this. That's much better! :)

mmoll:
Reopening, as npm is not available for Debian/stretch in the debian.org archives now.

domcleal:
Is it possible to build this in foreman-packaging rather than adding the PPA?

mmoll:
> Is it possible to build this in foreman-packaging rather than adding the PPA?

I think there's currently no place for such a package, as the "dependencies" folder is limited to ruby packages (is it?). Another alternative would be to use the nodesource "sid" repo (I'm leaning towards using the node 6.x repo for all Debian based OSes when it has stretch builds anyway).

domcleal:
Yes, I was looking at the nodesource packages as a source of npm too. I'd prefer to use them as it's more well-known than your PPA, but would you like this merged still in the meantime?

ohadlevy:
just another thought, would the npm package that exists in debian would allow to install yarn? yarn can replace npm easily and also aligns with rails 5.1 using yarn as asset management tooling.

mmoll:
@ohadlevy that's is a different discussion, as the Rails 5.1 things, also require Ruby 2.2+ and Node 6.something. Also, yarn would need to get installed first, which would in turn require npm, presumably. ;)

mmoll:
@domcleal Yes, I'd like to get this merged for now and when the nodesource stretch repos are ready, I'll propose to raise the Node version in the whole Foreman project to the lowest 6.x version possible (probably the one that's in EPEL).

domcleal:
Merged, thanks @mmoll.

PR #293, opened 2017-05-02
Adjusting to properly use RVM

evgeni:
See also https://github.com/SatelliteQE/robottelo-ci/pull/545

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/deploy_katello_site.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/deploy_katello_site.sh
@@ -1,9 +1,7 @@
-#!/bin/bash -xe
+#!/bin/bash -exl
 
 # Build the site on the slave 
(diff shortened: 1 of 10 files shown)


evgeni:
IMHO slightly cleaner as #282 ;)

zjhuntin:
@evgeni Agreed, I took my PR down.

mmoll:
I'm a bit concerned about this PR as the errors this PR is supposed to fix can't be seen on ci.theforeman.org, if I understood correctly...

evgeni:
Correct, this is to fix an issue we have on our downstream (Satellite) Jenkins, which does not have `rvm.sh` in `/etc/profile.d/`, albeit RVM is working fine otherwise.

mmoll:
I think this needs to get changed on the downstream CI side, sorry...

evgeni:
Fair enough, we have implemented a workaround in https://github.com/SatelliteQE/robottelo-ci/issues/635, so we'll keep it.

PR #294, opened 2017-05-08
Update $stable on web to 1.15

dLobatog:
To be merged after 1.15.0 is announced


dLobatog:
@domcleal I think this can be merged now, I'll send the announcement in a few moments

domcleal:
Merged, thanks.

PR #295, opened 2017-05-31
Pin npm to < 5.0.0

dLobatog:
5.0.0 is having some problems installing some of our dependencies 
as explained in http://projects.theforeman.org/issues/19717


mmoll:
hrm, should `puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh` also get updated?

dLobatog:
Indeed, thanks @mmoll !

mmoll:
merged, gracias @dLobatog!

PR #301, opened 2017-06-21
Puppet module update (theforeman modules)

mmoll:
this brings the theforeman modules up to the level of the 1.10 installer


ohadlevy:
thanks @mmoll

PR #303, opened 2017-06-22
update theforeman-* modules to 1.11 level

mmoll:
as there are no dhcp proxies, this should just work without setting parameters (OK, maybe for not configuring the logs module of the proxy, if it doesn't ignore unknown paramters...).


GregSutcliffe:
Matches https://github.com/theforeman/foreman-installer/blob/1.11-stable/Puppetfile and tests green. Pulling the trigger, thanks @mmoll!

PR #308, opened 2017-06-26
Make the users module puppet-lint clean

ekohl:
I've also removed some automatic requires file parent directories and users.

### DIFF
--- a/puppet/modules/users/manifests/account.pp
+++ b/puppet/modules/users/manifests/account.pp
@@ -1,50 +1,48 @@
 define users::account(
   $fullname,
-  $passwd = undef
+  $passwd = undef,
+  $homedir = "/home/${title}",
 ) {
   user { $name:
     ensure     => present,
     comment    => $fullname,
-    home       => "/home/$name",
+    home       => $homedir,
     managehome => true,
     shell      => '/bin/bash',
     password   => $passwd,
   }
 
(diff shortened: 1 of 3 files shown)


puppet/modules/users/manifests/account.pp
```diff
@@ -1,50 +1,48 @@
...
-  file { "/home/$name/.ssh/authorized_keys":
-    ensure  => present,
-    source  => "puppet:///modules/users/$name-authorized_keys",
-    owner   => $name,
-    group   => $name,
-    mode    => '0600',
-    require => File["/home/$name/.ssh"]
+  file { "${homedir}/.ssh/authorized_keys":
+    ensure => file,
+    source => "puppet:///modules/users/${name}-authorized_keys",
```
mmoll: is `$module_name` usage advised here? (it was in 3.x times)

ekohl: I was thinking about this and also considered switching to ```ssh_authorized_key``` together with ```purge_keys``` on the ```user```. That should allow easier extension and might also be faster since you don't need the puppet file server for a small file. This is just a first clean up.

mmoll: allright 👍

mmoll:
merged, bedankt @ekohl!

PR #309, opened 2017-06-27
cleanup old bower/grunt for JS tests

mmoll:
@ehelms @jlsherrill I can only speculate about the current state here...


ehelms:
I am pretty sure this all still current and its removal would break at least two unit test jobs for bastion and katello.

mmoll:
ok, thx!

PR #313, opened 2017-07-06
use psych gem with katello to avoid double loading

jlsherrill:


mmoll:
I also saw this as a workaround for that class name error that can be seen sometimes, let's try it 👍

PR #314, opened 2017-07-08
enable KeepAlive in Apache vhosts

mmoll:


puppet/modules/freight/templates/vhost.erb
```diff
@@ -1,3 +0,0 @@
-<Directory />
```
mmoll: @GregSutcliffe was the some reason to set this to `Directory /` instead of `Location /`?

GregSutcliffe: Honestly, I can't recall. Happy to change it, I can't see it causing an issue

GregSutcliffe:
Sorry for the delay. LGTM, I'll merge and we can monitor for issues.

PR #315, opened 2017-07-15
run puppet tests in release_nightly_* jobs

mmoll:


mmoll:
The breakage on jessie of https://github.com/theforeman/foreman-installer/pull/236 would have been visible earlier then. As puppet5 is only available for 1.16+, this is not feasible, yet.

ekohl:
Thanks!

PR #316, opened 2017-07-20
Update latest docs to 1.16-develop

tstrachota:
1.15 manual was copied to 1.16-develop in https://github.com/theforeman/theforeman.org/pull/921


GregSutcliffe:
According to https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/templates/web.conf.erb this *currently* redirects `/manuals/1.16` to `/manuals/1.15`. That kinda makes sense, but this change would redirect `/manuals/1.17` to `/maunals/1.16-develop/` which feels a bit weird.

Would it not make sense for `/manuals/1.16` to redirect to `/manuals/1.16-develop` instead?

GregSutcliffe:
Merged as 60116b7, ty for the reminder @tstrachota

PR #318, opened 2017-07-21
Update jobs for 1.16

dLobatog:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
```diff
@@ -30,7 +30,7 @@
           - trusty
           - xenial
     execution-strategy:
-      combination-filter: 'os == "el7" || os == "f24" || os == "jessie" || (os == "stretch" && major_version != "1.14" && major_version != "1.15") || (os == "trusty" && major_version == "1.15") || os == "xenial"'
+      combination-filter: 'os == "el7" || os == "jessie" || (os == "stretch" && major_version != "1.15" && major_version != "1.16") || (os == "trusty" && major_version == "1.16") || os == "xenial"'
```
GregSutcliffe: are we not building 1.15 for Trusy any more then? should this not be 1.15 && 1.16?

GregSutcliffe: Also, I thought stretch support was in 1.16... is that meant to be != 1.14 / 1.15?

GregSutcliffe:
Presumably this is waiting for the 1.16 branching?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
dLobatog: Correct on both, thanks for spotting it

mmoll:
@dLobatog did you forget to push here also?
Unfortunately Debian/stretch won't be testable (but should be built) until PL releases Puppetserver 5.x packages for it.

mmoll:
Trusty is no longer built for 1.16, Stretch should get built but not tested, yet

ehelms:
I feel like maybe we need a table somewhere that tracks this. I know per version of the docs there is (https://theforeman.org/manuals/1.15/index.html#3.1.1SupportedPlatforms) but I am thinking maybe a table of versions by OSes? Maybe we keep it here in the infra docs?

ekohl:
We also have https://github.com/theforeman/forklift/blob/master/config/versions.yaml which is essentially a subset. Ideally we have one ultimate source of truth somewhere. I don't care if we duplicate it in other places. Forklift might make sense for performance, but then I'd like a scripted way to update versions.yaml.

mmoll:
Just tell me, where and in which format I should put the table regarding Debian/Ubuntu versions :)

GregSutcliffe:
For docs, either the manual or the wiki is the place to go - this feels like a wiki thing we could put under the [Infrastructure](http://projects.theforeman.org/projects/foreman/wiki/Infrastructure) page?

ekohl:
What do you think about copying the puppet ```metadata.json``` format:
```json
{
  "1.15": {
    "katello": "3.4",
    "requirements": [
      {
        "name": "puppet",
        "version_requirement": ">= 4.6.1 < 5.0.0"
      }
    ],
    "operatingsystem_support": [
      {
        "operatingsystem": "RedHat",
        "operatingsystemrelease": [
          "7"
        ]
      },
      {
        "operatingsystem": "CentOS",
        "operatingsystemrelease": [
          "7"
        ]
      },
      {
        "operatingsystem": "Scientific",
        "operatingsystemrelease": [
          "7"
        ]
      },
      {
        "operatingsystem": "Fedora",
        "operatingsystemrelease": [
          "24"
        ]
      },
      {
        "operatingsystem": "Debian",
        "operatingsystemrelease": [
          "8"
        ]
      },
      {
        "operatingsystem": "Ubuntu",
        "operatingsystemrelease": [
          "16.04"
        ]
      }
    ]
  }
}
```

We could host this on ```https://theforeman.org/versions/foreman.json```. You could then make a Foreman plugin to check for new versions. Similarly you can also create a ```https://theforeman.org/versions/smart-proxy.json``` with the same format.

The included puppet is for the installer, but that might not be obvious.

An alternative to including katello in ```foreman.json``` we could also have a ```katello.json``` which has a key ```foreman``` instead.

GregSutcliffe:
I'm ok with some static json on the site...

mmoll:
One note here: PL didn't manage to publish puppetserver packages for Debian/stretch, so I'd just omit stretch here. The next release is scheduled for 26th of September, hopefuly we can get stretch going in nightly again and then also have it in RC2.

mmoll:
ok, merged, I'll revist the Debian/Ubuntu situation soonish

PR #319, opened 2017-07-30
add ARMv7 multiarch pbuilder support

mmoll:


mmoll:
This way, we can add ARMv7 hosts for ARMv7 build only and ARMv8 hosts for either ARMv7 or ARMv8. On x86, the i386 build still follows the arm64 build, while for ARMv8 an ARMv7 build would be triggered directly. This way the ARMv* builds can be parallelized, to speed up the build overall.

I'd merge the first commit, wait for the Puppet runs to finish and then the second to update the Jenkins jobs.

PR #321, opened 2017-08-21
Add pipeline based Jenkins job for Katello PRs

ehelms:
This is an attempt at re-writing test-katello as a pipeline using parallelism. The job can be browsed at [1]. This still needs the JS tests added to it. I wanted to open this up to get some thought around this approach as compared to a more linear approach. @mmoll @egolov  If you wouldn't mind lending some thoughts.

[1] http://ci.theforeman.org/blue/organizations/jenkins/test-katello/activity

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
@@ -1,0 +1,63 @@
+def databaseFile (id) {
+    writeFile(file: 'config/database.yml', text: """
+test:
+  adapter: postgresql
+  database: ${id}-test
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+
+development:
+  adapter: postgresql
+  database: ${id}-development
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+
+production:
+  adapter: postgresql
+  database: ${id}-development
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+""")
+
+}
+
+def addGem() {
+    writeFile(text: "gemspec :path => '../', :development_group => :dev", file: 'bundler.d/Gemfile.local.rb')
+}
+
+def addSettings(settings) {
+    sh "cp config/settings.yaml.example config/settings.yaml"
+
+    if (settings['locations']) {
+        sh "sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml"
+    }
+
+    if (settings['organizations']) {
+        sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
+    }
+}
+
+def configureDatabase(ruby) {
+    withRVM(['bundle update'], ruby)
+    withRVM(['bundle exec rake db:drop -q || true'], ruby)
+    withRVM(['bundle exec rake db:create -q'], ruby)
+    withRVM(['bundle exec rake db:migrate -q'], ruby)
+}
+
+def cleanup() {
+    try {
+
+        withRVM(['bundle exec rake db:drop DISABLE_DATABASE_ENVIRONMENT_CHECK=true'])
+
+    } finally {
+
+        cleanupRVM('', ruby)
+
+    }
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
@@ -9,8 +9,14 @@ def gemset(name) {
     base_name
 }
 
-def cleanup_rvm(name = '') {
-    withRVM(["rvm gemset delete ${gemset(name)} --force"])
+def configureRVM(ruby = '2.0') {
+    withRVM(['rvm gemset empty --force'], ruby)
+    withRVM(['gem update --no-ri --no-rdoc'], ruby)
+    withRVM(['gem install bundler'], ruby)
+}
+
+def cleanupRVM(name = '', ruby = '2.0') {
+    withRVM(["rvm gemset delete ${gemset(name)} --force"], ruby)
 }
 
 def withRVM(commands, ruby = '2.0', name = '') {
@@ -20,7 +26,6 @@ def withRVM(commands, ruby = '2.0', name = '') {
 
     sh """#!/bin/bash -l
         rvm use ruby-${ruby}@${gemset(name)} --create
-        gem install bundler
         ${commands}
     """
 }
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
@@ -1,0 +1,141 @@
+def ruby = '2.4'
+def katello_versions = [
+    'master': 'develop',
+    'KATELLO-3.6': '1.17-stable',
+    'KATELLO-3.5': '1.16-stable'
+]
+
+pipeline {
+    agent { label 'fast' }
+
+    stages {
+        stage('Setup Git Repos') {
+            steps {
+               checkout changelog: true, poll: false, scm: [
+                   $class: 'GitSCM',
+                   branches: [[name: '${ghprbActualCommit}']],
+                   doGenerateSubmoduleConfigurations: false,
+                   extensions: [[$class: 'PreBuildMerge', options: [fastForwardMode: 'FF', mergeRemote: 'origin', mergeStrategy: 'default', mergeTarget: '${ghprbTargetBranch}']]],
+                   userRemoteConfigs: [
+                       [refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*', url: 'https://github.com/katello/katello']
+                   ]
+               ]
+
+                dir('foreman') {
+                    git url: "https://github.com/theforeman/foreman", branch: katello_versions[ghprbTargetBranch]
+                }
+            }
+        }
+        stage("Setup RVM") {
+            steps {
+
+                configureRVM(ruby)
+
+            }
+        }
+        stage('Configure Environment') {
+            steps {
+
+                dir('foreman') {
+                    addSettings([
+                        organizations: true,
+                        locations: true
+                    ])
+                    addGem()
+                    databaseFile(gemset())
+                }
+
+            }
+        }
+        stage('Configure Database') {
+            steps {
+
+                dir('foreman') {
+                    configureDatabase(ruby)
+                }
+
+            }
+        }
+        stage('Run Tests') {
+            parallel {
+                stage('tests') {
+                    steps {
+                        dir('foreman') {
+                            withRVM(['bundle exec rake jenkins:katello TESTOPTS="-v"'], ruby)
+                        }
+                    }
+                }
+                stage('rubocop') {
+                    steps {
+                        dir('foreman') {
+                            withRVM(['bundle exec rake katello:rubocop TESTOPTS="-v" --trace'], ruby)
+                        }
+                    }
+                }
+                stage('react-ui') {
+                    when {
+                        expression { fileExists('package.json') }
+                    }
+                    steps {
+                        sh 'npm install'
+                        sh 'npm test'
+                    }
+                }
+                stage('angular-ui') {
+                    steps {
+                        dir('foreman') {
+                            withRVM(['bundle show bastion > bastion-version'], ruby)
+
+                            script {
+                                bastion_install = readFile('bastion-version')
+                                bastion_version = bastion_install.split('bastion-')[1]
+                                echo bastion_install
+                                echo bastion_version
+                            }
+                        }
+
+                        sh "cp -rf \$(cat foreman/bastion-version) engines/bastion_katello/bastion-${bastion_version}"
+
+                        dir('engines/bastion_katello') {
+                            sh "npm install npm"
+                            sh "node_modules/.bin/npm install bastion-${bastion_version}"
+                            sh "grunt ci"
+                        }
+                    }
+                }
+                stage('assets-precompile') {
+                    steps {
+                        dir('foreman') {
+                            withRVM(['bundle exec rake plugin:assets:precompile[katello] RAILS_ENV=production --trace'], ruby)
+                        }
+                    }
+                }
+            }
+        }
+        stage('Test db:seed') {
+            steps {
+
+                dir('foreman') {
+
+                    withRVM(['bundle exec rake db:drop'], ruby)
+                    withRVM(['bundle exec rake db:create'], ruby)
+                    withRVM(['bundle exec rake db:migrate'], ruby)
+                    withRVM(['bundle exec rake db:seed'], ruby)
+
+                }
+
+            }
+        }
+    }
+
+    post {
+        always {
+            dir('foreman') {
+                archiveArtifacts artifacts: "Gemfile.lock"
+                junit keepLongStdio: true, testResults: 'jenkins/reports/unit/*.xml'
+                cleanup()
+            }
+            deleteDir()
+        }
+    }
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/generate_jobs.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/generate_jobs.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-jenkins-jobs -l debug test -r . -o /tmp/jobs
+jenkins-jobs -l debug test -r -o /tmp/jobs yaml "$@"
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-pr-test.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-pr-test.yaml
@@ -1,0 +1,18 @@
+- job:
+    name: katello-pr-test
+    project-type: workflow
+    concurrent: true
+    logrotate:
+      daysToKeep: 30
+      numToKeep: -1
+    properties:
+      - github:
+          url: https://github.com/Katello/katello
+    triggers:
+      - github_pr:
+          context: 'katello'
+    dsl:
+      !include-raw:
+        - pipelines/test/testKatello.groovy
+        - pipelines/lib/rvm.groovy
+        - pipelines/lib/foreman.groovy


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
```diff
@@ -0,0 +1,135 @@
...
+            'gem update --no-ri --no-rdoc',
+            'gem install bundler --no-ri --no-rdoc'
+        ])
+
+    }
+
+    stage('Configure Environment') {
+
+        try {
+            sh "cp config/settings.yaml.example config/settings.yaml"
```
sean797: I would prefer if this step was, created `settings.yaml` like this, defining it in here allows any changes to tests to be made a PR and other obvious benefits 

Same thing with `database.yml`

```
cat << EOF 

foo
bar

EOF

Comment (ehelms) [inline puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy]: Sorry, I don't entirely follow what you are asking for. This step does create `settings.yaml` via the copy.

Comment (sean797) [inline puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy]: Sorry, it's not actually relevant on this line 👍

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy [hunk 2]
```diff
@@ -0,0 +1,135 @@
...
+            sh "sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml"
+            sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
+
+            sh "cp $HOME/postgresql.db.yaml config/database.yml"
```
sean797: I meant this line @ehelms. Copying from $HOME enables changes to `$HOME/postgresql.db.yaml` without raising a PR or review. I think it would be better to instead define the contents of `config/database.yml` in here.

ehelms: Yes and no. This technically comes from -- https://github.com/theforeman/foreman-infra/blob/2b425b40fca1f790f0b25ae7a8811485d26f11c0/puppet/modules/slave/files/db/postgresql which means its managed by this repo.

ehelms: We can try do better going forward, and re-think this pattern.

sean797: Okay, I guess it's not different than currently :-)

ehelms:
For anyone curious, I have updated the testKatello pipeline to use the declarative syntax style [1] which is newer and testForeman still uses the older style to provide some contrast between the two. After re-writing to the declarative style, I think it is the way to go.

[1] https://jenkins.io/doc/book/pipeline/syntax/#declarative-pipeline

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
ehelms: Because, the part below this is I think entirely ugly and something like ERB or some ansible would do wonders.

ekohl:
You test everything in parallel on different rubies and different databases. Does it also setup for every test or do you only run the setup once? If so, how do you know whether to use postgresql, mysql or sqlite?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
```diff
@@ -0,0 +1,135 @@
...
+        ${commands}
+    """
+}
+
+def cleanup() {
+    stage('Cleanup') {
+        dir('foreman') {
+            try {
+
+                sh "rm -rf node_modules/"
```
timogoebel: Is this necessary? Every job should run in its own temp dir as far as I know. I believe, there is no "workspace" for Pipeline Jobs anymore.

evgeni: at least in Jenkins 1.6.x style workflow there is, not sure about 2.x

ehelms: There are still workspaces, but you are correct @timogoebel I am still playing a bit with parallelism and isolation and am finding I need to do more workspace locking or creativity with how database configurations are deployed.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    stage('Configure Database') {
+
+        try {
+
+            withRVM([
+                'bundle update',
+                'bundle exec rake db:drop db:create -q',
+                'bundle exec rake db:migrate -q'
```
evgeni: One thing that came up when talking to @ShimShtein earlier today, can we also run `bundle exec rake db:seed` here? Quick test suggests yes.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
```diff
@@ -0,0 +1,135 @@
...
+            ])
+
+        } catch (all) {
```
evgeni: Should this be a `finally`? Otherwise you are catching errors, but not raising them. Thus the job would end w/o FAILED?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
evgeni: I added it to the old jobs here: https://github.com/theforeman/foreman-infra/pull/325

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testForeman.groovy
```diff
@@ -0,0 +1,135 @@
...
+def gemset() {
+    "${JOB_NAME}-${BUILD_ID}"
+}
+
+def withRVM(commands) {
+
+    commands = commands.join("\n")
+
+    sh """#!/bin/bash -l
+        rvm use ruby-2.2@${gemset()} --create
```
jlsherrill: maybe i'm misunderstanding, but it looks like the current test_develop tests with different rubies, but this is only testing with 2.2, intentional?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/tests/testKatello.groovy
```diff
@@ -0,0 +1,126 @@
...
+
+                    withRVM(['bundle update'], '2.4')
+                    withRVM(['bundle exec rake db:create -q'], '2.4')
+                    withRVM(['bundle exec rake db:migrate -q'], '2.4')
+
+                }
+
+            }
+        }
+        stage('Run Tests') {
```
jlsherrill: Now that rails 5 is merged, can we add rubocop and javascript testing to this?

ehelms:
I have updated this to a new state. The goal here is to re-write the existing Katello PR test fully as a Jenkins pipeline, splitting out the various testing concerns in parallel actions. An example of this can be seen at [1].

@evgeni The db:seed testing fails currently for me. I know you did some testing when you originally added this. Can you check that section to see if I am missing something?

@jlsherrill @sean797 If you'd like to re-review the re-focus please do.

I am considering this as the job to set the standard for future jobs for Foreman and other plugins. I'd like to get a sense of whether the structure makes sense, whether the parallel split makes sense and any overall improvements we could establish here as best practices.

[1] http://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test-test/detail/katello-pr-test-test/21/pipeline

timogoebel:
This is some WIP code to demonstrate, how this could look using a docker container: https://gist.github.com/timogoebel/a535c7ec095ee43d6c075b019ab09206

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+        stage('Test db:seed') {
+            steps {
+
+                dir('foreman') {
+
+                    withRVM(['bundle exec rake db:drop db:create'], ruby)
+                    withRVM(['bundle exec rake db:migrate'], ruby)
+                    withRMV(['bundle exec rake db:seed'], ruby)
```
evgeni: @ehelms if you write this as `withRVM` instead of `withRMV`, the `db:seed` tests will also pass :laughing:

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+pipeline {
+    agent { label 'fast' }
+
+    stages {
+        stage('Setup Git Repos') {
+            steps {
+                deleteDir()
+
+                sh "git config --global user.email 'ci@theforeman.org'"
+                sh "git config --global user.name 'jenkins'"
```
sean797: Is stuff like these 2 lines better put in Puppet instead?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -20,7 +20,6 @@ def withRVM(commands, ruby = '2.0', name = '') {
 
     sh """#!/bin/bash -l
         rvm use ruby-${ruby}@${gemset(name)} --create
-        gem install bundler
```
evgeni: why moving this out?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+                    sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
+
+                    sh 'echo "gemspec :path => \'../\', :development_group => :katello_dev" >> bundler.d/Gemfile.local.rb'
+                    sh 'echo "gem \'psych\'" >> bundler.d/Gemfile.local.rb'
+
+                    sh "cp $HOME/postgresql.db.yaml config/database.yml"
+
+                    sh "sed -i \"s/database:.*/database: ${gemset()}-test/\" config/database.yml"
+                    sh """
+cat <<EOT >> config/database.yml
```
sean797: I wounder if changing these from `sh` shells to use `dir`, `writeFile`, ect.. would have an impact on running time ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -9,7 +9,7 @@ def gemset(name) {
     base_name
 }
 
-def cleanup_rvm(name = '') {
+def cleanup_rvm(name = '', ruby = '2.0') {
```
evgeni: the ruby version is not used?

evgeni:
@timogoebel what are the benefits of the container solution? (besides getting a fresh db server each time)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+            deleteDir()
+        }
+    }
+}
+
+def cleanup() {
+    dir('foreman') {
+        try {
+
+            sh "rm -rf node_modules/"
```
sean797: Won't this happen with `deleteDir()` above?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+                    branches: [[name: '${ghprbActualCommit}']],
+                    doGenerateSubmoduleConfigurations: false,
+                    extensions: [[$class: 'PreBuildMerge', options: [fastForwardMode: 'FF', mergeRemote: 'origin', mergeStrategy: 'default', mergeTarget: '${ghprbTargetBranch}']]],
+                    userRemoteConfigs: [
+                        [refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*', url: 'https://github.com/katello/katello']
+                    ]
+                ]
+
+                dir('foreman') {
+                    git url: "https://github.com/theforeman/foreman", branch: "develop"
```
ekohl: Would we need to map the target branch to some foreman branch?

ekohl:
> This is some WIP code to demonstrate, how this could look using a docker container: https://gist.github.com/timogoebel/a535c7ec095ee43d6c075b019ab09206

That's interesting. I talked to some users at devconf who could sponsor containers in their Kubernetes cluster. Would we be able to use that with this style?

timogoebel:
@evgeni: Basically, you always have a fresh environment for the tests and no hassle with rvm. Your jenkins does not need to be set up in a special way to run tests, it just needs to be able to start a container.

ShimShtein:
@evgeni, @timogoebel I actually like the concept of containerized tests - it will make plugin tests way easier on local machine.

evgeni:
@ShimShtein I agree that properly containerized tests would be cool. but at the moment it still would be "take a container and run this massive shell script inside", which is not really helping, imho.

ehelms:
@timogoebel I do agree with you container usage in these pipelines is the way of the future. I had been avoiding it mostly to avoid figuring out which of our slaves can run docker, and configuring that. For all I know, they already work. Part of this update is that I do owe Katello an update to their PR testing given they are straddling two worlds and it's a bit of a mess since Rails 5.1. Could you point me at the definition of `tfm-ci` so I can see what you are using as the base container?

@ekohl I think using an external Kube cluster to run your containers on is possible but is more advanced than the native slave spinning the container up.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
ehelms: I thought they were from my prior research, but they weren't working:

https://github.com/theforeman/foreman-infra/blob/7b7d83e2a75fa46879ab52806c30bc84cb27e909/puppet/modules/slave/files/gitconfig

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+
+pipeline {
+    agent { label 'fast' }
+
+    stages {
+        stage('Setup Git Repos') {
+            steps {
+                deleteDir()
+
+                sh "git config --global user.email 'ci@theforeman.org'"
```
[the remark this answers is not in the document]
ehelms: See https://github.com/theforeman/foreman-infra/pull/321#discussion_r165147278

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
ehelms: I knew you'd know the reason! :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,170 @@
...
+                sh "git config --global user.name 'jenkins'"
+
+                checkout changelog: true, poll: false, scm: [
```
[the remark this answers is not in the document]
ehelms: In the old way, yes, however, if we do that then Jenkins can't track the "changes" or give any information related to it because it does it based off of what the main workspace git repository is from everything I could reason.

timogoebel:
> Could you point me at the definition of tfm-ci so I can see what you are using as the base container?

@ehelms: Sure, I added it to the [gist](https://gist.github.com/timogoebel/a535c7ec095ee43d6c075b019ab09206).

I don't want to block this PR or anything, feel free to go ahead this way (but don't forget the other plugins, too 👼). I just thought this was a good place to share some stuff.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
ehelms: I moved this into a configureRVM script so it does it once. I was finding that in parallel, this can cause random failures.

ehelms:
@sean797 @evgeni @jlsherrill If y'all wouldn't mind doin another review, I'd like to get this rolled out to clean up the developer workflow for PRs to Katello

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,136 @@
...
+                        dir('foreman') {
+                            withRVM(['bundle exec rake jenkins:katello TESTOPTS="-v"'], ruby)
+                        }
+                    },
+                    'rubocop': {
+                        dir('foreman') {
+                            withRVM(['bundle exec rake katello:rubocop TESTOPTS="-v" --trace'], ruby)
+                        }
+                    },
+                    'react-ui': {
```
evgeni: can we mark this step as skipped when there is no `package.json` available? (if that's not too complicated, as it's just a visual thing I'd love to have)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    if (settings['organizations']) {
+        sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
+    }
+}
+
+def configureDatabase(ruby) {
+    withRVM(['bundle update'], ruby)
+    withRVM(['bundle exec rake db:create -q'], ruby)
```
evgeni: can we *try* to drop the db before creating it? like it was done in https://github.com/theforeman/foreman-infra/pull/436

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,136 @@
...
+                )
+
+            }
+        }
+        stage('Test db:seed') {
+            steps {
+
+                dir('foreman') {
+
+                    withRVM(['bundle exec rake db:drop db:create'], ruby)
```
evgeni: should we drop and create in two steps like in https://github.com/theforeman/foreman-infra/pull/438?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,136 @@
...
+
+                        sh "cp -rf \$(cat foreman/bastion-version) engines/bastion_katello/bastion-${bastion_version}"
+
+                        dir('engines/bastion_katello') {
+                            sh "npm install npm"
+                            sh "node_modules/.bin/npm install bastion-${bastion_version}"
+                            sh "grunt ci"
+                        }
+                    },
+                    'assets-precompile': { dir('foreman') {
```
evgeni: `\n` after `{` please ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -9,8 +9,14 @@ def gemset(name) {
     base_name
 }
 
-def cleanup_rvm(name = '') {
-    withRVM(["rvm gemset delete ${gemset(name)} --force"])
+def configureRVM(ruby) {
```
evgeni: should `ruby` have a default value? (and should that be bumped to something newer than `2.0`?)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,141 @@
...
+            steps {
+
+                dir('foreman') {
+                    configureDatabase(ruby)
+                }
+
+            }
+        }
+        stage('Run Tests') {
+            parallel {
```
ehelms: @evgeni Thanks for your question in this area it helped me fine this syntax which allows for https://github.com/theforeman/foreman-infra/pull/321/files#diff-912a62e12e77edab1326c08754e1b7e9R77

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -0,0 +1,63 @@
...
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+""")
+
+}
+
+def addGem() {
+    sh 'echo "gemspec :path => \'../\', :development_group => :dev" >> bundler.d/Gemfile.local.rb'
```
evgeni: Could this be a `writeFile` to `bundler.d/plugin.local.rb` (or whatever else name), so that we don't need an ugly `>>` to append stuff to an existing file?

ehelms:
Updated to use writeFile for addGem function

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -0,0 +1,63 @@
...
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+""")
+
+}
+
+def addGem() {
+    writeFile(text: 'echo "gemspec :path => \'../\', :development_group => :dev"', file: 'bundler.d/Gemfile.local.rb')
```
evgeni: I think the `echo` is not needed anymore ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -0,0 +1,141 @@
...
+
+                            script {
+                                bastion_install = readFile('bastion-version')
+                                bastion_version = bastion_install.split('bastion-')[1]
+                                echo bastion_install
+                                echo bastion_version
+                            }
+                        }
+
+                        sh "cp -rf \$(cat foreman/bastion-version) engines/bastion_katello/bastion-${bastion_version}"
```
evgeni: can this be `sh "cp -rf ${bastion_install} engines/bastion_katello/bastion-${bastion_version}"`?

evgeni: or even `sh "cp -rf ${bastion_install} engines/bastion_katello/${bastion_install}"`?

evgeni:
I left two inline comments, and with them addressed I'd say this is good to go from a "that's a nice pipeline" PoV. I'd like someone with more @Katello knowledge than myself to comment/ack the individual test steps, though.

@ehelms the next step (after this is merged) would be to make it more generic to also accommodate other plugins?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
ehelms: I wanted it to be. But for some reason it seems like the fact that there are special characters in the path such as `@` throws off this variable substitution causing groovy to not create the right string :/ I tried 4-5 different ways to avoid this intermediate file.

ehelms:
@evgeni Aye -- I like to use the most complex case first to iron out issues (plus I owe Katello finishng what I started with Rails 5) to then start iterating this for other plugins and core. There are a few things I'd also like to explore post this:

 * using github notification plugin to set status on the PR for each parallel job (e.g. allowing PR to see a status for rubocop). I think that part might be easy. However, I think developers will think they can trigger the failed parts which is the tricky part. For example, if rubocop failed I think there will be an inclination to do [test rubocop].
 * use of docker

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
evgeni: Uh, well, that explains it. Then let's go with the ugly :cat:

jlsherrill:
from a brief read over, it LGTM

PR #325, opened 2017-09-12
run rake db:seed during tests

evgeni:


ekohl:
Do (unit) tests need seeds? I had hoped that unit tests would have factories that took care of all the needed data. Would it be better to run unit tests and *then* seed so you test both?

evgeni:
The tests run fine w/o the seed. I just want to make sure that the seed scripts actually work too. So yeah, running seed after the tests should also work.

GregSutcliffe:
+1 for the concept, these should be tested. After feels good, I agree.

evgeni:
Updated to run a full cycle of `db:drop db:create db:migrate db:seed` *after* the tests.

ekohl:
One the one hand I agree this is the best option but I wonder about the additional overhead of running all migrations again. Maybe @mmoll or @ehelms know better how much our build slaves would suffer.

mmoll:
it seems to be well under 30 seconds, so I think this is OK 👍

evgeni:
Had a talk with @ehelms in the "office" yesterday about this.

He pointed out that at least Foreman (https://github.com/theforeman/foreman/blob/develop/test/unit/tasks/seeds_test.rb) and Katello (https://github.com/Katello/katello/blob/master/test/lib/tasks/seeds_test.rb) have seed tests, but other plugins usually do not.

We also discussed that it might be interesting to run the seeds twice, just to see that they actually work on both, a clean and a "dirty" db.

mmoll:
This is ready to roll, correct?

mmoll:
merged, спасибо Женя!

PR #326, opened 2017-09-18
Remove skottler key as per dev-list thread

GregSutcliffe:
Just cleanup I think - I don't see Sam's user in the Foreman YAML ENC...


ekohl:
Shouldn't there be a ```user { 'samkottler': ensure => absent }``` as well?

mmoll:
it's in `puppet/modules/users/manifests/init.pp` and should be changed there, too

mmoll:
hrm, or rather added with absent... It's strange that he's not in the YAML, as $users is probably not undef.

ohadlevy:
I suspect that in order to actually remove it, you need to change the
internal foreman ENC output. (change the parameter value).

mmoll:
@GregSutcliffe were there insights on how to remove the user?

ekohl:
@mmoll shouldn't we just maintain the user in the ENC but with ```ensure => absent```?

mmoll:
`users:account` always set `ensure     => present`

ekohl:
https://github.com/theforeman/foreman-infra/pull/332 should solve that but I haven't tested it.

mmoll:
mergedm thanks @GregSutcliffe!

PR #328, opened 2017-09-21
create Jenkins job for foreman-tasks 0.9.x

mmoll:


mmoll:
@iNecas As 1.15 seems be kind of an LTS release... 😁
http://ci.theforeman.org/job/test_plugin_foreman-tasks_0.9-stable/ is failing currently

mmoll:
BTW, is there also a different version of remote_execution for 1.15? I'd also add a job for it then.

iNecas:
Thanks, that's great. I've also triggered the tests against 1.15 in https://github.com/theforeman/foreman-tasks/pull/285#issuecomment-331380218 .
The remote-execution should be compatible though nightlies back to 1.15 without issues. The main reason we diverged in tasks was move of dynflow into core in 1.16

mmoll:
Ah, OK. Unfortunately it seems there's no easy way to have one branch (master in that case) of a plugin against multiple Foreman core versions, which would help for rex.

mmoll:
it's http://ci.theforeman.org/job/test_plugin_foreman-tasks_0.9.x/ now.

PR #329, opened 2017-09-22
Update plugins test job to include 1.16

dLobatog:


mmoll:
merged, gracias @dLobatog!

PR #330, opened 2017-09-23
Drop Ruby 2.1 support in nightly

mmoll:


mmoll:
Rails 5.0 requires Ruby >= 2.2

mmoll:
@ekohl could you have a good look on the combination filters?

The support matrix for Debian/Ubuntu should be:


x | 1.15 | 1.16 | nightly
-- | -- | -- | --
Ubuntu/trusty | x |  |  
Debian/jessie | x | x |
Ubuntu/xenial | x | x | x
Debian/stretch | | x | x

Also, I sneaked in a change to run Debian/stretch systest with PL nightlies, as puppetserver DEBs seem to exist in these repos now.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_coreproject.yaml
```diff
@@ -32,7 +32,7 @@
           values:
           - debian
     execution-strategy:
-      combination-filter: '((onlyarch == "all") || onlyarch == arch) && ((onlyos == "all") || onlyos == os) && !((version != "1.16" || (arch == "armv7" || arch == "armv8")) && os == "trusty") && !((version == "1.15" || version == "1.16") && os == "stretch")'
+      combination-filter: '((onlyarch == "all") || onlyarch == arch) && ((onlyos == "all") || onlyos == os) && !((version != "1.15" || (arch == "armv7" || arch == "armv8")) && os == "trusty") && !((version != "1.15" || version != "1.16") && os == "jessie") && !((version == "1.14" || version == "1.15") && os == "stretch")'
```
ekohl: Nit: the braces are a bit inconsistent. the ```"all"``` arch has them but ```arch``` doesn't. Same for onlyos.

I wonder if we can write this in multi line because for my understanding I rewrote it to:
```
(onlyarch == "all" || onlyarch == arch)
&& (onlyos == "all" || onlyos == os)
&& !(os == "trusty" && (version != "1.15" || (arch == "armv7" || arch == "armv8")))
&& !(os == "jessie" && (version != "1.15" || version != "1.16"))
&& !(os == "stretch" && (version == "1.14" || version == "1.15"))
```

ekohl:
Your matrix has no 1.14 but it's still mentioned for Stretch.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_coreproject.yaml
mmoll: I would need to test this, without tryiong I'd say no.

mmoll:
I guess the 1.14 condition could get deleted when we drop 1.14...

PR #337, opened 2017-10-04
Allow running Jenkins over HTTPS

ekohl:
Things to verify:

* [ ] Doesn't break existing web nodes (refactor should be safe)
* [ ] Doesn't break jenkins forwarding (left out the ```<Proxy *>```, not sure if that's needed)
* [ ] Letsencrypts .well-known directory might be forwarded to Jenkins as well causing cert requests to fail
* [ ] Enable SSL works and Jenkins is configured to work with it
* [ ] Do we want to disable plain text and move everything over?

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -11,11 +11,7 @@
 class web($stable = "1.15", $latest = "1.16", $next = "1.17", $htpasswds = {}, $https = false) {
   include apache
   include rsync::server
-
-  class { 'letsencrypt':
-    email          => 'foreman-infra-notifications@googlegroups.com',
-    configure_epel => false,
-  }
+  include web::letsencypt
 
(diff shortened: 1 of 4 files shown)


ekohl:
I'd like to roll this out somewhere during the day. If you spot anything that'd be greatly appreciated. The first commit shall also be the basis (combined with https://github.com/theforeman/foreman-infra/pull/340) to run redmine over HTTPS

mmoll:
I'll have a look and will deploy it in the (European) night when no jobs are running

ekohl:
I don't mind deploying it myself but you're welcome to deploy it. I do recall Jenkins needed some additional settings inside it to properly use HTTPS but just having the vhost enabled would be a great improvement.

mmoll:
merged with a small typo fix, unfortunately it doesn't work, see `/var/log/letsencrypt/letsencrypt.log` on the Jenkins server.

PR #338, opened 2017-10-10
Add AWStats module for web02

GregSutcliffe:
This has already been applied to Web02 using "puppet apply" while developing it, so it should be fine. Resulting output is viewable at http://downloads.theforeman.org/awstats

### DIFF
--- a/puppet/modules/awstats/manifests/init.pp
+++ b/puppet/modules/awstats/manifests/init.pp
@@ -1,0 +1,38 @@
+class awstats (
+  $vhosts = [
+    'all',
+    'deb',
+    'debugs',
+    'downloads',
+    'stagingdeb',
+    'web',
+    'yum'
+  ]
+) {
+
+  package { 'awstats': ensure => present }
+
+  # Use a cron per vhost instead of the one shipped in the package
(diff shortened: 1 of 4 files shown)


GregSutcliffe:
That's a fair point. I can move it to the debugs vhost which is password protected, would that do?

ekohl:
Sounds like a good idea

GregSutcliffe:
@ekohl updated and applied on web02

ohadlevy:
Is it just me or the links are broken?

GregSutcliffe:
@ohadlevy example? the menu the top works for me, drilldown subpages like awstats.deb.alldomains.html are not currently being generated.

ohadlevy:
I mean the vhost file (Deb, yum etc) in the link you sent earlier

GregSutcliffe:
As per @ekohl's suggestion to protect them, they're now at http://debugs.theforeman.org/awstats

GregSutcliffe:
Oh, i do see a typo in the puppet code that was leftover, thanks @ohadlevy - updated.

mmoll:
merged, thanks @GregSutcliffe!

PR #340, opened 2017-10-11
Initial import of Redmine config module

GregSutcliffe:
Really hacky initial commit for a Redmine module. Mostly compiled while working on the migration to Scaleway today. Lots we can improve, especially in passing parameters from Foreman, but at least it's a start.

### DIFF
--- a/puppet/modules/redmine/manifests/init.pp
+++ b/puppet/modules/redmine/manifests/init.pp
@@ -63,9 +63,6 @@ class redmine {
 
   # Apache / Passenger
 
-  class { 'apache':
-    default_vhost => false,
-  }
   include ::apache::mod::headers
   include ::apache::mod::passenger
 
@@ -74,7 +71,7 @@ class redmine {
   $docroot          = "${app_root}/public"
   $min_instances    = 1
   $start_timeout    = 600
-  $priority         = 30
+  $priority         = '05'
 
   apache::vhost { $servername:
     add_default_charset     => 'UTF-8',
--- a/puppet/modules/redmine/manifests/init.pp
+++ b/puppet/modules/redmine/manifests/init.pp
@@ -120,4 +120,13 @@ class redmine {
 sudo -u ${db_username} /usr/local/bin/redmine_repos.sh"
   }
 
+  # Logrotate
+  file { '/etc/logrotate.d/redmine':
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    content => template('redmine/logrotate.erb'),
+  }
+
 }
--- a/puppet/modules/redmine/manifests/init.pp
+++ b/puppet/modules/redmine/manifests/init.pp
@@ -1,0 +1,123 @@
+# Hacky class to set up Redmine
+# Needs VCS handling for cloning and bundling Redmine itself
+#
+class redmine {
+  $app_root    = '/usr/share/redmine'
+  $db_name     = 'redmine4'
+  $db_username = 'adminpz8bn8d'
+  $db_password = cache_data('foreman_cache_data', 'db_password', random_password(32))
+  $password    = postgresql_password($db_username, $db_password)
+
+  # Prevents errors if run from /root etc.
+  Postgresql_psql {
+    cwd => '/',
+  }
+
+  include ::postgresql::client, ::postgresql::server
+
+  postgresql::server::db { $db_name:
+    user     => $db_username,
+    password => $password,
+    owner    => $db_username,
+    encoding => 'utf8',
+    locale   => 'en_US.utf8',
+  }
+
+  # Create ident user
+  user { $db_username:
+    ensure  => 'present',
+    shell   => '/bin/false',
+    comment => 'Redmine',
+    home    => $app_root
+  }
+
+  # TODO: handle cloning the redmine repo and bundle install...
+  file { $app_root:
+    ensure => directory,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755'
+  }
+
+  file { "${app_root}/config/database.yml":
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    content => template('redmine/database.yml.erb')
+  }
+
+  # Needed for bundle install
+  $packages = [
+    'rubygem-bundler.noarch',
+    'ruby-devel',
+    'gcc',
+    'gcc-c++',
+    'libxml2-devel',
+    'ImageMagick-devel',
+    'postgresql-devel',
+    'sqlite-devel'
+  ]
+
+  ensure_packages($packages)
+
+  # Apache / Passenger
+
+  class { 'apache':
+    default_vhost => false,
+  }
+  include ::apache::mod::headers
+  include ::apache::mod::passenger
+
+  $servername       = 'projects.theforeman.org'
+  $redmine_url      = "http://${servername}/"
+  $docroot          = "${app_root}/public"
+  $min_instances    = 1
+  $start_timeout    = 600
+  $priority         = 30
+
+  apache::vhost { $servername:
+    add_default_charset     => 'UTF-8',
+    docroot                 => $docroot,
+    manage_docroot          => false,
+    port                    => 80,
+    options                 => ['SymLinksIfOwnerMatch'],
+    passenger_app_root      => $app_root,
+    passenger_min_instances => $min_instances,
+    passenger_start_timeout => $start_timeout,
+    priority                => $priority,
+    servername              => $servername,
+  }
+
+  file { ["${app_root}/config.ru", "${app_root}/config/environment.rb"]:
+    owner => $db_username,
+  }
+
+  # cron jobs ported from .openshift
+
+  file { '/etc/cron.daily/redmine_backup':
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => template('redmine/postgresql_backup.sh')
+  }
+
+  file { '/usr/local/bin/redmine_repos.sh':
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => template('redmine/git_repos.sh')
+  }
+
+  file { '/etc/cron.hourly/redmine_repos':
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => "#!/bin/bash
+sudo -u ${db_username} /usr/local/bin/redmine_repos.sh"
+  }
+
+}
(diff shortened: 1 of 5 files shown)


puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  ]
+
+  ensure_packages($packages)
+
+  # Apache / Passenger
+
+  class { 'apache':
+    default_vhost => false,
+  }
+  include ::apache::mod::headers
```
ekohl: These includes should not be needed.

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  include ::apache::mod::passenger
+
+  $servername       = 'projects.theforeman.org'
+  $redmine_url      = "http://${servername}/"
+  $docroot          = "${app_root}/public"
+  $listen_interface = undef
```
ekohl: Let's leave this out.

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  $keepalive_onoff  = 'off'
+  $min_instances    = 1
+  $http_prestart    = false
```
ekohl: Any reason to include this and yet turn it off? :)

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  $suburi_parts = split($redmine_url, '/')
+  $suburi_parts_count = size($suburi_parts) - 1
+  if $suburi_parts_count >= 3 {
+    $suburi_without_slash = join(values_at($suburi_parts, ["3-${suburi_parts_count}"]), '/')
+    if $suburi_without_slash {
+      $suburi = "/${suburi_without_slash}"
+    }
+  }
+
+  apache::vhost { 'projects.theforeman.org':
```
ekohl: Can use $servername here.

puppet/modules/redmine/templates/postgresql_backup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
ekohl: Are these backups taken offsite?

puppet/modules/redmine/templates/postgresql_backup.sh
```diff
@@ -0,0 +1,10 @@
...
+# Backs up the OpenShift PostgreSQL database for this application
+ 
+FILENAME="/usr/share/redmine_data/redmine.backup.sql.gz"
+
+touch $FILENAME
+chown postgres $FILENAME
```
ekohl: This shouldn't be needed since the pipe runs as root.

ekohl:
Also wondering if this should be ```web::redmine``` similar to ```web::jenkins```.

puppet/modules/redmine/manifests/init.pp
GregSutcliffe: I guess we could set $priority < 15, otherwise the default vhost has to go. Is the passenger include enough to pull in the other two?

puppet/modules/redmine/manifests/init.pp
GregSutcliffe: Done

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+    default_vhost => false,
+  }
+  include ::apache::mod::headers
+  include ::apache::mod::passenger
+
+  $servername       = 'projects.theforeman.org'
+  $redmine_url      = "http://${servername}/"
+  $docroot          = "${app_root}/public"
+  $listen_interface = undef
+  $keepalive_onoff  = 'off'
```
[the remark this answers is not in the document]
GregSutcliffe: Done

puppet/modules/redmine/manifests/init.pp
GregSutcliffe: Copy paste fail

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  $min_instances    = 1
+  $http_prestart    = false
+  $start_timeout    = 600
+  $priority         = 30
+
+  $suburi_parts = split($redmine_url, '/')
```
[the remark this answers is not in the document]
GregSutcliffe: done

puppet/modules/redmine/manifests/init.pp
GregSutcliffe: done

puppet/modules/redmine/manifests/init.pp
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  # cron jobs ported from .openshift
+
+  file { '/etc/cron.daily/redmine_backup':
+    ensure  => present,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => template('redmine/postgresql_backup.sh')
```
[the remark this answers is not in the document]
GregSutcliffe: I was assuming we'd drop the .openshift dir once we're no longer on openshift, to make diffs from the upstream repo smaller.

puppet/modules/redmine/templates/postgresql_backup.sh
GregSutcliffe: Yes, the puppetmaster has Dirvish set up. I'll update it's config once this is live - see http://projects.theforeman.org/projects/foreman/wiki/Redmine#Moving-gears point 9

puppet/modules/redmine/templates/postgresql_backup.sh
GregSutcliffe: Root has no ident access to the DB, so you need another user. We could use the Redmine user, put `postgres` is guaranteed to exist anyway...

GregSutcliffe:
`web::jenkins` exists because we have a whole separate module to set up Jenkins, and then a small class to configure the vhost. Much of this class has nothing to with the vhost, so I don't think it should all be there.

I have no objection splitting it up into `redmine::*` and `web::redmine` for the vhost at some point though.

GregSutcliffe:
Two secrets things we should add as class parameters and set in Foreman:

[ ] secret token - https://github.com/theforeman/redmine/blob/master/config/configuration.yml#L157
[ ] email pass - https://github.com/theforeman/redmine/blob/master/config/configuration.yml#L206

These are currently hardcoded in `/usr/shar/redmine'

puppet/modules/redmine/templates/postgresql_backup.sh
ekohl: You can still sudo for pg_dump, just this chown line isn't needed I think.

puppet/modules/redmine/manifests/init.pp
ekohl: https://github.com/puppetlabs/puppetlabs-apache/blob/ca3747d91243c0843092bdcf927e35a73a94a93c/manifests/vhost.pp#L236-L238 means the passenger include is redundant. The headers module is used in the foreman to send expire headers but we don't use that here.

puppet/modules/redmine/manifests/init.pp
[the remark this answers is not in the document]
ekohl: I suppose that could be a good idea.

puppet/modules/redmine/templates/git_repos.sh
```diff
@@ -0,0 +1,53 @@
...
+
+cd /usr/share/redmine
+
+# Create repositories in the Redmine projects for all known git repos
+curl http://prprocessor-theforeman.rhcloud.com/redmine_repos | sudo -u <%= @db_username %> script/rails runner -e production '
+JSON.load(STDIN).each do |project_name,repos|
+  repos.each do |repo,branches|
+    org_name, repo_name = repo.split("/", 2)
+    repo_path = File.join('/usr/share/redmine_data', "git", "#{repo_name}.git") + File::SEPARATOR
+puts repo_path.inspect
```
ekohl: Should we leave out this debug info?

puppet/modules/redmine/templates/postgresql_backup.sh
```diff
@@ -0,0 +1,10 @@
...
+
+cd /tmp
+sudo -u postgres pg_dump redmine4 | gzip > $FILENAME
```
ekohl: If it's a template, we can use ```<%= @db_name %>```

puppet/modules/redmine/templates/git_repos.sh
GregSutcliffe: Oh jeez, sorry. I was trying to debug an error this puts out, but I'm unsure yet if the error actually impacts the updating

puppet/modules/redmine/templates/postgresql_backup.sh
GregSutcliffe: Oh right, I read the wrong thing. Done.

puppet/modules/redmine/templates/postgresql_backup.sh
GregSutcliffe: Done

puppet/modules/redmine/manifests/init.pp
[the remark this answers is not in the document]
GregSutcliffe: That said, keeping the relevant crons in redmine repo probably makes sense, maybe we should rename the directory though...

puppet/modules/redmine/templates/git_repos.sh
```diff
@@ -0,0 +1,51 @@
...
+done
+
+cd /usr/share/redmine
+
+# Create repositories in the Redmine projects for all known git repos
+curl http://prprocessor-theforeman.rhcloud.com/redmine_repos | sudo -u <%= @db_username %> script/rails runner -e production '
+JSON.load(STDIN).each do |project_name,repos|
+  repos.each do |repo,branches|
+    org_name, repo_name = repo.split("/", 2)
+    repo_path = File.join('/usr/share/redmine_data', "git", "#{repo_name}.git") + File::SEPARATOR
```
ekohl: This single quotes break at least highlighting.

GregSutcliffe: Fixed, ty

puppet/modules/redmine/manifests/init.pp
GregSutcliffe: Updated

puppet/modules/redmine/templates/postgresql_backup.sh
GregSutcliffe: Sone, and confirmed `/usr/share/redmine_data/` is now being backed up as usual.

PR #341, opened 2017-10-12
Set the docroot owner for the jenkins vhost

ekohl:
Since it's set to root:root 750 by default this allows access to the .well-known directory which makes letsencrypt work.


mmoll:
merged, bedankt @ekohl!

PR #342, opened 2017-10-12
Use a unique name for the https vhost

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #343, opened 2017-10-12
Install letsencrypt using packages

ekohl:


ekohl:
This is not a valid way since there is no EL6 package so the module has correct autodetection logic. https://github.com/theforeman/puppet-git/pull/32 should be a better approach.

PR #345, opened 2017-10-13
Fixes #21318 - disable environments check in plugin PR tests

iNecas:


iNecas:
An example of such failures is here http://ci.theforeman.org/job/test_plugin_pull_request/4204/. I was not able to reproduce locally, therefore I'm not sure if this fixes the issue, but everything suggests it should (we also use it this way in core PRs)

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
```diff
@@ -46,7 +46,7 @@ fi
 ) > $APP_ROOT/config/database.yml
 
 # Create DB first in development as migrate behaviour can change
-bundle exec rake db:drop db:create db:migrate
+bundle exec rake db:drop db:create db:migrate DISABLE_DATABASE_ENVIRONMENT_CHECK=true
```
ekohl: Give we configure the development database above, shouldn't we just use ```RAILS_ENV=development``` instead?

iNecas: Let's give it a try

ekohl:
I wonder if you should export it before running the command. Later on is another db:migrate. Maybe @mmoll has more insight.

mmoll:
I think it should be there on the second db:migrate also. I also spotted the following diff from test_develop, which should probably be synced(?):
```diff
diff --git a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
index 7666348..a7f667b 100644
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
@@ -32,8 +32,8 @@ while ! bundle install --without=development -j5; do
   fi
 done
 
-# we need to install node modules for integration tests
-if [ -e "$APP_ROOT/package.json" ]; then
+# we need to install node modules for integration tests (which only run on postgresql)
+if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
   npm install npm@'<5.0.0' # first upgrade to newer npm
   $APP_ROOT/node_modules/.bin/npm install
 fi
```

mmoll:
related: GH-325

iNecas:
The other `db:migrate` updated as well. I would like to keep the other changes outside of scope for this PR, as I've seen no issues with running the integration tests in plugins and would like to avoid breaking something that worked :)

mmoll:
let's bring this in and see, if it helps :)

PR #346, opened 2017-10-13
Deploy secure config from Foreman params

GregSutcliffe:
Part1 of deploying the tokens from Foreman. Once this is on disk i'll update the redmine repo to use it.

PR #347, opened 2017-10-13
Enable HTTPS on Redmine

ekohl:
Includes https://github.com/theforeman/foreman-infra/pull/346


puppet/modules/redmine/manifests/init.pp
```diff
@@ -1,13 +1,34 @@
...
+#
+# $db_password::    The password used to connect to redmine
+#
+# $https::          Whether to enable the https vhost
+#
+class redmine (
+  String $secret_token           = 'token',
+  String $email_password         = 'pass',
+  Stdlib::Absolutepath $app_root = '/usr/share/redmine',
+  Stdlib::Absolutepath $data_dir = '/usr/share/redmine_data',
```
GregSutcliffe: Set this to where we really want it, I can set a temporary override in Foreman for now.

puppet/modules/redmine/manifests/init.pp
```diff
@@ -1,13 +1,34 @@
...
+  String $username               = 'adminpz8bn8d',
+  String $db_name                = 'redmine4',
```
GregSutcliffe: We should probably use generic names for user/pass and set the real ones in Foreman?

puppet/modules/redmine/manifests/init.pp
```diff
@@ -86,44 +133,64 @@
...
 
   file { '/etc/cron.hourly/redmine_repos':
-    ensure  => present,
+    ensure  => file,
     owner   => 'root',
     group   => 'root',
     mode    => '0755',
     content => "#!/bin/bash
-sudo -u ${db_username} /usr/local/bin/redmine_repos.sh"
+sudo -u ${username} /usr/local/bin/redmine_repos.sh ${app_root} ${data_dir}",
```
GregSutcliffe: Good call, this will make it easy to return the cron scripts to the redmine git repo if we want...

puppet/modules/redmine/manifests/init.pp
ekohl: Good one. Is it a good idea to submit this, import in foreman, set the overrides and fix it in git? it's not like these values aren't in git already.

GregSutcliffe: I'm fine with that.

PR #350, opened 2017-10-14
Fix letsencrypt cron for VCS installs

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #352, opened 2017-10-16
Add webpack build step to test_develop

ehelms:
Adding this check ensures that the RPM build step that has caused
failing builds multiple times is tested before code goes into the
repository.


mmoll:
untested, but looks good 👍

PR #355, opened 2017-10-17
Fix redmine backup script

ekohl:
This was still dumping the database to the old location. It also adds the .erb extension and introduces the DBNAME variable


mmoll:
merged, bedankt @ekohl!

PR #363, opened 2017-10-21
Add plugin pipeline

ehelms:


ehelms:
Will require the following to work:

 * https://github.com/theforeman/foreman-packaging/pull/1876
 * https://github.com/theforeman/foreman-packaging/pull/1877
 * https://github.com/theforeman/foreman-packaging/pull/1878

See failures at http://ci.theforeman.org/blue/organizations/jenkins/foreman-plugins-release/detail/foreman-plugins-release/6/pipeline/27 for reviewing

ekohl:
If I understand this correct then the mash script mashes all plugin repos we have. Then if one plugin repo breaks, it holds back all plugin repos. I think this matches what we currently have.

This looks good and makes it much clearer what is broken. Yay pipelines.

ehelms:
I updated this to now include the RPM push step. Please review carefully.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanPluginsRelease.groovy
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+    }
+}
+
+void push_rpms(version) {
+
+    dir('deploy') {
+
+        withRVM(["cap yum repo:sync -S overwrite=false -S merge=true -S repo_source=foreman-plugins-${version}/RHEL/7 -S repo_dest=plugins/${version}/el7"])
```
ekohl: Won't this skip f24 which 1.14 and 1.15 still have builds for?

ehelms: Yes - which I did intend to fix. However, from what I could tell, the current job has been not pushing F24 on those for quite some time (no excuse just pointing out holes I found).

ekohl: Looks like https://github.com/theforeman/foreman-infra/commit/11f911af51cc6449a830d7b771b0bbad5a0aaa68#diff-6f833b6369f41e3e65a2e93f31e495b4 dropped F24 indeed.

ekohl: Would this work if you add a ```dist``` parameter to ```push_rpms```?

```groovy
def source = 'RHEL/7'

if (dist == 'f24') {
    source = 'Fedora/24'
}

withRVM(["cap yum repo:sync -S overwrite=false -S merge=true -S repo_source=foreman-plugins-${version}/${source} -S repo_dest=plugins/${version}/${dist}"]) 
```

ehelms:
@ekohl I've updated this to include F24 pushes for the relevant versions.

PR #364, opened 2017-10-21
do letsencrypt renew quietly

mmoll:


ekohl:
Why would you not do this weekly? It should check the expiration dates on the certificate and be a noop if nothing is about to expire.

mmoll:
I'm not filtering any mails, so I'm now getting three mails each Saturday.

ekohl:
How about using ```--quiet``` then? That's what Debian does by default as well. They even have a note that upstream recommends running twice a day.

mmoll:
that sounds perfectly fine, changed 👍

puppet/modules/web/manifests/letsencrypt.pp
```diff
@@ -7,7 +7,7 @@
   }
 
   cron { 'letsencrypt_renew':
-    command => "${::letsencrypt::command} renew --renew-hook '/sbin/service httpd reload'",
+    comman  => "${::letsencrypt::command} renew --quiet --renew-hook '/sbin/service httpd reload'",
```
ekohl: You dropped a letter here.

mmoll: fixed.

ekohl:
Thanks!

PR #367, opened 2017-10-23
Fixes #21318 - disable environments check in plugin PR tests

iNecas:


iNecas:
I guess we need to get to the original approach https://github.com/theforeman/foreman-infra/pull/345#pullrequestreview-69188946, as the db:drop and create by default deals with all environments, and by adding RAILS_ENV=development, we just limited it to the devel

evgeni:
should the same be done in https://github.com/iNecas/foreman-infra/blob/4566dab7df25647288a8bb8170c3dd344f3de9aa/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh#L71?

iNecas:
No, AFAIK, the `db:migrate` always touches just the devel database by default

ekohl:
But I think that line should no longer have the ```RAILS_ENV``` var

iNecas:
Why not? It's just more explicit on what we want to achieve

evgeni:
It fixes the issue I was reporting on IRC, so I'd accept either version you think is the cleanest :)

ekohl:
To me it's odd you do a full DB migration (drop + creation + migration), then install the DB and then do a migration only one env. Then there's a DB migration. Then again, I'm not that familiar with rails testing so it might be normal.

iNecas:
The test db gets loaded not via migration, but via loading a db schema, that gets generated though the migration in dev.

ekohl:
Thanks, let's try this then and see if we still have failures.

PR #374, opened 2017-11-05
Create foreman_datacenter.yaml

Zenya:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_datacenter.yaml
```diff
@@ -0,0 +1,6 @@
+- project:
+    name: foreman_datacenter
+    repo: foreman_datacenter
```
ekohl: I don't see this project on https://github.com/theforeman?utf8=%E2%9C%93&q=foreman_datacenter&type=&language= and I think we can currently only build from theforeman github org.

GregSutcliffe:
Thanks @Zenya 

We only offer testing for projects under our namespace, that's correct. Transferring the project (you'll retain commit rights, of course!) helps the community maintain the plugin in the long term. Take a look at https://theforeman.org/plugins/#3.1Makingyourpluginofficial and see if that's something you'd like to do.

mmoll:
AFAICT, that plugin doesn't even have tests, so I'm closing this anyway :)

PR #377, opened 2017-11-10
Enable cleanup of old user accounts

GregSutcliffe:
We don't actually have a way to remove old keys (although we can stop deploying them by settting ensure to something else, it's a pain to remove them by hand).

### DIFF
--- a/puppet/modules/users/manifests/account.pp
+++ b/puppet/modules/users/manifests/account.pp
@@ -36,6 +36,11 @@ define users::account(
       group   => $name,
       mode    => '0600',
     }
+  } elsif $ensure == 'absent' {
+    # Allow to revoke users access for cleanup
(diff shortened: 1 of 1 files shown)


mmoll:
merged, thanks @GregSutcliffe!

ekohl:
Isn't this redundant? If we ensure the user to be absent, then the user can't login anyway. Doesn't hurt, but I don't see that it's needed.

ekohl:
And since ```managehome``` is true, the homedir should even be deleted.

GregSutcliffe:
Ugh, i missed managehome entirely, was just going for a belt'n'braces cleanup on the keys. Sees to properly remove the homedir, so this can be reverted. Apologies!

ekohl:
https://github.com/theforeman/foreman-infra/pull/383

PR #380, opened 2017-11-12
Add tfm-ror51 scratch build testing on PRs

ehelms:
A bit of a WIP

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR51.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR51.groovy
@@ -1,0 +1,56 @@
+def packages_to_build
+
+pipeline {
+    agent { label 'rpmbuild' }
+
+    parameters {
+        string(name: 'previousCommit', defaultValue: 'HEAD', description: 'Previous commit in master')
+    }
+
+    triggers {
+        pollSCM('* * * * *')
+    }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+        disableConcurrentBuilds()
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+                git url: 'https://github.com/theforeman/tfm-ror51-packaging', branch: 'master'
+
+            }
+        }
+        stage('Find Packages to Build') {
+            steps {
+
+                script {
+                    merge_info = sh(returnStdout: true, script: "git rev-list --parents -n 1 ${previousCommit}").split()
+                    packages_to_build = sh(returnStdout: true, script: "git diff ${merge_info[1]}...${merge_info[0]} --name-only -- 'packages/*.spec' | cut -d'/' -f2 |sort -u").trim()
+                    packages_to_build = packages_to_build.split().join(':')
+                }
+
+            }
+        }
+        stage('Release Build Packages') {
+            steps {
+
+                script {
+                    runPlaybook {
+                        inventory = 'package_manifest.yaml'
+                        playbook = 'release_package.yml'
+                        limit = packages_to_build
+                    }
+                }
+
+            }
+        }
+    }
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51CoprCleanup.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51CoprCleanup.groovy
@@ -1,0 +1,40 @@
+pipeline {
+    agent { label 'rpmbuild' }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+    }
+
+    triggers {
+        cron('0 0 * * *')
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+                git url: 'https://github.com/theforeman/tfm-ror51-packaging', branch: 'master'
+
+            }
+        }
+        stage('Find and Remove Copr Scratch Repositories') {
+            steps {
+
+                script {
+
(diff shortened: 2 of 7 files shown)


ehelms:
@evgeni I borrowed this idea from you if you don't mind reviewing.

evgeni:
While I am not 100% familiar with the new syntax, this looks pretty fine to me.

Two questions:
* Will that properly update GitHub CI status on failures?
* Is there any saner way of saying "just merge the changes GH told us about" instead of doing the raw `GitSCM` dance we need for GitLab?

ehelms:
This will -- it is using Github Pull Request Builder plugin which handles everything you can see this in action -- https://www.google.com/url?hl=en&q=https://github.com/theforeman/tfm-ror51-packaging/pull/1

The only other way I could find to avoid the GitSCM dance was to use a new way of approaching this via Github Organizations and Jenkinsfile's in the source repository itself -- the Travis model.

ehelms:
Going to have to resolve -- https://github.com/theforeman/foreman-infra/pull/382 before the job will actually pass

ehelms:
@ekohl @evgeni Before I merge the first big PR for the Rails SCL, I'd like to ensure this CI job is complete and in the codebase (so I am holding on merging https://github.com/theforeman/tfm-ror51-packaging/pull/1)

ehelms:
@evgeni @ekohl I added another job to this. A take on adding a release pipeline. The idea being that when a PR is merged, this will auto-build to the copr-nightly tags.

One issue I faced, that I am still unclear on is how to ensure that it only checks packages from the commits that were actually merged. That is, I couldn't find good info on how, using Github Plugin to trigger builds (which I am still not sure its doing cleanly) to know the previous point in Git time was X and we are now at commit Y to do the diff between. So right now it just assumes that there is only ever one commit HEAD (but would not handle a rebase and merge with multiple commits).

Also, @evgeni based on your prior experience, can you look at and tell me if I am doing the git checking correctly.

ekohl:
@ehelms I think that @xsuchy suggested to create a temporary copr for a PR test run based on the scl copr and build the packages into that.

ehelms:
He did suggest that as a method. The only advantage I can see with it is potentially avoidiing race condition of updating the same package in two separate PRs.

ehelms:
I suppose repoclosure could also be slightly off.

ekohl:
And you also have the possibility to test a repository just with that change. Otherwise you need to clean up changes that end up not being merged which can fail. Not having to worry about cleaning up a repository but rather delete it altogether makes for an easy workflow.

ehelms:
This and https://github.com/theforeman/tfm-ror51-packaging/pull/2 have been updated to move to a strategy of creating a copr repository for each scratch build. This also includes a release job and a cleanup copr job. The idea is the cleanup Copr job will run daily at midnight and erase all scratch build repositories.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR51.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+                script {
+                    merge_info = sh(returnStdout: true, script: "git rev-list --parents -n 1 ${previousCommit}").split()
+                    packages_to_build = sh(returnStdout: true, script: "git diff ${merge_info[1]}...${merge_info[0]} --name-only 'packages/*.spec' | cut -d'/' -f2 |sort -u").trim()
+                    packages_to_build = packages_to_build.split().join(':')
+                }
+
+            }
+        }
+        stage('Scratch Build Packages') {
```
evgeni: That should be "Release Build Packages", right?

ehelms: Aye.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-ror51-cleanup-copr.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- job:
+    name: tfm-ror51-cleanup-copr
+    project-type: workflow
+    logrotate:
```
evgeni: you configure `buildDiscarder` in he groovy file, so I'd think this is not needed here anymore?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-ror51-pr-test.yaml
```diff
@@ -0,0 +1,18 @@
+---
+- job:
+    name: tfm-ror51-pr-test
+    project-type: workflow
+    logrotate:
```
evgeni: you configure `buildDiscarder` in he groovy file, so I'd think this is not needed here anymore?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror51-release.yaml
```diff
@@ -0,0 +1,17 @@
+---
+- job:
+    name: tfm-ror51-release
+    project-type: workflow
+    logrotate:
```
evgeni: you configure `buildDiscarder` in he groovy file, so I'd think this is not needed here anymore?

ehelms:
I'd like to tackle the re-factor after it's merged and I have a few test runs. I am still not sure, given not having enough uses yet, that the release job will cover all edge cases.

ehelms:
@ekohl this awaits your review approval as well

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51CoprCleanup.groovy
```diff
@@ -0,0 +1,40 @@
...
+    triggers {
+        cron('0 0 * * *')
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+                git url: 'https://github.com/theforeman/tfm-ror51-packaging', branch: 'master'
```
ekohl: In many cases I wondered if we can reuse the github -> url property here to reduce duplication.

PR #381, opened 2017-11-13
Install google-chrome in slaves

dLobatog:
In order to use chrome headless with Selenium, and get rid of phantomjs.
For the moment, both can be installed to test https://github.com/theforeman/foreman/pull/4987
When that PR is merged and we no longer need phantomjs and firefox, 
we can remove them on another PR.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -215,7 +215,7 @@ class slave (
       ensure => directory,
       owner  => 'jenkins',
       group  => 'jenkins',
-      mode    => '0775',
+      mode   => '0775',
     }
 
     # Cleanup temporary dir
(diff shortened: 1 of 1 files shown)


puppet/modules/slave/manifests/init.pp
```diff
@@ -193,6 +195,15 @@
...
   # packages only really available on EL6+
   if $::osfamily == 'RedHat' {
+    yumrepo { 'google-chrome':
+      name => 'google-chrome - \$basearch',
+      ensure => present,
+      baseurl => 'http://dl.google.com/linux/chrome/rpm/stable/\$basearch',
+      enabled => '1',
+      gpgcheck => '1',
+      gpgkey => 'https://dl-ssl.google.com/linux/linux_signing_key.pub',
+      before => Package['google-chrome-stable'],
```
ekohl: alignment is off here

dLobatog: @ekohl I've updated the PR to add a space after the `yumrepo` resource, don't see any alignment issues here, indentation is the same as the other keys, could you clarify?

ekohl: Looks like we don't have lint enabled but we align the arrows. See the file declaration above it.

puppet/modules/slave/manifests/init.pp
```diff
@@ -168,6 +168,8 @@
       ensure => present;
     'ansible':
       ensure => latest;
+    'google-chrome-stable':
```
ekohl: I just realized that this will break on Debian since the package is not available there. You could install chromium but that might result in different paths.

mmoll:
I'd rather add the apt repo for Debian, but AFAICT, this will never be used on Debian slaves currently, so this can also get ommited for Debian.

ekohl:
@mmoll so move the package installation to the osfamily check just below the repo?

mmoll:
yes.

dLobatog:
@ekohl What do you need the binary location in Debian for?

ekohl:
@dLobatog if we don't run tests on Debian machines then it doesn't matter. It could have mattered if we did ran them and installed chromium because the chromedriver package might not look for the default chromium location.

ohadlevy:
@ekohl whats the status of the pr? thanks :)

ekohl:
@ohadlevy it was suggested to only install chrome on the RH-osfamily and skip Debian so we're waiting on that.

mmoll:
merged, gracias @dLobatog!

ohadlevy:
does it mean that in about an hour i can retry the tests on the foreman pr?

ekohl:
@ohadlevy yes

PR #382, opened 2017-11-13
Add copr config template

ehelms:
@GregSutcliffe Check me if I am wrong in how we handle configuring files with secrets here. I think these then get defined in our Foreman instance?

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -4,7 +4,10 @@ class slave (
   $jenkins_build_token = undef,
   $koji_certificate    = undef,
   $rackspace_username  = undef,
-  $rackspace_api_key   = undef
+  $rackspace_api_key   = undef,
+  $copr_login          = undef,
+  $copr_username       = undef,
+  $copr_token          = undef,
 ) {
   file { '/var/lib/workspace':
     ensure => directory,
@@ -392,6 +395,13 @@ class slave (
     owner  => 'jenkins',
     group  => 'jenkins',
     source => 'puppet:///modules/slave/katello-ca.cert',
+  }
+  file { '/home/jenkins/.config/copr':
+    ensure   => file,
+    mode     => '0640',
+    owner    => 'jenkins',
+    group    => 'jenkins',
+    content  => template('slave/copr.erb'),
   }
   if $koji_certificate {
     file { '/home/jenkins/.katello.cert':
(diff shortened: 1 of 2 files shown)


puppet/modules/slave/manifests/init.pp
```diff
@@ -393,6 +393,13 @@
     group  => 'jenkins',
     source => 'puppet:///modules/slave/katello-ca.cert',
   }
+  file { '/home/jenkins/.config/copr':
+    ensure   => file,
+    mode     => '0644',
```
ekohl: I think 640 might be better since it has tokens

puppet/modules/slave/templates/copr.erb
```diff
@@ -0,0 +1,7 @@
+[copr-cli]
+login = <%= @copr_login %>
```
ekohl: Where do these variables come from? Global variables?

puppet/modules/slave/templates/copr.erb
```diff
@@ -0,0 +1,7 @@
...
+username = <%= copr_username %>
```
ekohl: Do these also need ```@```?

puppet/modules/slave/templates/copr.erb
ehelms: I *think* so, but hoping @GregSutcliffe can confirm

GregSutcliffe: Foreman can hold secure variables and pas them in as smart class parameters, yes. Just set sane defaults here.

puppet/modules/slave/manifests/init.pp
GregSutcliffe: +1

GregSutcliffe:
Does this have to live on the slaves? I trust all our sponsors but technically that's a large group of people who could get our certificate. Is there another way to pass this in?

ehelms:
The only other option I know about is to create a temporary config file in the CI job, and point copr-cli at that, then clean it up with the workspace.

GregSutcliffe:
I guess that's not much better, given it'd be predictable. Does Copr handling signing or is that step done elsewhere?

ehelms:
I'll give the above a try before we go this route.

ehelms:
RPM signing? Copr handles that. The temporary config file would only be persistent for a short period, however, this is proving difficult to do with our use of tito on top of things. Unfortunately, Koji tends to work this same way, so we have this issue regardless.

GregSutcliffe:
Fair enough. I was under the impression that the releases were signed separately.

ehelms:
@ekohl Updated

ekohl:
I wonder if we could solve the credentials issue is with splitting the builds: jenkins can only builds to the scratch repo(s) and handle production releases through github webhooks to copr.

PR #383, opened 2017-11-16
Revert "Enable cleanup of old user accounts"

ekohl:
This reverts commit 7c4e69995f9207916d25109e79a3fd82d88cabcd.

The previous code had managehome => true so the homedir is already
cleaned up.


GregSutcliffe:
Thanks @ekohl sorry again for the mess.

PR #385, opened 2017-11-20
only list jessie, stretch, trusty and xenial

evgeni:
all other releases have way too outdated foreman releases

also replace `rc` with `<version>` because we don't really have an rc pocket

### DIFF
--- a/puppet/modules/freight/files/deb-HEADER.html
+++ b/puppet/modules/freight/files/deb-HEADER.html
@@ -9,24 +9,21 @@
 
 <p>Currently we have</p>
 
-<pre>deb http://deb.theforeman.org/ wheezy stable
-deb http://deb.theforeman.org/ jessie stable
+<pre>deb http://deb.theforeman.org/ jessie stable
 deb http://deb.theforeman.org/ stretch stable
 deb http://deb.theforeman.org/ trusty stable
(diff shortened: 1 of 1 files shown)


puppet/modules/freight/files/deb-HEADER.html
```diff
@@ -9,24 +9,21 @@
...
-deb http://deb.theforeman.org/ precise stable
-deb http://deb.theforeman.org/ wheezy rc
-deb http://deb.theforeman.org/ jessie rc
-deb http://deb.theforeman.org/ stretch rc
-deb http://deb.theforeman.org/ trusty rc
-deb http://deb.theforeman.org/ precise rc
-deb http://deb.theforeman.org/ wheezy nightly
+deb http://deb.theforeman.org/ xenial stable
+deb http://deb.theforeman.org/ jessie &lt;version&gt;
+deb http://deb.theforeman.org/ stretch &lt;version&gt;
```
mmoll: is that syntax needed in <pre>?

evgeni: yes: https://jsfiddle.net/ckx4ec22/

mmoll:
merged, спасибо Женя!

PR #387, opened 2017-11-20
Move CI jobs folder to top-level

ehelms:
Per -dev thread, attempting to move the jobs to a more obvious location for management of them. I assume this will just work with the symlink.

### DIFF
--- a/ci
+++ b/ci
@@ -1,0 +1,1 @@
+puppet/modules/jenkins_job_builder/files/theforeman.org/


ekohl:
I wonder if the deployment job installs a symlink or the actual jobs.

ehelms:
Got me -- the whole use of puppet here is a black box to me. I make PRs, wait 30 minutes and things happen.

GregSutcliffe:
@ehelms Soooo... it's a little convoluted, but we didn't want the slaves having unrestricted ssh-key access to the master. Instead, the ssh connection triggers https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/deploy/templates/script.erb on the master side.

Somewhat convoluted, since it could just change for changes via cron. I'd be open to changing it, but as it stands, you'll want to update that script to include copying the the CI directory from the clone.

GregSutcliffe:
@ehelms Soooo... it's a little convoluted, but we didn't want the slaves having unrestricted ssh-key access to the master. Instead, the ssh connection triggers https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/deploy/templates/script.erb on the master side.

Somewhat convoluted, since it could just change for changes via cron. I'd be open to changing it, but as it stands, you'll want to update that script to include copying the the CI directory from the clone.

ehelms:
So at the top level, that script would not provide access to the top level `ci` directory? So then the only benefit over a separate repository is just keeping it in the same git repository?

GregSutcliffe:
Indeed, moving it to another repo is just as feasible - except that we'd want to update the Jenkins job to check for changes in the new repo as well as foreman-infra (or change the script to a cron or something...)

ehelms:
@ekohl @GregSutcliffe How about this simpler approach? This gets me a faster way to get to the CI data without any additional overhead.

PR #392, opened 2017-11-24
Install chromedriver in slaves

dLobatog:
Precedes https://github.com/theforeman/foreman/pull/4987, tests are failing with 

```
Selenium::WebDriver::Error::WebDriverError:  Unable to find chromedriver. Please download the server from http://chromedriver.storage.googleapis.com/index.html and place it somewhere on your PATH. More info at https://github.com/SeleniumHQ/selenium/wiki/ChromeDriver....
```

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -251,6 +251,10 @@ class slave (
     package { 'google-chrome-stable':
       ensure => latest,
     }
+
+    class { 'chromedriver':
+      ensure => latest,
+    }
   }
 
   # needed by katello gem dependency qpid-messaging
--- a/.gitmodules
+++ b/.gitmodules
@@ -119,3 +119,6 @@
 [submodule "puppet/modules/datacat"]
 	path = puppet/modules/datacat
 	url = https://github.com/richardc/puppet-datacat.git
+[submodule "puppet/modules/chromedriver"]
+	path = puppet/modules/chromedriver
+	url = https://github.com/rfletcher/puppet-chromedriver


puppet/modules/slave/manifests/init.pp
```diff
@@ -251,6 +251,12 @@
     package { 'google-chrome-stable':
       ensure => latest,
     }
+
+    include ::chromedriver
```
ekohl: chromedriver is also in EPEL. Currently at version 61.0.3163.100. Would that be new enough?

puppet/modules/slave/manifests/init.pp
```diff
@@ -251,6 +251,12 @@
...
+
+    class { 'chromedriver':
```
ekohl: Also, you can't both include it and define it as a class. That's redundant.

dLobatog: Sorry, leftover from testing

puppet/modules/slave/manifests/init.pp
```diff
@@ -251,6 +251,10 @@
     package { 'google-chrome-stable':
       ensure => latest,
     }
+
+    class { 'chromedriver':
+      ensure => latest,
```
ekohl: Looking at the source code this will install version 2.24 but the notes on https://chromedriver.storage.googleapis.com/2.33/notes.txt say we need 2.33 to work with the current chrome version.

dLobatog:
@ekohl I've opened https://github.com/theforeman/foreman-infra/pull/396, so you can compare both solutions and decide which you prefer

ohadlevy:
@dLobatog whats the status of the pr? thanks

dLobatog:
Abandoned - https://github.com/theforeman/foreman-infra/pull/396 got merged

PR #393, opened 2017-11-24
Fix gathering version info from Katello source

iNecas:


ekohl:
I just had the same file open in my editor and was going to make the same change. Thanks!

PR #397, opened 2017-12-04
Drop db:drop from Katello develop PR test

ehelms:
Rails 5 does not handle dropping a non-existent database gracefully


mmoll:
let's go for it, thanks @ehelms!

PR #399, opened 2017-12-06
set user in jenkins .gitconfig

mmoll:


mmoll:
I saw this when watching puppet diffs... just to not loose it...

ekohl:
Thanks!

PR #402, opened 2017-12-07
Update jenkins jobs for 1.17

xprazak2:
Do not merge as this is work in progress.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_1_17_stable.yml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_1_17_stable.yml
@@ -1,0 +1,45 @@
+- job:
+    name: test_1_17_stable
+    project-type: matrix
+    logrotate:
+      daysToKeep: -1
+      numToKeep: 32
+    quiet-period: 2700
+    properties:
+      - github_foreman
+    scm:
+      - foreman:
+          branch: '1.17-stable'
+    triggers:
+      - scm_fifteen_minutes
+      - github
+      - schedule_failed_builds
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+            - 2.3
+            - 2.4
+      - axis:
+          type: user-defined
+          name: database
+          values:
+            - postgresql
+            - mysql
+            - sqlite3
+      - axis:
+          type: label-expression
+          name: slave
+          values:
+            - fast
+    builders:
+      - test_develop
+    publishers:
+      - gemset_cleanup
+      - ircbot_freenode
+      - archive:
+          artifacts: 'pkg/*,Gemfile.lock'
+          only-if-success: false
+      - junit:
+          results: 'jenkins/reports/unit/*.xml'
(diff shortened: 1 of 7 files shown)


xprazak2:
This is no longer wip and can be merged as we now have 1.17 branches

mmoll:
I think there's a test_proxy_1_17_stable job missing.

release_plugins_push is probably in `puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanPluginsRelease.groovy` now, but probably @ehelms knows better.

xprazak2:
I updated `foremanPluginsRelease.groovy` and added `test_proxy_1_17_stable` job

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_1_17_stable.yml
```diff
@@ -0,0 +1,46 @@
...
+    triggers:
+      - scm_fifteen_minutes
+      - github
+      - schedule_failed_builds
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+            - 2.2
```
ekohl: I think we can assume Ruby 2.3+ since Debian Stretch and Ubuntu Xenial are both 2.3 and our RH uses the 2.4 SCL.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_1_17_stable.yaml
```diff
@@ -0,0 +1,46 @@
...
+          - 2.0
+          - 2.1
+          - 2.2
+          - 2.3
+          - 2.4
+      - axis:
+          type: user-defined
+          name: puppet
+          values:
+          - 3.4.0
```
ekohl: We can drop Puppet 3.4.0, 4.2.0 and 4.4.0. Add to that 4.10.9 and 5.3.3 so we test all the major versions but only their latest ones.

xprazak2:
I made changes as suggested.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_1_17_stable.yaml
```diff
@@ -0,0 +1,45 @@
...
+          - 2.4
+      - axis:
+          type: user-defined
+          name: puppet
+          values:
+          - 3.8.0
+          - 4.10.9
+          - 5.3.3
+    execution-strategy:
+      combination-filter: '!( (ruby ==~ /2\.[^0]*/ && puppet ==~ /3\.[0-4].*/) || (ruby ==~ /2\.[2-9].*/ && puppet ==~ /3.*/) || (ruby ==~ /2\.[3-9].*/ && puppet ==~ /4\.[0-3].*/) )'
```
ekohl: This needs an update now. I'm not sure what we exactly want, but I think:
* 3.8.0 + 2.0-2.1
* 4.10.9 + 2.1-2.4
5.3.3 + 2.4

xprazak2:
I updated the combination filter for Puppet and Ruby versions.

mmoll:
I think the combination filters are not correct, as 1.17 will only have xenial and stretch.

xprazak2:
I updated the combination filters for debs so that 1.17 has only xenial and stretch

mmoll:
merged, diky @xprazak2!

PR #407, opened 2017-12-15
Rewrite systests to forklift

ekohl:
Here just so it's not just on my laptop. Feel free to give some early feedback, but this is not tested at all.

* [x] Convert umask to ansible role + args
* [x] Choose different databases via ansible
* [x] Convert run_hammer_tests to ansible role
* [ ] Run tests in various ways
* [x] Convert Jenkins jobs calling this test (align jenkins parameters to ansible settings)

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
@@ -16,20 +16,19 @@
           block: true
       - trigger-builds:
         - project: systest_foreman
-          predefined-parameters: "os=${os}\nrepo=${major_version}\nstaging=true\nexpected_version=${version}"
+          predefined-parameters: "os=${os}\nrepo=${major_version}\nrepo_flavor=staging\nexpected_version=${version}"
           block: true
     axes:
       - axis:
           type: user-defined
           name: os
           values:
-          - el7
-          - jessie
-          - stretch
-          - trusty
-          - xenial
+          - centos7
+          - debian8
+          - debian9
+          - ubuntu1604
     execution-strategy:
-      combination-filter: 'os == "el7" || (os == "jessie" && (major_version == "1.15" || major_version == "1.16")) || (os == "stretch" && major_version != "1.15") || (os == "trusty" && major_version == "1.15") || os == "xenial"'
+      combination-filter: 'os != "debian8" || major_version == "1.16"'
     publishers:
       - trigger-parameterized-builds:
           - project: 'release_push_deb, release_push_rpm'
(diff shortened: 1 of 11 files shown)


ehelms:
What actually uses systest still? Is it the release jobs? The nightly is already using Forklift right?

ekohl:
RPM nightly is, not Debian/Ubuntu. This is also the only umask test we currently have. Another thing we need to cover is the [fb-hammer-tests.bats](https://github.com/theforeman/foreman-bats/blob/master/fb-hammer-tests.bats).

I'd be happy to fully move to groovy pipelines also for Debian but refactoring the current job and making sure we have feature parity was the easiest way for me.

ehelms:
Any hold up to merging this @ekohl ?

ekohl:
Yes, it's 100% untested :) Added tasks so you can see the checklist.

ekohl:
https://github.com/theforeman/forklift/pull/590 adds the hammer tests. Haven't had much time to properly test this in the various configurations. Also doesn't help nightly is currently broken.

ekohl:
Current issue is that retrieving the logs fails with a permission denied on my libvirt setup because it logs in as vagrant, not root. Quite likely this is not an issue on rackspace. I'll move on to testing with rackspace soon.

Otherwise I think this could use another pair of eyes.

ehelms:
The code here seems sane enough

ekohl:
I think this is now complete in terms of code. Will need testing.

ehelms:
You can drop anything katello out of the surviving code. The project has moved entirely off using this job.

ekohl:
Would it make sense to expand ```foremanRelease.groovy``` to run multiple systests in parallel (vanilla, umask 077, mysql)?

Then we only have puppet_pc1 and puppet_nightly left as cronjobs. Personally I don't care much for them and I don't think we really react to them currently.

Lastly there's ```release_test``` in our release flow which this change currently breaks. Going to look at that tomorrow.

ehelms:
@ekohl Generally speaking +1 to having each configuration run as a single test with parallel branches or in sequence.

Some of those tests are ran once a week currently, running nightly would increase the load on Rackspace usage which might be a limiting factor. My use of ci.centos.org has been mostly successful with containers except I need them to address [1] so we can write in pipelines. More and more I am sensing a need to solve juggling the use of their CI system for provisioning and our CI system for the secretive actions (e.g. packaging mash and RPM push).

[1] https://bugs.centos.org/view.php?id=14496

komidore64:
:+1:

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
```diff
@@ -16,20 +16,19 @@
...
-          - stretch
-          - trusty
-          - xenial
+          - centos7
+          - debian8
+          - debian9
+          - ubuntu1604
     execution-strategy:
-      combination-filter: 'os == "el7" || (os == "jessie" && (major_version == "1.15" || major_version == "1.16")) || (os == "stretch" && major_version != "1.15") || (os == "trusty" && major_version == "1.15") || os == "xenial"'
+      combination-filter: 'os != "debian8" || major_version == "1.16"'
```
ekohl: Note this drops support for releasing 1.15 versions.

ekohl:
Dropped the katello compatibility code and fixed the release job to handle centos7 instead of el7.

PR #416, opened 2017-12-29
remove old stuff

mmoll:


ekohl:
With this merged, the jobs still aren't purged I think. That's a manual action, right?

PR #418, opened 2017-12-30
fix letsencrypt http challenge

mmoll:


mmoll:
@ekohl could you have a look here? At the moment the cert renews fails. The code is untested.

puppet/modules/web/manifests/init.pp
```diff
@@ -68,6 +68,7 @@
     docroot_group   => 'website',
     docroot_mode    => '0755',
     custom_fragment => template('web/web.conf.erb'),
+    default_vhost   => true,
```
ekohl: Is this used and does it have an effect? I thought it defined a mapping above to serve it via plain HTTP.

mmoll: hrm, yes and no... this was a brainfart, but the original problem is still unsolved then: LE is putting the response into /var/www/html, which is not served

ekohl:
It looks like the entire vhost web02 no longer exists (probably because I removed the default vhost) and I'm not sure that cert is actually used. Since it only reloads apache I'm leaning to no and we could remove it.

mmoll:
ah, I thought the theforeman.org cert was in the web02 config, while it isn't :)

mmoll:
all fine 👍

ekohl:
Should we remove the web02 cert then so we don't get the renewal warnings anymore?

mmoll:
I deleted the respective certbot config file and puppet didn't regenerate it, so I think it should be OK.

PR #423, opened 2018-01-08
generate theforeman.org with Ruby 2.4

mmoll:


mmoll:
This is in line with Github Pages itself: https://pages.github.com/versions/

PR #424, opened 2018-01-08
drop Ruby 2.2 tests for develop

mmoll:


mmoll:
Debian 9 and Ubuntu 16.04 use 2.3, so no environment supported with current develop (or even 1.17) does use 2.2.

PR #425, opened 2018-01-09
further tune postgresql on slaves

mmoll:


mmoll:
I hope this will make test runs a little bit faster.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -26,4 +31,27 @@
   postgresql::server::config_entry { 'synchronous_commit':
```
ekohl: Should we rewrite this to a hash:

```puppet
$settings = {}
$settings.each |$setting, $value| {
  postgresql::server::config_entry { $setting:
    value => $value,
  }
}
```

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -3,6 +3,11 @@
     cwd => '/',
   }
 
+  class { 'postgresql::globals':
+    manage_package_repo => true,
+    version             => '9.6',
```
ekohl: Why 9.6 in particular? Shouldn't we try with the OS versions since that's what we deploy on?

mmoll: That's 9.2 with CentOS 7, while later versions do have some performance improvements.

puppet/modules/slave/manifests/postgresql.pp
mmoll: good point 👍

mmoll:
A different location wouldn't be problem as everything is provisioned newly by Puppet. The worst case could be that the old version DB server is used, but then a failed start of the new server would show up in the Puppet logs every run, I presume.

ekohl:
I wonder about server sockets or duplicate running postgres servers.

mmoll:
As we don't have remote_execution or something similar to control things when updating, I removed the update to 9.6 and included only parameters that are compatible with 9.2 now.

PR #427, opened 2018-01-11
tune postgres only on RH slaves

mmoll:


mmoll:
As Debian/Ubuntu do bring newer versions, not all parameters work there.

mmoll:
Yes, in theory we could stop installing the DBs on non-RHs these days.

ekohl:
Let's merge this now then and later split it into build slave and test slave classes.

PR #430, opened 2018-01-12
Add --trace to rake tasks

iNecas:
I'm getting some unexpected aborts on rake test command in remote
execution. I don't see reason hiding the trace when the rake calls in
our CI fail on exception. I believe this can help not just me but all
the future generations trying to figure out WTH it going on.


ekohl:
Thanks!

PR #436, opened 2018-01-30
try hard to drop the db in the katello tests

evgeni:
3cfb9889a769d6cfb872bc4b6d7aaaf4406d8951 removed db:drop alltogether,
which means we can end up running on a DB that had migrations from
another branch applied, which is no good.

try to drop the DB, but ignore all errors, thanks Rails 5


mmoll:
sounds legit!

mmoll:
merged, спасибо Женя!

PR #437, opened 2018-01-30
Bumped discovery to v11

lzap:
Preparing 11.0 release, thanks.


lzap:
Our standard upstream procedure for core is N-1 for security fixes. If we have one in the short window before 1.17 is out, I will test locally.

mmoll:
As long as the 9.1 discovery job doesn't get deleted manually from Jenkins, it'll run normally.

mmoll:
merged, díky @lzap!

PR #438, opened 2018-01-31
Drop & migrate on separate rake commands

dLobatog:
The way this works currently, db:drop db:create db:migrate runs prior to
running the tests. This works, because when these 3 tasks are called by
rake, there is no environment, so the migrations cannot assume any
schema.

However, after the tests run, there's already an existing schema. When
we call db:drop, db:create db:migrate all 3 on the same rake command, by
the time we get to 'db:migrate' - the application has already loaded the
schema we had prior to db:drop, instead of an empty one.

This means, a call like "Host.unscoped.all" is translated to
"SELECT \"hosts\".* FROM \"hosts\" WHERE \"hosts\".\"type\" IN
('Host::Managed')""
when it should be:
"SELECT \"hosts\".* FROM \"hosts\"

This affects an early migration
20090916053824_change_host_build_default_to_false.rb, at that date,
there was no "type" column. This patch prevents the problem from
happening in other migrations as well.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_plugin.sh
@@ -49,7 +49,12 @@ fi
 ) > $APP_ROOT/config/database.yml
 
 # Create DB first in development as migrate behaviour can change
-bundle exec rake db:drop db:create db:migrate DISABLE_DATABASE_ENVIRONMENT_CHECK=true --trace
+bundle exec rake db:drop DISABLE_DATABASE_ENVIRONMENT_CHECK=true --trace
+# We drop and create in 2 separate commands (speed penalty, 2 Rails initializations)
+# as it's necessary to make sure at db:migrate models are loaded from a clean slate.
+# e.g: model Host isn't loaded with a 'type' attribute before the column is added
+# in the migration.
+bundle exec rake db:create db:migrate DISABLE_DATABASE_ENVIRONMENT_CHECK=true --trace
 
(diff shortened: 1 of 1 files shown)


dLobatog:
[This migration is the problem](https://github.com/theforeman/foreman/blob/develop/db/migrate/20090916053824_change_host_build_default_to_false.rb#L5)

If you run `bundle exec rake db:drop db:create db:migrate` and prior to the `db:drop` the database was "fully migrated", you will find that the migration will NOT pass. 

`"SELECT "hosts".* FROM "hosts"` is called as `"SELECT "hosts".* FROM "hosts" WHERE "hosts"."type" IN
('Host::Managed')""` because the schema is loaded *before* running db:drop. With 2 different commands, I'm able to make it pass. 

I have some [plugin tests running here](http://ci.theforeman.org/job/test_plugin_matrix/4630/) - I temporarily modified the Jenkins job to include this change, just to prove it works. Without the patch, [it would hit this error](http://ci.theforeman.org/job/test_plugin_matrix/4629/database=postgresql,ruby=2.4,slave=fast/console)

dLobatog:
@ekohl Where have you seen this? I've only seen it in plugins, however `test_develop` also does the same, run 'bundle exec rake db:drop db:create db:migrate --trace` at the end of the job before running `db:seed`, and I haven't seen the error there. My guess is that initializing plugins during the first `db:drop` is forcing Rails to load the schema in memory, but then again I have not truly proven it.

ekohl:
We have observed in the psat that ```rake db:drop db:create``` doesn't work because it caches database table information. That's why 2 runs does work.

PR #439, opened 2018-01-31
Create foreman_datacenter.yaml

Zenya:
Hi
I added some tests for a start
plugin's organization is still 'cloudevelops' - https://github.com/cloudevelops/foreman_datacenter
we already created deb package in foreman-packaging
is transfering organization to 'theforeman' necessary?

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_datacenter.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_datacenter.yaml
@@ -1,0 +1,6 @@
+- project:
+    name: foreman_datacenter
+    defaults: plugin
+    branch: develop
+    jobs:
+      - 'test_plugin_{name}_{branch}'


ekohl:
> is transfering organization to 'theforeman' necessary?

AFAIK our testing infrastructure only works in our own github namespace.

Zenya:
now foreman_datacenter is in theforeman namespace

mmoll:
are there **any** tests in that plugin?

Zenya:
I added some controllers tests for start
There will be much more tests later

mmoll:
> I added some controllers tests for start

Thanks, I'll check this out locally and will merge if it works in the standard plugin test setup. 👍

mmoll:
@Zenya Please make sure this works against current Foreman **develop**. 1.17+ is on Rails 5.1, which needs updates to plugins. As a first step, the DB migrations need their versions set. After that with PostgreSQL I'm getting `Index name 'index_datacenter_comments_on_commentable_id_and_commentable_type' on table 'datacenter_comments' is too long; the limit is 63 characters`. We do test against PostgrSQL, MySQL and SQLite3.

Zenya:
@mmoll I added `:name` option in migration to fix this error. New changes in develop branch
https://github.com/theforeman/foreman_datacenter/tree/develop

ekohl:
@Zenya recently I added quite a few fixes but master was still failing. I don't see how these tests ever passed. Could you have a look?

PR #441, opened 2018-02-04
Ruby 2.5

mmoll:


mmoll:
Debian/unstable switched its default Ruby to 2.5.0, so Ubuntu 18.04 will use 2.5.

PR #446, opened 2018-02-06
Fix release cron and pipeline job

ehelms:
The cron trigger from the pipeline itself seemed to partially or never truly work. This reverts back to normal cron trigger on the job definition itself.

This also updates a bug in the plugins release pipeline where it never actually called `push_rpms` (oops). This also make a strategic change to do repoclosure + push per release stream. This prevents broken release streams from blocking others. i.e. in progress 1.17 doesn't block nightly.


mmoll:
merged, thanks @ehelms!

PR #449, opened 2018-02-07
Add instructions for local jjb setup

xprazak2:


ehelms:
This might need a warning with it that the docker based Jenkins setup may not work in all cases as it may not have the same configuration (e.g. plugins, environment variables) that our current Jenkins has.

xprazak2:
I added a warning that local setup may not work for some jobs

xprazak2:
I changed it to use docker volume instead of local files, removed the `sudo` for docker command and removed the ambiguity regarding puppet module.

puppet/modules/jenkins_job_builder/README.md
```diff
@@ -0,0 +1,34 @@
...
+`files/theforeman.org` contains job definitions for Jenkins
+
+## development setup
+
+Warning: Your local Jenkins may not work for all the jobs you try to import, because some jobs rely on plugins and global environment variables. Unfortunately, we do not have a list of plugins and variables in a source control yet.
+
+* Set up a local Jenkins instance
+  1. In Docker
+      * you can use official [Docker images](https://hub.docker.com/r/jenkins/jenkins/)
+      * `docker run --name jenkins-lts -p 8080:8080 -p 50000:50000 -v jenkins_home:/var/jenkins_home  jenkins/jenkins:lts`
```
ekohl: The double space triggers my OCD :)

puppet/modules/jenkins_job_builder/README.md
```diff
@@ -0,0 +1,34 @@
+# jenkins_job_builder
+
+`files/theforeman.org` contains job definitions for Jenkins
+
+## development setup
+
+Warning: Your local Jenkins may not work for all the jobs you try to import, because some jobs rely on plugins and global environment variables. Unfortunately, we do not have a list of plugins and variables in a source control yet.
```
ekohl: The rebuild plugin was also removed from the Jenkins index because it didn't have a release for 8 years.

puppet/modules/jenkins_job_builder/README.md
xprazak2: :)

puppet/modules/jenkins_job_builder/README.md
xprazak2: Updated.

xprazak2:
Updated.

ehelms:
Note that https://github.com/theforeman/foreman-infra/pull/492/files will require a slight change in these instructions depending on the order

PR #453, opened 2018-02-12
Update the kafo test matrix

ekohl:
While it may work on Ruby 1.x and old Puppet version, we no longer really care about it. This allows us to simplify the test matrix.


mmoll:
merged, bedankt @ekohl!

PR #454, opened 2018-02-13
Convert test_kafo_master_pull_request to JJB

ekohl:
This is untested and would need a proper review if this covers everything. I am changing the matrix to only test on modern versions.


ehelms:
I *think* this would need to be removed from the pull request scanner as well -- https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/slave/templates/test_pull_requests_kafo.json.erb

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - 2.0.0
+          - 2.1
+          - 2.2
+          - 2.3
+          - 2.4
+      - axis:
+          type: user-defined
+          name: puppet
+          values:
+          - 4.10.0
```
ehelms: Should this be latest of 4.10.Z line?

ehelms:
Small comment otherwise looks good

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
ekohl: I think that this should be true since we use ```~> VERSION```. Maybe it could also be ```4.0``` or ```4.10``` since that should also match it. See https://github.com/theforeman/kafo/blob/93aaa70eff832fc6ccfbcbbf672f5cd775007420/Gemfile#L22

ekohl:
https://projects.theforeman.org/projects/foreman/wiki/Jenkins#Version-3-GHPRB mentions we also need to do some changes when deploying this, is that correct?

ehelms:
Ahh yes, will need to add the webhook to kafo for this.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
```diff
@@ -0,0 +1,50 @@
...
+    logrotate:
+      daysToKeep: 45
+      numToKeep: -1
+    properties:
+      - github:
+          url: https://github.com/theforeman/kafo
+    wrappers:
+      - abort_build:
+          timeout_min: 60
+    parameters:
```
ekohl: I discovered these parameters can be removed when switching to ghprb

ehelms:
@ekohl approved.... but I'll let you deploy it when you want

ekohl:
Updated to also remove the old config.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - axis:
+          type: user-defined
+          name: puppet
+          values:
+          - 4.10
+          - 5.0
+    execution-strategy:
+      combination-filter: '!(ruby ==~ /^2\.[0-3]\..*/ && puppet ==~ /^5\..*/) )'
+      touchstone:
+        expr: 'ruby=="2.1" && puppet=="4.10.0"'
```
ekohl: Oh, this should be 4.10 now instead of 4.10.0

ekohl: Updated. Also noticed the combination filter would probably not work.

ekohl:
Please deploy when you have time. I'll test it out when it's set up and iterate further.

PR #458, opened 2018-02-14
Convert packaging_test_pull_request & handle new directory layout

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_test_pull_request.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - string:
+          name: branch
+          default:
+          description: "Base branch to merge PR into"
+      - pr_parameters
+    scm:
+      - foreman-rpm-packaging-pr
+    triggers:
+      - github_pr:
+          context: 'foreman-packaging'
```
ehelms: Maybe context here would be better as a scratch-build or test-build? or package? something to reflect that its doing a test of scratch build

ekohl: For Debian it ends up on https://stagingdeb.theforeman.org/ and doesn't know the term scratch. Just packaging?

ehelms: Sure - packaging works for now. In the future I envision splitting deb from rpm to make this clearer on a given PR.

ehelms: ACK pending this little update

ekohl: updated

PR #460, opened 2018-02-14
Set status url to pipeline link for pipelines

ehelms:


ehelms:
@jlsherrill since you asked

PR #461, opened 2018-02-14
Remove old Katello PR hook

ehelms:


ekohl:
The config should be purged rather than just the template removed:
```
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Could not find template 'slave/test_pull_requests_katello.json.erb' at /etc/puppetlabs/code/environments/production/modules/slave/manifests/pr_test_config.pp:14:20 at /etc/puppetlabs/code/environments/production/modules/slave/manifests/init.pp:51 on node arm32slave01.scaleway.theforeman.org
```

https://github.com/theforeman/foreman-infra/pull/484 should fix it

PR #464, opened 2018-02-19
Fix nightly RPM builds

ehelms:


pcreech:
ACK

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop.yaml
```diff
@@ -45,7 +45,7 @@
       - junit:
           results: 'jenkins/reports/unit/*.xml'
       - release_nightly_build_foreman:
-          project: foreman
+          project: packages/foreman/foreman
```
ekohl: This broke Debian packaging.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_develop.yaml
```diff
@@ -45,4 +45,4 @@
           artifacts: 'pkg/*'
           only-if-success: true
       - release_nightly_build_foreman:
-          project: foreman-proxy
+          project: packages/foreman/foreman-proxy
```
ekohl: This probably broke Debian packaging too

PR #465, opened 2018-02-20
Split release_nightly_build foreman into rpm and deb

ehelms:
@ekohl Thanks for the heads up. I split deb and rpm here to be explicit.


mmoll:
looking good visually, let's give this a whirl :+1:

PR #483, opened 2018-02-21
Use Puppet5 for Stretch

xprazak2:


mmoll:
merged, díky @xprazak2!

PR #485, opened 2018-02-21
allow for packaging being tested in a pipeline

evgeni:
the code is based on tfmROR51.groovy, which should also migrate to this
new lib as soon as obal supports copr properly

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy
@@ -1,0 +1,12 @@
+def ghprb_git_checkout() {
+    checkout changelog: true, poll: false, scm: [
+        $class: 'GitSCM',
+        branches: [[name: '${ghprbActualCommit}']],
+        doGenerateSubmoduleConfigurations: false,
+        extensions: [[$class: 'PreBuildMerge', options: [fastForwardMode: 'FF', mergeRemote: 'origin', mergeStrategy: 'default', mergeTarget: '${ghprbTargetBranch}']]],
+        submoduleCfg: [],
+        userRemoteConfigs: [
+            [refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*', url: 'https://github.com/${ghprbGhRepository}']
+        ]
+    ]
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
@@ -1,0 +1,22 @@
+def obal(body) {
+
+    def config = [:]
+    body.resolveStrategy = Closure.DELEGATE_FIRST
+    body.delegate = config
+    body()
+
+    def tags = config.tags ? "--tags ${config.tags}" : ""
+    def extra_vars = config.extraVars ?: [:]
+    def extra_vars_file = 'extra_vars.yaml'
+
+    writeYaml file: extra_vars_file, data: extra_vars
+    archive extra_vars_file
+
+    dir('obal') {
+        git url: "https://github.com/theforeman/obal.git", branch: "master"
+    }
+
+    wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
+        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${config.action} ${config.packages}"
+    }
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
@@ -1,0 +1,52 @@
+def packages_to_build
+
+pipeline {
+    agent { label 'rpmbuild' }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+                ghprb_git_checkout()
+
+            }
+        }
+
+        stage('Find Packages to Build') {
+            steps {
+
+                script {
+                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec' | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    packages_to_build = changed_packages.split().join(' ')
+                    update_build_description_from_packages(packages_to_build)
+                }
+
+            }
+        }
+
+        stage('Scratch Build Packages') {
+            steps {
+
+                obal {
+                    action = "scratch"
+                    tags = "wait,download"
+                    packages = packages_to_build
+                }
+
+            }
+        }
+    }
+}
+
+def update_build_description_from_packages(packages_to_build) {
+    build_description = "${packages_to_build}"
+    currentBuild.description = build_description
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-rpm-pr-test.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-rpm-pr-test.yaml
@@ -1,0 +1,15 @@
+---
+- job:
+    name: foreman-packaging-rpm-pr-test
+    project-type: workflow
+    properties:
+      - github:
+          url: https://github.com/theforeman/foreman-packaging
+    triggers:
+      - github_pr_rpm_develop:
+          context: 'rpm'
+    dsl:
+      !include-raw:
+        - pipelines/lib/rpm_packaging.groovy
+        - pipelines/lib/git.groovy
+        - pipelines/lib/obal.groovy
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/triggers/github_pr.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/triggers/github_pr.yaml
@@ -21,6 +21,14 @@
           <<: *ghprb_defaults
 
 - trigger:
+    name: github_pr_rpm_develop
+    triggers:
+      - github-pull-request:
+          <<: *ghprb_defaults
+          white-list-target-branches:
+            - rpm/develop
+
+- trigger:
     name: github_pr_pipeline
     triggers:
       - github-pull-request:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+    writeYaml file: extra_vars_file, data: extra_vars
+    archive extra_vars_file
+
+    dir('obal') {
+        git url: "https://github.com/theforeman/obal.git", branch: "master"
+    }
+
+    wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
+        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal/ python -m obal.__init__ ${tags} -e @${extra_vars_file} ${config.action} ${config.packages}"
```
ekohl: We should introduce a ```__main__``` so we can use ```python -m obal```

ekohl: https://github.com/theforeman/obal/pull/53

evgeni: tell me more? is it just adding a `__main__.py`?

evgeni: like this?
```python
# cat obal/__main__.py
import obal
if __name__ == '__main__':
    obal.main()
```

ekohl: See my obal PR :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,73 @@
...
+                    update_build_description_from_packages(packages_to_build)
+                }
+
+            }
+        }
+        stage('Scratch Build Packages') {
+            steps {
+
+                script {
+                    obal {
```
ehelms: This has to be in a `script` call even though its a function?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-packaging-pr-test.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- job:
+    name: tfm-packaging-pr-test
+    project-type: workflow
+    properties:
+      - github:
+          url: https://github.com/theforeman/foreman-packaging
+    triggers:
+      - github_pr_rpm_develop:
+          context: 'packaging'
```
ehelms: Will this conflict with the existing context for the current PR job?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
evgeni: This is not The Greatest Pipeline in the World, no.
This is just a tribute.

https://github.com/theforeman/foreman-infra/blob/4afb997715c82d9d76082f8c6567fe0b6f4a75cd/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51.groovy#L45-L51

But I'll try without.

ehelms: +1 I think that was originally mis-guided and custom functions allow you to avoid the `script` wrapper.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-packaging-pr-test.yaml
evgeni: It will indeed. Shall we use another name for now? `obal`? `rpm`?

ehelms: I lean towards rpm for the time being

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-packaging-pr-test.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- job:
+    name: tfm-packaging-pr-test
```
ehelms: Should this be `foreman-packaging` to reflect the pattern of repository name?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/tfm-packaging-pr-test.yaml
```diff
@@ -0,0 +1,14 @@
...
+    project-type: workflow
```
ehelms: Should this be `pipeline` out of the gate since `workflow` is deprecated?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+pipeline {
+    agent { label 'rpmbuild' }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+    }
```
ehelms: So, I've learned that these seem to not always work when deployed through JJB. We can keep them but let's keep an eye on it.

evgeni: we'll see how that goes I guess

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,61 @@
...
+                ]
+
+            }
+        }
+
+        stage('Find Packages to Build') {
+            steps {
+
+                script {
+                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec' | cut -d'/' -f2 |sort -u").trim()
```
ekohl: Rather than cut, would dirname + basename be possible? That wouldn't assume a directory structure making it work for older branches too.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,61 @@
+def packages_to_build
```
ekohl: Should this be named ```rpm_packaging``` or ```packaging_rpm```?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- job:
+    name: foreman-packaging-pr-test
+    project-type: pipeline
+    properties:
+      - github:
+          url: https://github.com/theforeman/foreman-packaging
+    triggers:
+      - github_pr_rpm_develop:
```
ekohl: Can you also set that it only triggers on ```rpm/*``` (unless you start handling debian packaging too)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- job:
+    name: foreman-packaging-pr-test
```
ekohl: pr-test-rpm?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
ekohl: While reviewing I see that you do that in the trigger. Could we use ```github_pr``` and override the white-list-target-branches here inline or am I asking too much?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
ehelms: I'd prefer the `rpm` part being earlier, something like `foreman-packaging-rpm-pr-test` to stay in line with conventions we established on the mailing list.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,61 @@
...
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+                checkout changelog: true, poll: false, scm: [
```
ehelms: I think we can generalize this now or as a follow up item as its used in 3-4 places now we've got a pattern established

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
evgeni: why? that will contain the list of packages to build later.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
evgeni: I have no idea how to do that, but patches welcome ;)

ehelms: I think it could be turned into a variable, every job would have to specify '*' though. So there is a tradeoff as far as my understanding of yaml goes.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/foreman-packaging-pr-test.yaml
evgeni: k

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
evgeni: done

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
ekohl: I was looking at the filename but failed to write ```file``` in that sentence.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+    writeYaml file: extra_vars_file, data: extra_vars
+    archive extra_vars_file
+
+    dir('obal') {
+        git url: "https://github.com/theforeman/obal.git", branch: "master"
+    }
+
+    wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
+        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal/ python -m obal ${tags} -e @${extra_vars_file} ${config.action} ${config.packages}"
```
ekohl: I think the ```/``` in PYTHONPATH is not needed

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy
```diff
@@ -0,0 +1,12 @@
+def github_checkout() {
```
ekohl: How about ```ghprb_git_checkout```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/packaging.groovy
evgeni: fair enough.

evgeni:
@ehelms @ekohl I think I adressed all (but the trigger params) comments.

ekohl:
Previously we had a limit of 15 packages. Is that still a good thing to have?

ehelms:
@ekohl That's a really good question and I think we can get away with either a much higher hard limit or none at all for these two reasons:

 1) Our koji is quite fast these days and overload is not a worry
 2) This change means only a single job slot is ever used. Now while this will hold the job slot until it completes, it will not back up the job queue with 500 jobs like the current model does

Theoretically, we could even pursue a parallel pipeline run in the future to increase performance.

ekohl:
If it's serial, then I don't think a limit is needed.

PR #487, opened 2018-02-23
Only retrieve the relevant PR with GHPRB

ekohl:
By setting this refspec we only retrieve the PR that we need to build
rather than every single PR. This saves a bandwidth, IO and thus time,
especially as the repository gets older.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman.yaml
@@ -15,4 +15,4 @@
           wipe-workspace: true
           branches:
             - '${ghprbActualCommit}'
-          refspec: '+refs/pull/*:refs/remotes/origin/pr/*'
+          refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/origin/pr/${ghprbPullId}/*'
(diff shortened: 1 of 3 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman.yaml
```diff
@@ -15,4 +15,4 @@
           wipe-workspace: true
           branches:
             - '${ghprbActualCommit}'
-          refspec: '+refs/pull/*:refs/remotes/origin/pr/*'
+          refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/upstream/pr/${ghprbPullId}/*'
```
ehelms: Copy/paste? This should still be origin instead of upstream.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/kafo.yaml
```diff
@@ -14,4 +14,4 @@
           wipe-workspace: true
           branches:
             - '${ghprbActualCommit}'
-          refspec: '+refs/pull/*:refs/remotes/origin/pr/*'
+          refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/upstream/pr/${ghprbPullId}/*'
```
ehelms: Copy/paste? This should still be origin instead of upstream.

ekohl:
Updated

PR #450, opened 2018-02-08
Add scaffolding for using ci.centos.org and an initial job

ehelms:
The initial job being added is a PR test for Forklift containers.

PR #488, opened 2018-02-23
Update tfm-ror51 to use Obal

ehelms:


ehelms:
This has some overlap with https://github.com/theforeman/foreman-infra/pull/485

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -0,0 +1,22 @@
+def obal(body) {
```
evgeni: this file changed a bit in #485, mind rebasing? ;-)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror51-release.yaml
```diff
@@ -10,4 +10,4 @@
     dsl:
       !include-raw:
         - pipelines/release/tfmROR51.groovy
-        - pipelines/lib/ansible.groovy
```
evgeni: #485 added a `git.groovy`, mind using it here too?

evgeni:
This needs a rebase on top of https://github.com/theforeman/foreman-infra/pull/493, LGTM otherwise

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror51-release.yaml
ehelms: This is the release job which uses a simple git checkout. The PR test job above is using it.

ehelms:
Updated and now passing -- https://github.com/theforeman/tfm-ror51-packaging/pull/44

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -5,7 +5,6 @@ def obal(args) {
     extra_vars = args.extraVars ?: [:]
 
     writeYaml file: extra_vars_file, data: extra_vars
-    archive extra_vars_file
```
evgeni: Why?

evgeni:
@ekohl the file itself? technically yes, but I think I'd keep it if we need it at some other place.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
ehelms: So, this function is generic, and when you use it twice in one job this file changes. So it was not clear to me what archiving would do for us other than potentially add confusion given it would only archive the last one in the end.

ehelms:
I'd make use of the `ansible.groovy` here as well -- https://github.com/theforeman/foreman-infra/pull/450/files

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
evgeni: Oooh, so it does not archive multiple versions of it? Stupid. Yeah, then it's kinda useless.

Issue #491, opened 2018-02-26
stop using schedule-failed-build plugin


evgeni:
The plugin is [deprecated](https://wiki.jenkins.io/display/JENKINS/Retry+Failed+Builds+Plugin) and has not seen an [upstream commit since 2012](https://github.com/jenkinsci/schedule-failed-builds-plugin).

Options:
* [Naginator Plugin](https://wiki.jenkins.io/display/JENKINS/Naginator+Plugin)
* [Periodic Reincarnation Plugin](https://wiki.jenkins.io/display/JENKINS/Periodic+Reincarnation+Plugin)

evgeni:
One question that arises is: do we even use this plugin now? I know it's referenced in the jobs, but I could not find any occurrence of it actually re-scheduling a failed job.

ekohl:
Let's try to get rid of it altogether. For CI it doesn't make sense and master builds are probably covered by nightly runs anyway.

PR #492, opened 2018-02-26
use upstream jjb and drop the use of schedule-failed-builds

evgeni:
Closes: #491

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ addons:
     packages:
     - yajl-tools
 install:
-  - git clone --depth 1 -b local git://github.com/theforeman/jenkins-job-builder.git
-  - (cd jenkins-job-builder && pip install --user -r requirements.txt && pip install --user .)
+  - pip install --user jenkins-job-builder
   - (cd puppet && bundle install)
 script: ./.travis.sh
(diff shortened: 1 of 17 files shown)


mmoll:
I have no idea what I'm doing, but this looks legit :+1:

mmoll:
merged, спасибо Женя!

evgeni:
Now I wonder if that should've been `ensure: latest` to actually update from PyPI.

mmoll:
hrm... on a related note: `Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Jenkins_job_builder]: has no parameter named 'git_revision' on node master01.rackspace.theforeman.org`

ekohl:
> Now I wonder if that should've been ensure: latest to actually update from PyPI.

I considered that too but not sure that's wise. Maybe wiser to make it a param and pin it to an explicit version?

evgeni:
Oups. Can you remove the two git_ params from Foreman?

mmoll:
I think this needs to get removed in `puppet/modules/jenkins_job_builder/manifests/*`

mmoll:
Aw, sorry... found it in an unexpected place. Fixed.

evgeni:
That's where I removed it ;)

PR #493, opened 2018-02-26
Switch obal function to named parameters

ehelms:
My understanding is that with declarative pipelines, this style is more the norm and can avoid this error:

```
WorkflowScript: 39: Expected a step @ line 39, column 21.
                       action = "scratch"
```

Which happens if a closure is not wrapped in a `script {}` which is kinda ugly to double wrap.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
@@ -1,13 +1,8 @@
-def obal(body) {
-
-    def config = [:]
-    body.resolveStrategy = Closure.DELEGATE_FIRST
-    body.delegate = config
-    body()
-
-    def tags = config.tags ? "--tags ${config.tags}" : ""
-    def extra_vars = config.extraVars ?: [:]
+def obal(args) {
     def extra_vars_file = 'extra_vars.yaml'
+
+    tags = args.tags ? "--tags ${args.tags}" : ""
+    extra_vars = args.extraVars ?: [:]
 
     writeYaml file: extra_vars_file, data: extra_vars
     archive extra_vars_file
@@ -17,6 +12,6 @@ def obal(body) {
     }
 
     wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
-        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${config.action} ${config.packages}"
+        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${args.action} ${args.packages}"
     }
 }
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
@@ -24,7 +24,13 @@ pipeline {
             steps {
 
                 script {
-                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec' | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec'")
+
+                    if (changed_packages) {
+                        changed_packages = sh(returnStdout: true, script: "echo ${changed_packages} | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    } else {
+                        changed_packages = ''
+                    }
                     packages_to_build = changed_packages.split().join(' ')
                     update_build_description_from_packages(packages_to_build)
                 }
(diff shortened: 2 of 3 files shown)


ehelms:
This is updated and happy now testing on CI

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -1,22 +1,19 @@
...
     writeYaml file: extra_vars_file, data: extra_vars
     archive extra_vars_file
 
     dir('obal') {
         git url: "https://github.com/theforeman/obal.git", branch: "master"
     }
 
-    wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
-        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${config.action} ${config.packages}"
+    if (args.packages) {
```
ehelms: This skips past the Obal run if there are no packages to build. We could handle this differently, but for non-package changes this at least gets a green pipeline.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -24,7 +24,13 @@ pipeline {
             steps {
 
                 script {
-                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec' | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec'")
+
+                    if (changed_packages) {
+                        changed_packages = sh(returnStdout: true, script: "${changed_packages} | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    } else {
+                        changed_packages = ''
```
ehelms: This splits the processing of changed_packages into two pieces to ensure that if nothing changes in `packages/` then it won't try to process that further.

ekohl:
What about changes in ```rel-eng```? That can affect builds but it's hard to determine which package needs to be rebuild. I don't think we do this now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
evgeni: I'd argue that "no packages" should result in a "skipped" build, not a "green" one. (But at the same time, green is better than red)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -24,7 +24,13 @@ pipeline {
             steps {
 
                 script {
-                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec' | xargs dirname | xargs -n1 basename |sort -u").trim()
+                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec'")
+
+                    if (changed_packages) {
+                        changed_packages = sh(returnStdout: true, script: "${changed_packages} | xargs dirname | xargs -n1 basename |sort -u").trim()
```
evgeni: should that be `echo ${changed_packages}`?

evgeni: it should https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-pr-test/detail/foreman-packaging-rpm-pr-test/22/pipeline/

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -1,22 +1,21 @@
...
+        wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
+            sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${args.action} ${args.packages}"
+        }
+    } else {
+        githubNotify status: 'ABORTED'
```
ehelms: @evgeni So we can try this and see?

ehelms: Actually, this is ugly here

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -33,17 +39,26 @@ pipeline {
...
             }
         }
     }
+
+    post {
+        always {
+
+            if (!packages_to_build) {
+                githubNotify status: 'ABORTED'
+            }
```
ehelms: Trying it here.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -33,17 +39,26 @@ pipeline {
         }
 
         stage('Scratch Build Packages') {
+            when {
+                expression { packages_to_build }
+            }
```
ehelms: This should just skip

ehelms:
Updated to remove the ABORT since github does not actually support that.

PR #497, opened 2018-02-27
Rewrite kafo PR tests to a pipeline

ekohl:
Please review carefully. This is the first time I convert something to a pipeline :)

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/kafo.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/kafo.groovy
@@ -1,0 +1,66 @@
+pipeline {
+    agent none
+    options {
+        timeout(time: 1, unit: 'HOURS')
+    }
+
+    stages {
+        stage('Test') {
+            matrix {
+                agent any
+                axes {
+                    axis {
+                        name 'ruby'
+                        values '2.0.0', '2.1', '2.2', '2.3', '2.4', '2.5', '2.6'
+                    }
+                    axis {
+                        name 'PUPPET_VERSION'
+                        values '4.10', '5.0', '6.0'
+                    }
+                }
+                excludes {
+                    exclude {
+                        axis {
+                            name 'PUPPET_VERSION'
+                            values '5.0', '6.0'
+                        }
+                        axis {
+                            name 'ruby'
+                            values '2.0.0', '2.1', '2.2', '2.3'
+                        }
+                    }
+                }
+                stages {
+                    stage('Setup Git Repos') {
+                        steps {
+                            ghprb_git_checkout()
+                        }
+                    }
+                    stage("Setup RVM") {
+                        steps {
+                            configureRVM(ruby)
+                        }
+                    }
+                    stage('Install dependencies') {
+                        steps {
+                            withRVM(['bundle install'], ruby)
+                        }
+                    }
+                    stage('Run Tests') {
+                        steps {
+                            withRVM(['bundle exec rake jenkins:unit TESTOPTS="-v" --trace'], ruby)
+                        }
+                    }
+                }
+                post {
+                    always {
+                        archiveArtifacts artifacts: "Gemfile.lock"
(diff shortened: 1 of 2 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
```diff
@@ -0,0 +1,35 @@
+pipeline {
+    agent { label 'fast' }
```
mmoll: I think slow nodes will do ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
```diff
@@ -0,0 +1,35 @@
...
+
+    stages {
+        stage('Setup Git Repos') {
+            steps {
+                ghprb_git_checkout()
+            }
+        }
+        stage("Setup RVM") {
+            steps {
+                configureRVM(ruby)
```
evgeni: `ruby` is undef?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
ekohl: In the past we've seen the job timeout and fail so I wasn't sure.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
ekohl: Doesn't that come from the job parameters?

evgeni: ah no it comes from axis?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
mmoll: indeed, but that isn't related to compute power, but (AFAICT) some timeout when facter wants to connect to some Amazon API when resolving a fact.

ekohl: It could be IO related, but I'll remove it for now since we need to get the job working again anyway.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
ekohl: That's what I expect

ekohl:
Doesn't look like it's actually testing something. Maybe pipeline and axis is incompatible?

ehelms:
I think our goal should be to avoid axis with Pipelines and I don't think it supports it. We should strive for using parallel to achieve the same concept.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testRuby.groovy
```diff
@@ -0,0 +1,33 @@
+pipeline {
+    stages {
```
ehelms: I think you want an agent definition above this?

ekohl:
I tried to look at this, but changing it to a parallel pipeline looked tricky with tests and gemfile.lock. Moving to ```WIP``` and going to focus on other things first.

mmoll:
What's the status here?

ekohl:
I updated it using https://jenkins.io/blog/2019/11/22/welcome-to-the-matrix/ but currently still testing.

ekohl:
Should be good for review now. When deploying this the job needs to be deleted since JJB can't change the job type.

mmoll:
looks straightforward :green_heart:

mmoll:
merged, bedankt @ekohl!

PR #499, opened 2018-02-27
add koji links as status to the GitHub PR

evgeni:


ekohl:
This assumes there's only one koji build per commit, which is not always true (SCL + non-SCL, multiple changes in 1 commit). Not sure how to solve it otherwise though.

evgeni:
@ekohl it does not. you get all the links

PR #504, opened 2018-03-01
Deploy koji configuration as koji file

ehelms:


puppet/modules/slave/manifests/init.pp
```diff
@@ -400,12 +400,12 @@
...
     source => 'puppet:///modules/slave/titorc',
   }
-  file { '/home/jenkins/.koji/katello-config':
+  file { '/home/jenkins/.koji/koji':
     ensure => file,
     mode   => '0644',
     owner  => 'jenkins',
     group  => 'jenkins',
-    source => 'puppet:///modules/slave/katello-config',
+    source => 'puppet:///modules/slave/koji',
```
ekohl: If you change the source here, then you also need to rename the file

ehelms:
This follows the documented model for users now.

puppet/modules/slave/manifests/init.pp
```diff
@@ -400,7 +400,7 @@
     group  => 'jenkins',
     source => 'puppet:///modules/slave/titorc',
   }
-  file { '/home/jenkins/.koji/katello-config':
```
ekohl: I'd like an ```ensure => absent``` on this.

puppet/modules/slave/manifests/init.pp
```diff
@@ -435,6 +435,12 @@
     owner  => 'jenkins',
     group  => 'jenkins',
   }
+  file { '/home/jenkins/.bin/kkoji':
```
ekohl: is that dot intended to be there?

ehelms:
@ekohl Good catch! I updated the one script I found using koji-katello which I think covers it.

PR #505, opened 2018-03-01
[wip] convert `deploy_yum` job to JJB

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
```diff
@@ -0,0 +1,33 @@
...
+          default: nightly
+          description: 'Relative path on yum.tf.org to sync to, e.g. <code>nightly</code> or <code>releases/1.2</code>'
+      - boolean:
+          name: merge
+          default: false
+          description: 'When true, will only add to and not delete the existing repo, then will run <code>createrepo</code> to merge the new packages into the existing repo.'
+      - string:
+          name: ruby
+          default: '2.0.0'
+          description: 'Version of Ruby to run deploy script under'
```
ehelms: This param is not used anywhere, and while it was on the original, I think we can just drop it.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
```diff
@@ -0,0 +1,33 @@
...
+    project-type: workflow
+    logrotate:
+      daysToKeep: 2
+      numToKeep: -1
+    triggers:
+      - timed: "" # ???
+    paramters:
+      - string:
+          name: os
+          default: el6
```
ehelms: Let's update this OS to el7

ehelms:
With (https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanRelease.groovy#L60) there is an opportunity for re-factoring some stuff out to a method.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
```diff
@@ -0,0 +1,33 @@
+- job:
+    name: deploy_yum
+    project-type: workflow
+    logrotate:
+      daysToKeep: 2
+      numToKeep: -1
+    triggers:
+      - timed: "" # ???
```
zjhuntin: I don't actually think you need triggers at all if you have nothing use with it.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
ekohl: But the commands do use ```withRvm``` so either it needs to fully go to rvm or drop it.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
ekohl: description could be updated as well then

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
komidore64: removed.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
zjhuntin: Currently withRVM defaults to use 2.0.0 so I think it can be dropped.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/deploy_yum.yaml
```diff
@@ -0,0 +1,31 @@
+- job:
+    name: deploy_yum
+    project-type: workflow
+    logrotate:
+      daysToKeep: 2
+      numToKeep: -1
+    paramters:
+      - string:
+          name: os
+          default: el6
```
ehelms: Update to EL7 please

ehelms:
I'd consider putting the job under `pipelines/release` since this is used for releasing

ehelms:
Besides the minor comments, I think this is GTG

zjhuntin:
Agreed, ACK as well pending other discussions.

komidore64:
commits squashed

ehelms:
ACK - thanks, we will want to keep an eye on the job during the release process to fix any bugs

PR #506, opened 2018-03-01
Use ./koji/config in .titorc

ehelms:


ehelms:
No idea, but let's try it

PR #507, opened 2018-03-02
Add release_build_hammer_cli_katello job

zjhuntin:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/release_build_hammer_cli_katello.yaml
```diff
@@ -0,0 +1,11 @@
+job:
+  name: release_build_hammer_cli_katello
+  project_type: workflow
+  logrotate:
+    daysToKeep: 2
+    numToKeep: -1
+  dsl:
+    !include-raw:
+      - pipelines/release/releaseBuildHammerCliKatello.groovy
+  archive:
```
ehelms: This aspect can be done in the pipeline code using a post:

```
post {
  always {
     archive "stoof"
  }
}
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseBuildHammerCliKatello.groovy
```diff
@@ -0,0 +1,28 @@
...
+            steps {
+                git url: "https://github.com/katello/hammer-cli-katello"
+            }
+
+        stage ("Remove old Gem")
+            steps {
+                sh "rm hammer_cli_katello-*.gem"
+            }
+
+        stage ("Build Gem")
```
ehelms: I think we are preferring to use full `{}` syntax with stage.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseBuildHammerCliKatello.groovy
```diff
@@ -0,0 +1,28 @@
...
+    stages {
+
+        stage ("Setup Repository")
+            steps {
+                git url: "https://github.com/katello/hammer-cli-katello"
+            }
+
+        stage ("Remove old Gem")
+            steps {
+                sh "rm hammer_cli_katello-*.gem"
```
ekohl: Should this be ```-f``` so the build doesn't fail when these files aren't present?

zjhuntin:
The no triggers, is my fault. It's supposed to be run daily. That being said, I still think that your point of getting it on Rubygems.org instead is far better. I'm willing to scrap the job instead of remaking it.

ehelms:
This implies doing a full on release if we are pushing to rubygems.org which would imply daily build and release tagging. I don't think that's what we want in this case? This, the installer, rubygems-katello, foreman, foreman-proxy nightlies are all about just building the test code regularly and having it available to ward against breakages. Requiring releases to rubygems.org seems like it muddies that?

zjhuntin:
The name of the job implies that it is building the gem for a release. Not a snapshot of where the project is right now, which is what i think may have confused me and @ekohl . If that's not the intent, should we just rename the job to `hammer_cli_katello_daily_build`? I still think that having a travis job to automatically upload a gem upon a new tag being pushed is still worth while, but something to pushed into the backlog.

ehelms:
I wonder if we shouldn't just combine this and http://ci.theforeman.org/job/build_rpm_rubygem-hammer_cli_katello_nightly/ into a single job that builds + releases the gem to the proper places? We could do the same for installers and other gems.

ehelms:
Rather than kick of the job this would be the job that builds the nightly RPM in Koji.

zjhuntin:
@ehelms @ekohl Okay, I've attempted to achieve the goals we laid out before with this. I don't think it's 100% going to work this way, but I also do not have a great way of testing it currently. Is there a way y'all would suggest testing locally other than building the job?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRpmBuild.groovy
```diff
@@ -0,0 +1,52 @@
...
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+
+        stage ("Build Gem") {
+            steps {
+
+                dir('hammer_cli_katello') {
```
ehelms: It's typically better to have the first thing cloned not be in a directory unless you put all things cloned into individual directories.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRpmBuild.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+                    git url: "https://github.com/katello/hammer-cli-katello", branch: 'master'
+                    sh "gem build hammer_cli_katello.gemspec"
```
ehelms: I think you could archive this at this point and then reference it below.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRpmBuild.groovy
```diff
@@ -0,0 +1,52 @@
...
+            steps {
+                git url: "https://github.com/theforeman/foreman-Packaging", branch: "rpm/develop"
+                sh "mv ./hammer_cli_katello/*.gem ./packages/katello/rubygem-hammer_cli_katello/"
+            }
+        }
+
+        stage ("Make Scratch Build") {
+            steps {
+
+                obal(
```
ehelms: I think if we take the same idea as (https://github.com/theforeman/foreman-packaging/blob/rpm/develop/package_manifest.yaml#L283) and add it to the rubygem-hammer_cli_katello entry, when the gem is archived in the step above, we could self reference this job to get the artifact and let the FetchBuilder do the rest.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -1,78 +1,19 @@
-def packages_to_build
-
```
ehelms: Did you intend to edit this file? This does not appear used by the job.

zjhuntin: Uh, I honestly have no idea why this is showing up here, I didn't change it, but it appears to be this way in the repo, and when I try to check it out from origin nothing changes.

zjhuntin:
@ehelms fixed

komidore64:
i'd like to suggest renaming `hammer_cli_katello_nightly_release.yaml` to `hammer-cli-katello-nightly-release.yaml` so that the filename exactly matches the job name.

zjhuntin:
Done!

PR #508, opened 2018-03-02
Add docker module and configure slave to install and run docker

ehelms:
This is untested and a bit of a leap of faith that the module will just work and configure the slaves that can support docker properly for us. According to (https://github.com/puppetlabs/puppetlabs-docker#description) this should configure on most of our slaves and I assume ignore the ARM slaves. This is my first time trying to add additional puppet modules to our infrastructure so please be gentle and let me know what thoughts are towards any additional config this might need.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -128,3 +128,6 @@
 [submodule "puppet/modules/translate"]
 	path = puppet/modules/translate
 	url = https://github.com/puppetlabs/puppetlabs-translate.git
+[submodule "puppet/modules/docker"]
+	path = puppet/modules/docker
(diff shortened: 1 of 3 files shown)


mmoll:
@GregSutcliffe that module could also be of use for the discourse install, I presume?

ehelms:
@ekohl Any further thoughts here? I assume this will just work but I anticipate having to deal with a few issues.

mmoll:
yeah, let's pull the trigger on that :gun:

mmoll:
merged, thanks @ehelms!

PR #509, opened 2018-03-02
set PYTHONPATH to an absolute path

evgeni:


evgeni:
I might end up using better syntax sugar

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -11,7 +11,7 @@ def obal(args) {
     }
 
     wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
-        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=obal python -m obal ${tags} -e @${extra_vars_file} ${args.action} ${args.packages}"
+        sh "ANSIBLE_FORCE_COLOR=true PYTHONPATH=\"\$(pwd)/obal\" python -m obal ${tags} -e @${extra_vars_file} ${args.action} ${args.packages}"
```
ekohl: Isn't there an env var for the workspace directory?

evgeni: maybe, `pwd()` from groovy works too :)

PR #511, opened 2018-03-02
Use standard koji config for tito

ehelms:


ehelms:
Attempting to fix (http://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-rpm-pr-test/detail/foreman-packaging-rpm-pr-test/79/pipeline/) and this change worked on the slave for tito. There may still be an issue with using koji directly but that's easier to handle and I think from testing this will work out of the box.

PR #514, opened 2018-03-02
[WIP] define a pipeline to release rpm packages from f-packaging

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -15,16 +15,26 @@ pipeline {
...
             steps {
 
                 script {
-                    changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec'").trim()
+                    def changed_packages
+                    if (package_action == 'scratch') {
+                        changed_packages = sh(returnStdout: true, script: "git diff origin/${ghprbTargetBranch} --name-only -- 'packages/**.spec'").trim()
+                    }
+                    else if (package_action == 'release') {
+                        changed_packages = sh(returnStdout: true, script: "git diff ${env.beforeCommit}...${env.headCommit} --name-only -- 'packages/**.spec'").trim()
```
ehelms: Looks like a re-usable function now that just needs passing in the thing to be diffed

ehelms: Those env variables set by the git or github plugin?

evgeni: no, those are randomly chosen params, that have to be filled in by a greater god, erm, `prprocessor` as the plugins don't know what we are building :(

ehelms:
When I did `tfm-ror51-release` I had to add (https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror51-release.yaml#L8) along with a webhook.

evgeni:
@ehelms the snippet from `tfm-ror51-release` uses [`jjb.triggers.github`](https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.github), which uses the [GitHub Plugin](https://wiki.jenkins.io/display/JENKINS/GitHub+Plugin). That will happily build what it's told to, but it does not propagate the `before` param that GitHub includes in the hook, thus making it impossible for us to know what we're actually building.

evgeni:
obsolete, as #745 has been merged

PR #515, opened 2018-03-02
Use sha1 for pull request builder

ehelms:
Attempting to fix (http://ci.theforeman.org/job/katello-pr-test/746/console) the test case we ran worked but was only a single data point.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy
@@ -1,7 +1,7 @@
 def ghprb_git_checkout() {
     checkout changelog: true, poll: false, scm: [
         $class: 'GitSCM',
-        branches: [[name: '${ghprbActualCommit}']],
+        branches: [[name: '${sha1}']],
         doGenerateSubmoduleConfigurations: false,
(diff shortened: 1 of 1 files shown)


ekohl:
The question is now: what are we actually building? Is this really the commit that was pushed? We may need better error reporting but this may actually hide errors.

ehelms:
Help me out, how does this hide errors or not ensure its the latest commit on the PR? The alternative is probably to drop the pullId from the refspec and fetch all PRs.

ekohl:
The case as I understand it (and this might be just me rubber ducking):
* author pushed a commit [abc]
* webhook was triggered queuing a job
* author pushed a new commit [xyz]
* job starts to run

Now the job tries to build [abc] but can no longer find it. What this change implies is that [xyz] is now built. Will this put the build status on commit [abc] or [xyz] though?

ehelms:
Should (https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/triggers/github_pr.yaml#L15) handle that scenario in theory? I'd think your scenario would be a problem for everyone using GHPRB given their docs mention {sha1} before they mention using ghprbActualCommit

akofink:
@ekohl In your scenario, you said that [abc] would no longer exist, but that's not true, afaiu - git never actually deletes commits, even if a branch no longer points to it. Am I missing something here?

PR #517, opened 2018-03-03
Ignore db:drop failures in gemset_cleanup

ekohl:
This is not available in every job and can sometimes fail the build
while new builds already run db:drop at the start.


mmoll:
merged, bedankt @ekohl!

PR #518, opened 2018-03-03
Switch to using vagrant-openstack-provider

ekohl:
This PR contains two experiments:

* Add spec tests to our repository
* Use of the Sensitive data type

The plugin also doesn't support API keys which appears to be a rackspace openstack extention. It's probably cleaner to create a subuser that doesn't have full user permissions.

That could use some feedback.

Other than that we'll need https://github.com/ggiamarchi/vagrant-openstack-provider/pull/356 to actually run. I'm not sure yet how best to include that patch. Probably manually build a gem and install that.


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanRelease.groovy
```diff
@@ -28,9 +28,6 @@ pipeline {
         stage('Install Test') {
             agent { label 'el' }
 
-            environment {
-                VAGRANT_DEFAULT_PROVIDER = 'rackspace'
```
ehelms: What handles this now?

ekohl: This is part of an experiment, but I'm hoping that the Vagrantfile in the homedir helps to set the precedence. I'll try this on a VM to see if it works.

puppet/Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'puppet-syntax'
-gem 'puppet', '~> 4.10'
+gem 'puppet', '~> 4.10', :require => false
+gem 'puppet-syntax', :require => false
```
ekohl: Technically this line is now redundant.

mmoll:
The first two commits could IMHO go in independently at any time and probably also the pure Vagrant part of the third commit.

For the provider switch, let's see how responsive the vagrant-openstack-provider maintainers are... ;)

ekohl:
> probably also the pure Vagrant part of the third commit.

The problem is that vagrant-rackspace is incompatible with modern vagrant versions. That's why I'm switching the provider.

ekohl:
This should now be ready to go. Changes:
* It now installs the gem from download.theforeman.org until we have a release including the fix.
* travis now runs the puppet tests.

ekohl:
For context: currently systest jobs are broken which is blocking the 1.17RC2 build.

ekohl:
Rebased. I changed the vagrant default provider from removing it to setting it to openstack as well.

PR #519, opened 2018-03-05
Use koji instead of kkoji

ehelms:


komidore64:
Why is this change needed?

ehelms:
The `kkoji` symlink we added doesn't actually work, and we adjusted the config for koji to be `.koji/config` which means standard `koji` will use the proper configuration value. This fixes the jobs that still use this bash script such as katello RPM nightly build.

komidore64:
awesome!

ACK

ekohl:
@komidore64 The way I understand it is that koji looks for a section as the binary is called. If it's called as ```kkoji```, it looks for ```[kkoji]``` in ```~/.koji/config```. If it's called as ```koji```, then ```[koji]```. Our config has ```[koji]```, so we should call it as koji.

komidore64:
thank yall for the clarification

ehelms:
@komidore64 Thanks for the question!

ehelms:
@ekohl I *think* so, once this is merged I'll monitor and remove it if so.

PR #523, opened 2018-03-06
Use distro docker

ehelms:


ehelms:
@ekohl Carefully triple check me :)

ekohl:
The README has a special example on doing this:

```puppet
class { 'docker':
  use_upstream_package_source => false,
  service_overrides_template  => false,
  docker_ce_package_name      => 'docker',
}
```

It appears the ```yumrepo { 'docker': }``` has no ```ensure``` option so it needs to be done manually. Perhaps you also need to purge the ```docker-ce``` package and ```/etc/systemd/system/docker.service.d/service-overrides.conf``` for it to work.

ehelms:
Ugg, it won't purge that itself? Lame. K.

ehelms:
Updated with the recommendations -- please triple check

ehelms:
Thought I had done that, forget puppet is not always linear to how the code reads. Updated now with dependency arrows.

evgeni:
travis says "dependency cycles found: (File[/etc/systemd/system/docker.service.d/service-overrides.conf] => File[/etc/yum.repos.d/docker.repo] => Class[Docker] => Class[Docker::Service] => File[/etc/systemd/system/docker.service.d] => File[/etc/systemd/system/docker.service.d/service-overrides.conf])" :(

ehelms:
Given the dependency cycle, I pulled just the `service-overrides.conf` out of the chain and left the others. I think this should still suffice when it runs.

ehelms:
@ekohl are you ok with the minor update?

puppet/modules/slave/manifests/docker.pp
```diff
@@ -1,5 +1,19 @@
 class slave::docker {
 
-  include docker
+  file { '/etc/systemd/system/docker.service.d/service-overrides.conf':
+    ensure => absent
```
ekohl: lint likes trailing comma's

PR #524, opened 2018-03-06
fix `Undefined section "option"`

komidore64:


mmoll:
merged, thanks @komidore64!

PR #525, opened 2018-03-06
Ensure the correct vagrant version is installed

ekohl:


mmoll:
`Provider dpkg must have features 'versionable' to set 'ensure' to '2.0.2'` :crying_cat_face:

ekohl:
I only tested this with yum :( I wonder if latest would work with local files

ekohl:
Updated to use latest instead. This appears to work.

mmoll:
merged, bedankt @ekohl!

PR #527, opened 2018-03-07
Wipe out old vagrant dirs

ekohl:
When upgrading from 1.4.2 to 2.0.2 vagrant wants to upgrade some
internal data. This shows an interactive prompt which needs a TTY. We
don't care about our boxes so it's easier to just remove the directory.
We only do this if we see the old rackspace dummy box to make it
idempotent.


ekohl:
I did this manually on all but 2 slaves. I can't log into the netways slave so here's the puppet version.

mmoll:
I wonder what's going on with these machines... anyway, merged, bedankt @ekohl!

PR #531, opened 2018-03-07
move git checkout into the pipeline

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/deployPuppet.groovy
```diff
@@ -1,14 +1,17 @@
 pipeline {
     agent { label 'admin && sshkey' }
-
     options {
         timestamps()
         disableConcurrentBuilds()
         ansiColor('xterm')
     }
-
```
ehelms: Oh sorry, do you mind if we keep these blank lines? I think at least myself and @evgeni find them to be useful for visually separating out major code sections

komidore64:
Upon further inspection, I don't believe this change is necessary

PR #532, opened 2018-03-07
convert `release_tarballs` to JJB and workflow

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
```diff
@@ -0,0 +1,61 @@
...
+
+void verify_tag(project) {
+    ext_job('release_verify_tag', project)
+}
+
+void build_tarball(project) {
+    ext_job('release_build_tarball', project)
+}
+
+void ext_job(project, job) {
```
ehelms: What is ext short for?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
```diff
@@ -0,0 +1,61 @@
+pipeline {
+    options {
```
ehelms: I think we will want an `agent` declaration always at the top of this to ensure the job runs on a slave.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
```diff
@@ -0,0 +1,61 @@
...
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        parallel {
```
ehelms: Can parallel be inside a `stages` ? The only examples I ever see have it inside a `stage`

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
```diff
@@ -0,0 +1,61 @@
...
+            stage('Foreman') {
+                steps {
+                    verify_tag('foreman')
+                    build_tarball('foreman')
```
ehelms: The logic in these two jobs appears to be silly simple, enough that instead of continuing the generic job pattern, I think we are better off just creating functions for them and dropping those jobs. Just looking at `verify_tag`:

```
#!/bin/bash -xe

if [ "$(git tag -l ${version} | wc -l)" -ne 1 ]; then
  echo "ERROR: cannot find tag ${version}"
  git tag -l
  exit 1
fi

exit 0
```

That makes more sense to me as a function, gets rid of a job and makes this job more self contained. The same is likely true for `build_tarball`. Given this is a re-write of the `release_tarballs` job anyway, might as well take it the small step forward and avoid the code that triggers another job.

ekohl: :+1: on that. It also helps with the overview to have one pipeline rather than many subjobs.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
komidore64: "external job" as in a job outside of this pipeline.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
komidore64: @ehelms from conversation earlier this week, I thought it was agreed upon that for now jobs would just be converted 1:1 rather than conversions along with refactoring.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
komidore64: i thought i saw that somewhere ... but now i can't find it.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
komidore64: can stages still run in parallel if they're all on the same node?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
komidore64: okay, yeah i can't find it. i'll switch it up.

komidore64:
updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/releaseTarballs.groovy
ehelms: For moving jobs into source control a 1:1 conversion is what I was intending since its a net no change. If we are going to re-factor the job (which converting things to a pipeline is) then that's not 1:1 and we might as well include reasonable updates to fit the new style since new code is being written. If this is not exactly clear, we can clarify it more when we sync up.

komidore64:
reworking to include those two external jobs

ekohl:
@komidore64 forgot to push? :)

komidore64:
still in progress

komidore64:
Once this is merged, we'll need to delete `release_tarballs`, `release_verify_tag`, and `release_build_tarball` jobs from jenkins.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+    dir(project) {
+        git url: 'https://github.com/theforeman/foreman', branch: version
+        configureRVM(ruby_ver)
+
+        if (project == 'foreman') {
+            sh "cat config/settings.yaml.example > config/settings.yaml"
+            sh "cat config/database.yml.example > config/database.yml"
+        }
+
+        if (project == 'foreman-installer') {
```
ekohl: Since 1.16 we set this correct in the Gemfile. I don't think we'll do any new 1.15 release do I think we can drop this.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+    deleteDir()
+}
+
+void build_tarball(project, version, ruby_ver) {
+    def base_dir = "/var/www/vhosts/downloads/htdocs/foreman"
+    def rake = "rake"
+
+    deleteDir()
+    dir(project) {
+        git url: 'https://github.com/theforeman/foreman', branch: version
```
ekohl: Branches typically ```VERSION-stable```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+            build job: 'release_packages', parameters: [
+                [$class: 'StringParameterValue', name: 'version', value: version],
+                [$class: 'StringParameterValue', name: 'major_version', value: major_version]
+            ]
+        }
+    }
+}
+
+void verify_tag(project, version) {
+    deleteDir()
```
ehelms: This will cause concurrency issues. Since the parallel steps above are running on a single node and single workspace, this could delete the directory entirely while another parallel process is working on it.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+    dir(project) {
+        git url: "https://github.com/theforeman/${project}", branch: 'develop'
+        sh """
+        if [ \"$(git tag -l ${version} | wc -l)\" -ne 1 ]; then
```
ehelms: I think we can reduce this to simply this line of code with a `sh`. If it returns 1, the `sh` will automatically error out the step which is the behavior I think we want.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+        """
+    }
+    deleteDir()
+}
+
+void build_tarball(project, version, ruby_ver) {
+    def base_dir = "/var/www/vhosts/downloads/htdocs/foreman"
+    def rake = "rake"
+
+    deleteDir()
```
ehelms: This will similarly cause concurrency issues. I think because you scope on `dir(project)` this `deleteDir()` is unneeded.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        withRVM(["${rake} pkg:generate_source", ruby_ver)
+
+        sh "ssh root@theforeman.org \"mkdir -p ${base_dir}/\" || true"
+        sh "rsync -v --ignore-existing pkg/* root@theforeman.org:${base_dir}/"
+
+        cleanupRVM('', ruby_ver)
+    }
+    deleteDir()
```
ehelms: This will similarly cause concurrency issues. I think because you scope on dir(project) this deleteDir() is unneeded.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: I believe `-stable` would be included in the `version` variable. It's not appended to the [current job](ci.theforeman.org/job/release_build_tarball) as seen below:

```sh
#!/bin/bash -xe

git checkout ${version}

if [ $project = foreman ]; then
  cat config/settings.yaml.example > config/settings.yaml
  cat config/database.yml.example > config/database.yml
fi

...
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: Would it be safer to include these **inside** the `dir(project)` block?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
ekohl: But the exact same variable is used to verify the tag exists. That doesn't sound correct.

komidore64: `release_verify_tag` doesn't append `-stable` either.

```sh
#!/bin/bash -xe

if [ "$(git tag -l ${version} | wc -l)" -ne 1 ]; then
  echo "ERROR: cannot find tag ${version}"
  git tag -l
  exit 1
fi

exit 0
```

ekohl: I meant that both use ```$version```. If I understand it correct then verify_tag would check ```git tag -l ${version}``` (which is either 1.17.0 I guess). Then package building results in ```git checkout 1.17.0``` (which works). That does refer to a tag rather than a branch and I'm not sure ```branch: SOME_TAG``` works because I don't know how Jenkins internally works.

komidore64: wow that's surprising. the `git:` plugin actually **doesn't** support referencing by tag ([ref](https://issues.jenkins-ci.org/browse/JENKINS-27018)).

komidore64:
rebased

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
ehelms: @komidore64 This comment got lost in the updated push.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
ehelms: Are you wanting to do the double deleteDir just to be opportunistic? Given the entire workspace is cleaned up at the end of the job these seems like overkill.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+void build_tarball(project, version, ruby_ver) {
+    def base_dir = "/var/www/vhosts/downloads/htdocs/foreman"
+    def rake = "rake"
+
+    dir(project) {
+        deleteDir()
+        checkout scm: [$class: 'GitSCM',
+            userRemoteConfigs: [[url: "git@github.com:theforeman/${project}.git"]],
+            branches: [[name: "refs/tags/${version}"]]], changelog: false, poll: false
```
ehelms: I think you can do this with the basic `git` step -- https://jenkins.io/doc/pipeline/steps/git/

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,105 @@
...
+
+        if (fileExists("Gemfile")) {
+            withRVM(["bundle install --without=development --retry=5"], ruby_ver)
+            rake = "bundle exec rake"
+        }
+
+        withRVM(["${rake} pkg:generate_source", ruby_ver)
+
+        sh "ssh root@theforeman.org \"mkdir -p ${base_dir}/\" || true"
+        sh "rsync -v --ignore-existing pkg/* root@theforeman.org:${base_dir}/"
```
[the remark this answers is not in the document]
ehelms: This might seem like an early optimization, but we have nightly jobs that build tarballs for example to then build nightly RPMs that could take use of this function.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: basic `git:` doesn't support referencing a tag. You can see here from the [jenkins bug report](https://issues.jenkins-ci.org/browse/JENKINS-27018).

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: Just trying to clean up after myself. These aren't needed?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
komidore64: removed

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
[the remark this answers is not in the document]
komidore64: Do you know of any jobs off the top of your head?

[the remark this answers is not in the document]
komidore64: I don't think this suggestion is a bad idea. In fact I think it's a very good idea, but maybe outside of the scope of this PR since it's been open for some time now.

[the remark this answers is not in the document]
ehelms: ACK

ehelms:
@xprazak2 I know you are planning RC2 for Foreman, this re-factors one of the release jobs so I don't want to merge it and break you unless you feel comfortable with it.

komidore64:
squashed

xprazak2:
@ehelm, thanks for letting me know, I would prefer waiting with merge until RC2 is out.

ehelms:
@ekohl @xprazak2 Now that RC2 is out, thoughts on this?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+void build_tarball(project, version, ruby_ver) {
+    def base_dir = "/var/www/vhosts/downloads/htdocs/foreman"
+    def rake = "rake"
+
+    dir(project) {
+        checkout scm: [$class: 'GitSCM',
+            userRemoteConfigs: [[url: "git@github.com:theforeman/${project}.git"]],
```
ekohl: Would it be better to use ```https://```? Not sure if we should rely on SSH keys if there's no need for it.

komidore64: updated

PR #533, opened 2018-03-08
Do no not run vagrant up with --debug

ekohl:
The debug info contains API auth tokens so this is a major security
issue.


mmoll:
merged, bedankt @ekohl!

PR #534, opened 2018-03-08
deploy_puppet - setup pollSCM and move logrotate into groovy

komidore64:


ehelms:
Just asking since you are making changes here, should we update this to the naming scheme? e.g. puppet-deploy ?

komidore64:
Ah yeah, that's probably a good call.

komidore64:
@ehelms updated

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/puppet_deploy.yaml
```diff
@@ -0,0 +1,13 @@
+- job:
+    name: puppet_deploy
```
ehelms: Little thing, `-` instead of `_`

komidore64: updated!

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/puppetDeploy.groovy
```diff
@@ -2,9 +2,14 @@ pipeline {
...
-        timestamps()
-        disableConcurrentBuilds()
         ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '3'))
+        disableConcurrentBuilds()
+        timestamps()
+    }
+
+    triggers {
+        pollSCM('H/5 * * * *')
```
akofink: This polls the foreman-infra repo every 10 minutes?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/puppet-deploy.yaml
```diff
@@ -0,0 +1,13 @@
+- job:
+    name: puppet-deploy
+    project-type: workflow
+    properties:
+      - github:
+          url: https://github.com/theforeman/foreman-infra
+    triggers:
+      - github
```
akofink: But is also triggered by git hooks from GitHub? Why do we need both?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/puppetDeploy.groovy
ekohl: every 5 I'd think

ekohl:
When deploying we should be careful to remove the old job as well.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/puppet-deploy.yaml
ehelms: See https://wiki.jenkins.io/display/JENKINS/GitHub+Plugin -- the two are inter-related

ehelms:
Thanks @komidore64 . Merging, I will delete old job once new job appears.

PR #535, opened 2018-03-09
Revert "Use sha1 for pull request builder"

akofink:
I don't think this change helped us. In fact, I'm starting to think we should definitely revert it asap.

This reverts commit 4fe4e86f06093331521e97bc1765ed2806f92dc6.


akofink:
Actually, after reading the Jenkins docs, it seems like sha1 is actually what we want:

> ${ghprbActualCommit} if you want to use the head of the pull request branch (e.g. refs/pull/4/head); or
> ${sha1}, to use GitHub's tentative merge of the compare and base branches (e.g. refs/pull/4/merge) if the PR can be automatically merged or the head of the pull request branch (e.g. refs/pull/4/head) if they can not be automatically merged.

PR #537, opened 2018-03-12
use utf8mb4 for MySQL in tests

mmoll:


mmoll:
I'm unsure what consequences this has for non-develop tests

ares:
You mean plugins e.g.? I'd merge and see what happens, we only had to fix two too long indices in core. We can always revert if it is too bad.

mmoll:
> You mean plugins

Nope, but I'd expect all versions that don't have the `:limit => 40` migrations to fail in `db:migrate`

ares:
oh, so we'd better to get the one fixing it in before? do you want me to send separate PR just with the limit?

mmoll:
> do you want me to send separate PR just with the limit

If there are no better options, yes. And once that got cherry-picked down to 1.15, this PR can get in.

ekohl:
Is this something we should set in the mysql connection so users don't forget to set it?

ares:
we could do it, not sure what the outcome for existing users would be though

PR #538, opened 2018-03-13
puppet-deploy - fix github polling once and for all?

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/puppet-deploy.yaml
```diff
@@ -6,6 +6,8 @@
           url: https://github.com/theforeman/foreman-infra
     triggers:
       - github
+      - pollscm
```
ehelms: I think this needs a `:` to be valid yaml

komidore64: oops good catch!

komidore64:
@ehelms updated

PR #542, opened 2018-03-14
Fix systest bats reporting

ekohl:
Forklift outputs test.tap files. This change ensures Jenkins can find
the correct results.


ekohl:
https://ci.theforeman.org/job/systest_foreman/14378 & https://ci.theforeman.org/job/systest_foreman/14379 are currently running with this change applied manually. Going to let them finish before merging this.

ekohl:
Updated. Applied locally and now it should work.

PR #547, opened 2018-03-15
Fix SCM polling in puppet-deploy

ekohl:
Without this Jenkins doesn't know what to poll.


komidore64:
ACK

mmoll:
merged, bedankt @ekohl!

ekohl:
I swear it worked once but now it no longer doesn't?

PR #549, opened 2018-03-15
convert `packaging_discovery_image` to JJB `foreman-discovery-image-publish`

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/scm/foreman-discovery-image.yml
```diff
@@ -0,0 +1,9 @@
+- scm:
+  name: foreman-discovery-image
```
ehelms: I think it might complain about this needing to be indented 2 more spaces.

komidore64: Oh okay, I thought since it was indented from the `-`, it'd parse properly. I can update.

komidore64: I was incorrect! I tried it on http://yaml-online-parser.appspot.com/. they parsed differently.

komidore64:
updated indentation!

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-discovery-image-publish.yaml
```diff
@@ -0,0 +1,30 @@
...
+    logrotate:
+      daysToKeep: 90
+      numToKeep: 30
+    scm:
+      - foreman-discovery-image
+    builders:
+      - foreman-discovery-image-publish
+    parameters:
+      - string:
+        name: proxy_repository
```
ehelms: I think this, for example, might need to be indented one level

komidore64: bah, I AM GOOD AT PROGRAMMING

komidore64:
updated indentation again

ehelms:
What normally triggers the job this is replacing?

komidore64:
there are no build triggers.

ehelms:
@lzap Please be aware that we have created this job, pushing it into source control as a replacement for http://ci.theforeman.org/job/packaging_discovery_image/ -- we will plan to delete that job in the next few days.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/foreman-discovery-image-publish.sh
```diff
@@ -0,0 +1,37 @@
...
+# monkey patch vagrant-rackspace - bug https://github.com/mitchellh/vagrant-rackspace/pull/153
+sed -i 's/retryable(:on => Fog::Errors::TimeoutError, :tries => 200) do/retryable(:on => [Excon::Error::Timeout, Fog::Errors::TimeoutError], :tries => 200) do/' /home/jenkins/.vagrant.d/gems/gems/vagrant-rackspace-0*/lib/vagrant-rackspace/action/create_server.rb
+
+base_dir=/var/www/vhosts/downloads/htdocs/discovery
+
+cd aux/vagrant-build
+
+export distro=f27
+export proxy_repo=$(eval echo ${proxy_repository})
+export VAGRANT_DEFAULT_PROVIDER=rackspace
```
ekohl: I wasn't aware that this existed. This provider no longer exists on slaves and the monkey patch above also no longer works.

ekohl:
This does explain the 2 Fedora images that are on rackspace running since January. It doesn't destroy an image when it's done.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/foreman-discovery-image-publish.sh
komidore64: Can they safely be removed without consequence?

lzap:
Thanks for letting me know @ehelms ! Yeah @ekohl we build FDI via Vagrant/Rackspace as we don't have ISO building setup on our Koji. There is no space either ;-)

lzap:
@ekohl what exactly explains there were some leftovers? I did not know about them obviously.

ekohl:
@lzap for the past 2 months we've had 2 Fedora machines running at Rackspace. I didn't know where they came from so didn't remove them.

Other than that I'm very unhappy with monkey patching the installed plugin. We manage our slaves and this should not happen in jobs.

Because we changed to the openstack provider this job will currently be broken. When rewriting it, it'll be very important to remove the ```--debug``` in the script since the openstack provider will log auth tokens.

komidore64:
> Okay. Having this merged, where do I trigger the job? I am triggering it manually very often to build scratchbuilds of PRs to test them.

http://ci.theforeman.org/job/foreman-discovery-image-publish/

PR #554, opened 2018-03-19
convert release_build_katello_installer to JJB workflow

komidore64:
renamed to `tarballs-katello-installer-release`


ekohl:
What the Foreman jobs do is trigger an RPM build so you update nightly automatically. Should that be done in this job as well? How do we currently get nightly katello RPMs?

We should also should look at setting this right in [package_manifest.yaml](https://github.com/theforeman/foreman-packaging/blob/ae0d2ae124301426cefd3df728d88db9db90f438/package_manifest.yaml#L273) (even if we don't do it right now).

komidore64:
katello-installer nightlies are built based on a timer

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tarballs-katello-installer-release.yaml
```diff
@@ -0,0 +1,11 @@
+- job:
+    name: tarballs-katello-installer-release
```
ehelms: I think based on naming scheme we'd want:

katello-installer-tarballs-release

ehelms:
Should this be triggering the packaging RPM build job for the installer at the end? I thought upstream did that

ekohl:
> Should this be triggering the packaging RPM build job for the installer at the end? I thought upstream did that

I'd be nice if we could use the generic RPM building job rather than a separate build_rpm_katello-installer_nightly similar to how the foreman nightlies are built. If we're changing jobs, it'd be good to decide on a standard now and to align the jobs we're rewriting anyway.

komidore64:
`release_build_katello_installer` only archives the tarball.

it looks like `build_rpm_katello_installer` (which is also timer based) kicks off a `packaging_build_rpm` job to brew `katello-installer` on koji, but i don't see it referencing any kind of stored artifacts. puzzling.

ekohl:
> it looks like build_rpm_katello_installer (which is also timer based) kicks off a packaging_build_rpm job to brew katello-installer on koji, but i don't see it referencing any kind of stored artifacts. puzzling.

I think it uses the `koji-katello-jenkins` releaser which takes the last successful job. The downside is that if `release_build_katello_installer` fails, `build_rpm_katello_installer` looks like it's succeeding but with an outdated tarball. Then the nightlies get an increased timestamp with old content. The model of the tarbal building job triggering the packaging avoid this problem.

It might still have the problem that tarbal building succeeds while packaging building fails. If that's one shared job, it could be easy to miss. If the tarbal job depends on the packaging, I think this is covered.

komidore64:
it would make more sense to me also to have the tarball job trigger the nightly build job upon successful completion

ekohl:
> it would make more sense to me also to have the tarball job trigger the nightly build job upon successful completion

You're right. Calling it as a subjob fails because the tito releaser retrieves the last successful job. If it hadn't completed, it wouldn't be used.

ehelms:
I swear http://ci.theforeman.org/job/build_rpm_katello-installer_nightly/ used ot be triggered upon completion of this job and not timer based. That is what we should strive for as y'all discussed.

Source job -> triggers packaging job

Or in a pipeline job re-factor, wrap both into the same job so its simply a nightly release job that does it all.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        stage('Trigger RPM Build') {
+            steps {
+                build job: 'packaging_build_rpm', propagate: true, parameters: [
+                    string(name: 'project', value: 'packages/katello/katello-installer'),
+                    booleanParam(name: 'gitrelease', value: false),
+                    booleanParam(name: 'scratch', value: false),
+                    string(name: 'releaser', value: 'koji-katello-jenkins'),
+                    string(name: 'nightly_jenkins_job', value: 'release_build_katello_installer'),
```
ehelms: This line would need to now reference this job right given this job generates the archived tarball?

komidore64: hah! that sure would make sense wouldn't it!

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
```diff
@@ -0,0 +1,45 @@
...
+                    string(name: 'nightly_jenkins_job_id', value: 'lastSuccessfulBuild')
```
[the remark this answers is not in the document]
komidore64: Nope. According to `build_rpm_katello-installer_nightly`, it's just a string, not a variable.

komidore64:
note: once this is merged, we'll need to remove `build_rpm_katello-installer_nightly` because its functionality is being rolled into `katello-installer-release`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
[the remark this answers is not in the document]
ehelms: Correct the current job does this today because there are two separate jobs. If you self-reference the `lastSuccessfulBuild` it will go find the previous build given this current job is in progress already.

For example, say the last job was #4. This kicks off as #5, it creates an artifact, then it kicks off packaging_build_rpm and waits for it. Job #5 is in progress, so `lastSuccessfulBuild` at that time references #4 giving the wrong artifact.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
komidore64: updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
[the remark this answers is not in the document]
komidore64: updated

ekohl:
The title doesn't really match the content. The current job is already in JJB, just not a pipeline.

ehelms:
Generally this looks good code wise -- have you created and tested it yet?

komidore64:
> The title doesn't really match the content. The current job is already in JJB, just not a pipeline.

Which job are you referring to @ekohl? There are multiple jobs referenced in this PR.

ekohl:
@komidore64 ah, right, it replaces one JJB job and one manual job (not in JJB) with a JJB pipeline.

komidore64:
```
Started by user Adam Price
Running in Durability level: MAX_SURVIVABILITY
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 16: Expected a step @ line 16, column 21.
                       git url: 'https://github.com/katello/katello-installer', branch 'master'
                       ^

1 error

	at org.codehaus.groovy.control.ErrorCollector.failIfErrors(ErrorCollector.java:310)
	at org.codehaus.groovy.control.CompilationUnit.applyToPrimaryClassNodes(CompilationUnit.java:1085)
	at org.codehaus.groovy.control.CompilationUnit.doPhaseOperation(CompilationUnit.java:603)
	at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:581)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:558)
	at groovy.lang.GroovyClassLoader.doParseClass(GroovyClassLoader.java:298)
	at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:268)
	at groovy.lang.GroovyShell.parseClass(GroovyShell.java:688)
	at groovy.lang.GroovyShell.parse(GroovyShell.java:700)
	at org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.doParse(CpsGroovyShell.java:133)
	at org.jenkinsci.plugins.workflow.cps.CpsGroovyShell.reparse(CpsGroovyShell.java:127)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.parseScript(CpsFlowExecution.java:557)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.start(CpsFlowExecution.java:518)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:290)
	at hudson.model.ResourceController.execute(ResourceController.java:97)
	at hudson.model.Executor.run(Executor.java:429)
Finished: FAILURE
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloInstallerRelease.groovy
```diff
@@ -0,0 +1,45 @@
...
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Setup Environment') {
+            steps {
+                dir('katello-installer') {
+                    configureRVM()
+                    git url: 'https://github.com/katello/katello-installer', branch 'master'
```
zjhuntin: `branch: 'master'`

komidore64: argh, thank you.

komidore64:
@ekohl granted, i can see the argument for naming this job `katello-installer-nightly-release` because that's what this is doing.

komidore64:
[successful build](http://ci.theforeman.org/job/katello-installer-release/9/)

PR #562, opened 2018-03-22
dry up the packaging pipelines

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51.groovy
```diff
@@ -23,8 +23,7 @@ pipeline {
             steps {
 
                 script {
-                    changed_packages = sh(returnStdout: true, script: "git diff origin/master --name-only -- 'packages/*.spec' | cut -d'/' -f2 |sort -u").trim()
-                    packages_to_build = changed_packages.split().join(' ')
+                    packages_to_build = find_changed_packages("origin/master")
```
ekohl: Doesn't ghprb provide a variable for the target branch?

evgeni: it does. but we did not use it in tfm-ror51 as it only has one branch. but let's still fix it

ehelms:
@evgeni Needs a rebase now

PR #567, opened 2018-03-27
Limit vagrant jobs to hosts with ipv6

ekohl:
Because we provision on Rackspace (which has IPv6) and the vagrant-openstack-plugin can't decide which IP it should use, we pin to nodes with IPv6.


komidore64:
looks sane to me

PR #569, opened 2018-03-29
refactor systest family of jobs

komidore64:
`systest_foreman` was triggered as a subjob for a handful [1] of other jenkins jobs. It made more sense to share a JJB builder to lessen the amount of job-chains that exist on ci.tfm.org.

1. The systest shell blob has been refactored to use builder macros. It lives in `ci/yaml/builders/systest.yaml` now.

1. `systest_*` jobs have been refactored to use the new `systest` builder and have been renamed to `*_systest` (`<entity>-<action>` style).

1. `release_nightly_test_deb` and `release_test` now use the `systest` builder rather than triggering a `systest_foreman` subjob.

1. `systest_foreman_hammer_nightly` has been converted to JJB.

1. `systest_foreman` has been removed.

### remove from jenkins

- [ ] [systest_foreman_hammer_nightly](http://ci.theforeman.org/job/systest_foreman_hammer_nightly)
- [ ] [systest_foreman_mysql](http://ci.theforeman.org/job/systest_foreman_mysql)
- [ ] [systest_foreman_puppet_nightly](http://ci.theforeman.org/job/systest_foreman_puppet_nightly)
- [ ] [systest_foreman_puppet_pc1](http://ci.theforeman.org/job/systest_foreman_puppet_pc1)
- [ ] [systest_foreman_umask_077](http://ci.theforeman.org/job/systest_foreman_umask_077)
- [ ] [systest_foreman](http://ci.theforeman.org/job/systest_foreman)

---

[1]:
- release_nightly_test_deb
- release_test
- systest_foreman_hammer_nightly
- systest_foreman_mysql
- systest_foreman_puppet_nightly
- systest_foreman_puppet_pc1
- systest_foreman_umask_077


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - axis:
+          type: user-defined
+          name: os
+          values:
+            - centos7
+    builders:
+      - systest:
+          repo: nightly
+          repo_environment: release
+          os: ''
```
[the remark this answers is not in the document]
komidore64: Unfortunately with how builder macros work, yes. Though I've noticed that `os` comes from the matrix in every one of these systest jobs, so it could simply be removed completely.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/systest.shell_template
```diff
@@ -0,0 +1,77 @@
...
+
+if [[ ${{run_hammer_tests}} == true ]]; then
+  args+=" --foreman-organizations-enabled=true --foreman-locations-enabled=true"
+fi
+
+if [[ -n "${{db_type}}" ]]; then
+  args+=" --foreman-db-type=${{db_type}}"
+fi
+
+cat > "${{filename}}" <<-EOF
```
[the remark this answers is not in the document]
komidore64: This is probably about the best thing I'd be able to come up with using bash.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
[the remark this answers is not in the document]
komidore64: `os` and `{os}` removed

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-puppet-pc1-systest.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - forklift
+    triggers:
+      - timed: 'H 10 * * 6'
+    axes:
+      - axis:
+          type: user-defined
+          name: os
+          values:
+            - centos7
+            # - debian9     # not available yet
```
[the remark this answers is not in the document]
komidore64: i removed these lines

komidore64:
i will squash this once it's ACK'd.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/systest.shell_template
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash -xe
```
ehelms: What makes this no longer runnable as a stand-alone shell?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-mysql-systest.yaml
```diff
@@ -0,0 +1,33 @@
+- job:
+    name: foreman-mysql-systest
```
ehelms: Is this nightly as well or ran against releases as well?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-puppet-pc1-systest.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - axis:
+          type: user-defined
+          name: os
+          values:
+            - centos7
+    builders:
+      - systest:
+          repo: nightly
+          repo_environment: release
+          pl_puppet: 4
```
ehelms: @ekohl Are we all Puppet 5 at this point? Should we update this to 5 or drop this job all together?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-umask-077-systest.yaml
```diff
@@ -0,0 +1,33 @@
+- job:
+    name: foreman-umask-077
```
ehelms: Should this end with `nightly-systest` ? @ekohl Is this a test that should be run stand-alone or can it be combined?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-puppet-pc1-systest.yaml
ekohl: I think we still support Puppet 4 for now. We could make an official statement for 1.18 that we're going to stop testing it which would make it unsupported.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-umask-077-systest.yaml
ekohl: Not sure what you would combine it with. It can't be combined with regular installs since they have a different umask. We should actually extend it to katello so we know that works. There have been bug reports about it IIRC.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
```diff
@@ -1,10 +1,10 @@
 # Testing already-built RPMs from Koji and Debian packages in stagingdeb.
 - job:
     name: release_test
+    node: 'el && ipv6'
     project-type: matrix
     logrotate:
-      daysToKeep: -1
-      numToKeep: 16
+      daysToKeep: 8
```
ekohl: Why change this to days? This is used in the release pipeline which you run at most every few weeks. A number feels logical to me.

zjhuntin: I agree, looking over the jobs we seem to have a number of ways we handle this over these jobs. I think keeping a number is better overall, as it will let anyone who is unfamiliar with the job see what a run should look like. Also making this uniform across all systest jobs would be awesome.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/systest.shell_template
komidore64: It is no longer runnable as a standalone shell script because all shell variables have to be escaped with an extra set of `{}` to avoid being mistaken for builder macros.

if bash doesn't have a problem with `${{os}}` vs `${os}` then I simply made an assumption that was incorrect.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-umask-077-systest.yaml
komidore64: Oops, yes. this does need `-systest` on the end.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
komidore64: I agree that we should have a unified `days-` and `numToKeep` for all jobs in our jenkins instance, but I believe that is outside the scope of this PR.

komidore64: I set this job's `daysToKeep` and `numToKeep` to be the same as what `systest_foreman` was previously because it isn't a job-chain anymore, but rather a build step of the job in question.

komidore64:
@ekohl, i incorporated your `systest_foreman.sh` conflicting change.

komidore64:
@ehelms @ekohl hey can i get some traction here?

ekohl:
@komidore64 I'd like to finish various packaging changes before we release 1.17 and branch 1.18 before I dive into this.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
```diff
@@ -0,0 +1,35 @@
...
+    axes:
+      - axis:
+          type: user-defined
+          name: os
+          values:
+            - centos7
+    builders:
+      - systest:
+          repo: nightly
+          repo_environment: release
```
ekohl: Shouldn't this be staging

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
```diff
@@ -0,0 +1,35 @@
...
+          pl_puppet: ''
```
ekohl: Are empty parameters needed?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-mysql-systest.yaml
ekohl: I think @ehelms is asking if `repo_environment` should be `staging`

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-puppet-nightly-systest.yaml
```diff
@@ -0,0 +1,33 @@
+- job:
```
ekohl: We know this job is currently broken because most (all?) modules don't allow puppet version 6 in their metadata. We should either pass in the parameter to skip the puppet version check to actually test it or disable the job. Right now I'm leaning to the latter since we don't really spend resources on it and I don't think anyone is actually checking this job.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_nightly_test_deb.yaml
```diff
@@ -1,28 +1,38 @@
...
       - axis:
           type: user-defined
           name: os
           values:
           - debian9
           - ubuntu1604
+    builders:
+      - systest:
+          repo: nightly
+          repo_environment: release
```
ekohl: I think this should be staging.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/systest.shell_template
ekohl: It is weird that `!include-raw` still applies a template over it. The docs suggest `!include-raw-escape` would escape it for you while still being a plain shell script in the repo. Could you try that?

komidore64:
rebased

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
ehelms: Agreed.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-mysql-systest.yaml
ehelms: Also asking if the name should include `nightly`

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_nightly_test_deb.yaml
komidore64: @ekohl `release` is correct. release_nightly_test_deb currently calls systest_foreman with a `staging=true` variable, but that parameter is not even defined on systest_foreman.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-puppet-nightly-systest.yaml
komidore64: @ekohl i will disable this job.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_nightly_test_deb.yaml
ekohl: I talked to @mmoll and I'm pretty sure that's actually a regression I introduced when refactoring. I still needed to find exactly where, but it was causing issues in the Debian pipeline.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-hammer-nightly-systest.yaml
komidore64: @ekohl @ehelms this one is also currently set to `release`.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_nightly_test_deb.yaml
komidore64: I'll do just this one because we're friends

komidore64:
@ekohl ping

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/builders/systest.yaml
```diff
@@ -0,0 +1,80 @@
...
+          #!/bin/bash -xe
+
+          # populate variables from builder macros if they're given
+          [ -n "{repo}" ] && repo="{repo}"
+          [ -n "{repo_environment}" ] && repo_environment="{repo_environment}"
+          [ -n "{pl_puppet}" ] && pl_puppet="{pl_puppet}"
+          [ -n "{run_hammer_tests}" ] && run_hammer_tests=$(echo "{run_hammer_tests}" | tr '[:upper:]' '[:lower:]')
+          [ -n "{db_type}" ] && db_type="{db_type}"
+          [ -n "{umask}" ] && umask="{umask}"
+          [ -n "{expected_version}" ] && expected_version=="{expected_version}"
```
ekohl: I think this should be `=` rather than `==`

komidore64: you are right!

komidore64: @ekohl fixed it!

komidore64:
@ehelms, @zjhuntin ?

PR #570, opened 2018-04-02
clean gemset after Katello tests

mmoll:


mmoll:
I'm unsure about the syntax of cleanupRVM, so no idea if this works :)

ekohl:
The cleanup function should take care of this: https://github.com/theforeman/foreman-infra/blob/81c0764ff12ef57f10d1d5da48d07246c77e9caa/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy#L53-L63

Are you seeing a specific error that needs this?

mmoll:
I'm seeing leftovers, e.g.:
```
mmoll@foreman-build:/usr/local/rvm/gems$ ls -1 |grep katello-pr-test
ruby-2.4.3@katello-pr-test-1033
ruby-2.4.3@katello-pr-test-1099
ruby-2.4.3@katello-pr-test-1111
ruby-2.4.3@katello-pr-test-1114
ruby-2.4.3@katello-pr-test-1129
ruby-2.4.3@katello-pr-test-1149
ruby-2.4.3@katello-pr-test-918
ruby-2.4.3@katello-pr-test-925
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -118,6 +118,11 @@ pipeline {
 
             }
         }
+        stage('Cleanup RVM') {
+            steps {
+                cleanupRVM(ruby)
```
ehelms: The post section below already calls `cleanup` which in turn calls cleanupRVM. So if gemsets are not being cleaned up I'd think its due to a different problem. Can you point to a failure to do this?

mmoll:
thx, closing :+1:

PR #571, opened 2018-04-02
Add test_bastion_assets_precompile job to source control

zjhuntin:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_bastion_assets_precompile.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - matches:
+          - log-text:
+          script: |
+            #!/bin/bash
+            # Clean gemset and database
+            . /etc/profile.d/rvm.sh
+            gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
+            rvm use ruby-${ruby}@${gemset}
+            rvm gemset delete ${gemset} --force
+            exit 0
```
ehelms: I think there is an existing publisher for doing this for us.

zjhuntin: gotcha, I'll find that

komidore64:
this probably should live with the other bastion jobs in `yaml/jobs/plugins/bastion.yaml`.

PR #572, opened 2018-04-03
Add test_bastion_pull_request job to source control

zjhuntin:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_bastion_pull_request.yaml
```diff
@@ -0,0 +1,69 @@
...
+            - trigger-builds:
+              - project: test_bastion_assets_precompile
+                current-parameters: true
+                block: true
+    axes:
+      - axis:
+          type: user-defined
+          name: test
+          values:
+            - javascript assets_precompile
```
komidore64: i think this is clearer as two lines?

```
values:
  - javascript
  - assets_precompile
```

zjhuntin:
@komidore64 done

PR #573, opened 2018-04-03
Always deleteDir and cleanup on Katello PR test

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -123,8 +131,6 @@ pipeline {
     post {
         always {
             dir('foreman') {
-                archiveArtifacts artifacts: "Gemfile.lock, log/test.log"
-                junit keepLongStdio: true, testResults: 'jenkins/reports/unit/*.xml'
                 cleanup(ruby)
```
ekohl: Does this still need to happen in the foreman dir?

PR #576, opened 2018-04-04
bastion consolidation and conversion to GHPRB

komidore64:
it looks like `test_bastion_assets_precompile` still had a `2` on the end, so i removed that.


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/bastion.yaml
```diff
@@ -39,7 +50,8 @@
           url: https://github.com/katello/bastion
     scm:
       - git:
-          url: https://github.com/katello/bastion
+          url: git://github.com/katello/bastion.git
```
ehelms: Could this re-use the SCM macro defined at the top of the file?

komidore64: it uses `master` branch instead of `${branch}`. so i don't think so?

komidore64: instead i just added a `branch` parameter that defaults to `master` so we can use the same scm.

komidore64:
@ehelms is this all that's necessary for using GHPRB? Or do I need to make changes elsewhere?

ehelms:
@komidore64 I need to make a change to the github repository but from a job perspective this should do it

PR #579, opened 2018-04-06
Install npm for plugins in test job

dLobatog:
Currently the plugins with webpack assets, are not getting their
dependencies fulfilled and webpack always fails to compile in the
test_plugin_pull_request job. The reason is that `npm install` is ran
**before** putting the plugin in the gemfile.

This change runs `npm install` a second time after the Gemfile has been
updated to ensure we have the npm dependencies ready for webpack.

https://github.com/theforeman/foreman_remote_execution/pull/130 for example benefited from this change (I changed it manually on Jenkins)


ekohl:
Thanks!

PR #580, opened 2018-04-09
test-hammer-cli-foreman-discovery - convert to JJB

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli_foreman_discovery.yaml
```diff
@@ -0,0 +1,32 @@
...
+          type: user-defined
+          name: ruby
+          values:
+            - 2.3
+            - 2.4
+            - 2.5
+    builders:
+      - shell: !include-raw: scripts/test/test_hammer_cli_foreman_discovery.sh
+    publishers:
+      - gemset_cleanup
```
ehelms: Should this also have an IRC notification publisher?

komidore64: oops. missed that! yes it should.

komidore64: fixed

komidore64:
added `test_hammer_cli_foreman_discovery_pull_request` because it also references `scripts/test/test_hammer_cli_foreman_discovery.sh`.

PR #583, opened 2018-04-10
Allow free form deb source branch names

ekohl:


ekohl:
Note I didn't test this, but I think it's enough. In the past we had this to determine the build type, but we already use the target branch for that.

The only downside now is that we can't rerun builds from before this change after it's deployed, but I think that's an acceptable downside.

mmoll:
let's try it...

mmoll:
merged, bedankt @ekohl!

PR #584, opened 2018-04-10
test_brakeman - convert to JJB

komidore64:
we need to get the JJB-brakeman plugin [1] added to this repo. i'm not sure how to do that. otherwise `update_jobs.sh` will complain about not understanding the `brakeman` entry point in publishers.

[1] https://github.com/garethr/jenkins-job-builder-brakeman


komidore64:
instead we'll just plan to delete this job on 18 april 2018.

PR #586, opened 2018-04-12
Drop db:seed run from develop tests

tbrisker:
we already have tests for the seeds in core, no need to rerun the seeds. This was added mostly to make sure plugin seeds run properly.


mmoll:
merged, thanks @dLobatog!

PR #589, opened 2018-04-12
Delete directory before katello tests

jlsherrill:


jlsherrill:
this is an attempt to fix the gem locked issues we're seeing randomly on PR tests

mmoll:
let's try this, thanks @jlsherrill!

PR #590, opened 2018-04-12
test_katello_rubocop - convert to JJB

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_katello_rubocop.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash -ex
+
+ruby=2.2.3
```
mmoll: 2.4.3 would be better here, I think.

komidore64: Is that what most dev setups use by default?

mmoll: Then I'd say it's time to update at least to 2.3. ;)

2.2 is EOL upstream-wise and RPMs are using 2.4. In Foreman, we also stopped testing with 2.2 at the beginning of this year.

PR #594, opened 2018-04-17
test_katello_assets_precompile - convert to JJB

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_katello_assets_precompile.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash -ex
+
+plugin_name=katello
+
+[ -e foreman ] && rm -rf foreman/
+git clone https://github.com/theforeman/foreman --branch "${base_foreman_branch}"
+
+ruby=2.3
```
mmoll: why not 2.4?

komidore64: Mainly because I just copied the script from the non-JJB'd job :)

I'm not opposed to changing it though.

mmoll: Ah, I see... I just remember I did update the ruby version in some katello job because bundle install failed on a gem that required 2.3+

mmoll:
let's bring this in :)

PR #595, opened 2018-04-18
do not require downstream stability

timogoebel:
This should disable the `Waiting for stable build of` message, see https://github.com/openshift/test-pull-requests/blob/699cc3085e3a1cd090d9c185781bf676ab183191/test_pull_requests#L2253 for reference.

cc: @lzap, @ehelms


lzap:
This is a GODSEND, if folks are not against. Crossing my fingers. Lately, discovery tests were all but green.

mmoll:
I'm for it, merged, danke @timogoebel!

timogoebel:
@mmoll: This unfortunately does not seem to work. Do you mind checking this out? Do we have the latest version of this script?

iNecas:
Aren't we using https://github.com/theforeman/test-pull-requests, which is fork of the openshift's one, that doesn't have this option?

timogoebel:
Ah, that makes sense. I wasn't able to figure out what version of `test-pull-requests` we use and how it's deployed.

PR #596, opened 2018-04-18
Do not use --force for katello grunt ci

jlsherrill:


jlsherrill:
@ehelms when using --force, the return code is always zero.  I'm not sure what the purpose of --force is though.  Does it make sense to get rid of it?

PR #598, opened 2018-04-18
delete root eslintrc to work around eslint bug

jlsherrill:


mmoll:
this is looking funky, but if it fixes the problem let's go for it :birthday:

akofink:
Ah! I think we might have wanted to `rm -f` it just in case it doesn't exist. i.e. http://ci.theforeman.org/blue/organizations/jenkins/katello-pr-test/detail/katello-pr-test/1458/pipeline/63

PR #599, opened 2018-04-18
npm install in foreman for asset compile

jlsherrill:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -101,6 +101,7 @@ pipeline {
                     steps {
                         sh "npm install"
```
mmoll: this is still needed, because bastion, correct?

jlsherrill: I'm not 100% sure it is, but i think it'd is because of the packages.json in the katello dir related to the react stuff there

mmoll:
merged, thanks @jlsherrill!

PR #607, opened 2018-04-23
_javascript couldn't get the PR ref it was handed from _pull_request

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/bastion.yaml
```diff
@@ -8,6 +8,7 @@
           basedir: 'plugin'
           branches:
             - '${branch}'
+          refspec: '+refs/heads/*:refs/remotes/origin/* +refs/pull/*:refs/remotes/origin/pr/*'
```
ekohl: In other ghprb jobs we use `refspec: '+refs/pull/${ghprbPullId}/*:refs/remotes/upstream/pr/${ghprbPullId}/*'` to only pull in the PR we're testing.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/bastion.yaml
```diff
@@ -8,6 +8,7 @@
           basedir: 'plugin'
           branches:
             - '${branch}'
```
ekohl: In foreman-rpm-packaging-pr we also changed this to `${ghprbActualCommit}`

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/bastion.yaml
komidore64: `test_foreman_javascript` isn't a GHPRB job itself, but it can sometimes get kicked by `test_bastion_pull_request` which is and it needs to know how to get at a PR. Here I am telling git about the standard refspec (`+refs/heads/*:refs/remotes/origin/*`) as well as the refspec for grabbing pull-requests (`+refs/pull/*:refs/remotes/origin/pr/*`).

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/bastion.yaml
komidore64: `test_foreman_javascript` isn't a GHPRB job itself, but it can sometimes get kicked by `test_bastion_pull_request`.

ekohl:
I would prefer to rebase it on https://github.com/theforeman/foreman-infra/pull/609 and see if it actually passes

komidore64:
I made the changes manually on jenkins yesterday and `test_bastion_javascript` ran properly after being kicked from a bastion PR.

PR #613, opened 2018-04-30
Typo rubygems should be ruby for bastion job

ehelms:


mmoll:
That Ruby version could also get updated to 2.4 then? :)

PR #615, opened 2018-04-30
test_katello - convert to JJB

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_katello.yaml
```diff
@@ -0,0 +1,78 @@
+- job:
+    name: test_katello_temp
+    description: |
+      Run Katello tests in the following categories:<ul>
+      <li><b>core</b> - ruby unit/functional tests</li>
+      <li><b>javascript</b> - Bastion UI JavaScript tests</li>
+      <li><b>rubocop</b> - Rubocop code analyser</li>
+      <li><b>rubygem_rpm</b> - Katello gem RPM nightly build</li>
```
ekohl: I don't see this in the `test` axis, but it does have `assets`. Am I missing something?

komidore64: It's not in the axis, but it is a publisher. That's the best I've got!

komidore64: Yup! You're right! This variable doesn't get populated from **anywhere**.

PR #621, opened 2018-05-07
fix "Mysql2::Error: SELECT command denied to user [...]" error 

komidore64:


komidore64:
@timogoebel

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-plugins-pull-request.yaml
```diff
@@ -35,6 +34,11 @@
             - postgresql
             - sqlite3
             - mysql
+      - axis:
+          type: slave
+          name: label
+          values:
+            - fast
```
mmoll: :heartbeat:

mmoll:
merged, thanks @komidore64!

PR #623, opened 2018-05-09
fix labels for systest family jobs

komidore64:
systest jobs were complaining about an unreachable host and bombing out.

i think i've figured out that `node: 'el && ipv6'` only works for non-matrix jobs.


mmoll:
yes, this is a known bug of the vagrant openstack module we're using.

mmoll:
merged, thanks @komidore64!

komidore64:
@zjhuntin the systest jobs run some vagrant stuff. The problem that was happening was if a systest was running on a ipv4 box only, vagrant would bomb out because it was unable to find a machine it was looking for that had an ipv6 address. I had to fix it because the pre-JJB systest jobs had their nodes set properly. I messed it up when I converted to JJB :)

komidore64:
uh oh

komidore64:
Ah, I should've used `label-expression`. @mmoll fix incoming!

komidore64:
https://github.com/theforeman/foreman-infra/pull/626

PR #624, opened 2018-05-10
Add events redirect for calendar and ICS file

GregSutcliffe:
As per https://community.theforeman.org/t/moving-event-data-to-discourse/8971/3 this redirects the calendar and ICS file to the equivalents on Discourse.


GregSutcliffe:
@ekohl fair enough, whats the correct syntax then? just drop the 301?

ekohl:
Yes. I'd also put it on line 6 so it happens before the redirect to https. That way http gets one redirect rather than two.

GregSutcliffe:
@ekohl thanks, updated

ekohl:
Thanks!

PR #626, opened 2018-05-11
set 'and' on labels for systest family of jobs

komidore64:
okay, this is properly what we want for the systest jobs!


mmoll:
let's try this :)

mmoll:
merged, thanks @komidore64!

PR #629, opened 2018-05-15
test katello 3.7 against foreman 1.18

evgeni:


evgeni:
@jturel this should fix the issues you had with CPs for 3.7

ekohl:
This should be a step in our branching guide on theforeman.org.

PR #632, opened 2018-05-18
Escape $ in tarballsRelease.groovy

xprazak2:
Should fix [tarballs-release](http://ci.theforeman.org/blue/organizations/jenkins/tarballs-release/detail/tarballs-release/1/pipeline)

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
@@ -57,7 +57,7 @@ pipeline {
 void verify_tag(project, version) {
     dir(project) {
         git url: "https://github.com/theforeman/${project}", branch: 'develop'
-        sh "test \"$(git tag -l ${version} | wc -l)\" -eq 1"
+        sh "git tag -l ${version} | grep ${version}"
     }
 }
 


xprazak2:
Updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -57,7 +57,10 @@ pipeline {
 void verify_tag(project, version) {
     dir(project) {
         git url: "https://github.com/theforeman/${project}", branch: 'develop'
-        sh "test \"$(git tag -l ${version} | wc -l)\" -eq 1"
+        def found_version = sh script: "git tag -l ${version} | grep ${version}", returnStdout: true
+        if (found_version.trim() != version) {
```
ekohl: This may be redundant. For me grep is sufficient to verify:
```
$ git tag -l 1.10.0
1.10.0
$ git tag -l 1.1
1.1
```

I think you can just rely on the exit code because of that. @evgeni can you confirm this?

xprazak2: Updated, I think you are right. As per [doc](https://jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#code-sh-code-shell-script), script returning with non-zero status will fail.

```sh
git tag -l  |  grep 1.17.0
1.17.0
echo $?
0

git tag -l  |  grep 1.20
echo $?
1
```

PR #640, opened 2018-05-22
Fix upload folders, create separate gemsets

xprazak2:
I changed the base_dir so that the tarballs are uploaded to proper directory and not all to 'foreman'.

Each project name is now added to a gemset name, so that parallel branches do not clash with each other.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
@@ -9,9 +9,13 @@ def gemset(name) {
     base_name
 }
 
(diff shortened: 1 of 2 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -9,11 +9,15 @@ def gemset(name) {
     base_name
 }
 
-def configureRVM(ruby = '2.0') {
-    withRVM(['rvm gemset empty --force'], ruby)
+def configureRVM(ruby = '2.0', name = '') {
+    emptyGemset(name, ruby)
```
ehelms: I think we may need to pass name to line 14

PR #641, opened 2018-05-23
Scaleway snapshot management

GregSutcliffe:
This deploys a cron script for creating snapshots over the Scaleway API. It's currently hardcoded to the AMS zone, where Discourse runs - we can expand the module to handle the Paris zone and snapshot Redmine once it's all working. It's quite verbose at the moment, we can tune that. Here's some sample output:

```
$ ruby ./manage_snapshots.rb
Scaleway Snapshot Management for discourse01.theforeman.org
Creating snapshot of 6f27df04-c0bf-4ded-bfa5-7626f40720d5
49dc1034-671f-43df-b98b-fb957f8c4827 in progress
// sits in 1 min loops for ~10-15 min
49dc1034-671f-43df-b98b-fb957f8c4827 created
Cleaning up snapshots on 6f27df04-c0bf-4ded-bfa5-7626f40720d5 - keeping 3
DELETE d2bb6d23-3865-4857-8f26-33e643689078: 204
// 204 is the correct return code for a successful delete
```

### DIFF
--- a/puppet/modules/scaleway/manifests/cron.pp
+++ b/puppet/modules/scaleway/manifests/cron.pp
@@ -1,0 +1,25 @@
+# Deploys a cronjob which calls the Scaleway API to
+# snapshot a server. Does not need to be on the Scaleway host.
+#
+define scaleway::cron (
+  String $api_key                   = 'NA',
+  String $org_id                    = 'NA',
+  Enum['present', 'absent'] $ensure = present,
+) {
+  file { "/etc/scaleway/${name}.yaml":
+    ensure  => $ensure,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0700', #root only, contains API keys
+    content => "api_key: <%= @api_key -%>\norg_id: <%= @org_id  -%>\n",
+  }
+
+  cron { "scaleway-snapshot-${name}":
+    ensure  => $ensure,
+    command => "/usr/bin/scaleway-snapshot ${name}",
+    user    => root,
+    hour    => '2',
+    minute  => '0'
+  }
+
+}
--- a/puppet/modules/scaleway/manifests/init.pp
+++ b/puppet/modules/scaleway/manifests/init.pp
@@ -1,0 +1,26 @@
+# Class to deploy snapshot management for Scaleway hosts
+#
+class scaleway(
+  Hash[String, Hash] $servers       = {},
+  Enum['present', 'absent'] $ensure = present,
+) {
+
+  # To hold the API config
+  file { '/etc/scaleway':
+    ensure => 'directory',
+    owner  => 'root',
+    group  => 'root',
+    mode   => 0700,
+  }
+
+  file { '/usr/bin/scaleway-snapshot':
+    ensure => $ensure,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
+    source => 'puppet:///modules/scaleway/manage_snapshots.rb',
+  }
+
+  # takes a hash like: { 'server' => { 'api_key' => 'abcde', ord_id => '12345' }, ...}
+  create_resources(scaleway::cron, $servers)
+}
--- a/puppet/modules/scaleway/files/manage_snapshots.rb
+++ b/puppet/modules/scaleway/files/manage_snapshots.rb
@@ -1,0 +1,118 @@
+#!/usr/bin/ruby
+#
+# Copyright Foreman Maintainers 2018
+# GPL-3.0
+#
+# This ruby script use the Scaleway API key to
+# ensure that recent snapshots of a volume are taken,
+# and that only the most recent snapshots are kept
+
+SERVER          = ARGV[0]
+exit 1 if SERVER.nil?
+
+SETTINGS        = YAML.load_file("/etc/scaleway/#{SERVER}.yaml")
+API_KEY         = SETTINGS[:api_key]
+ORG_ID          = SETTINGS[:org_id]
+
+require 'net/http'
+require 'json'
+require 'yaml'
+require 'date'
+require 'timeout'
+
+NUM_SNAPS = 3
+TIMEOUT   = 1800 # 30 min, snaps usually take ~10min
+
+puts "Scaleway Snapshot Management for #{SERVER}"
+
+def api_call(path,method='GET',args={})
+  uri = URI("https://cp-ams1.scaleway.com/#{path}")
+
+  if method == 'POST' then
+    req = Net::HTTP::Post.new(uri,'Content-Type' => 'application/json')
+    req.body = args.to_json
+  elsif method == 'DELETE' then
+    req = Net::HTTP::Delete.new(uri)
+  else
+    req = Net::HTTP::Get.new(uri)
+  end
+  req['X-Auth-Token'] = API_KEY
+
+  res = Net::HTTP.start(uri.hostname, uri.port, :use_ssl => true) {|http|
+    http.request(req)
+  }
+
+  if res.is_a?(Net::HTTPSuccess)
+    if res.body.nil?
+      return res.code
+    else
+      return JSON.parse(res.body)
+    end
+  else
+    raise "api call failed: #{res.code} - #{res.message}"
+  end
+end
+
+def get_volume_ids(name)
+  json = api_call('volumes')
+
+  json['volumes'].select {|v| v['server']['name'] == name}.map {|v| v['id'] }
+end
+
+def get_snapshots(vol_id)
+  json = api_call('snapshots')
+
+  json['snapshots'].select {|s| s['base_volume']['id'] == vol_id}
+end
+
+def create_snapshot(vol_id, server='unknown')
+  body = {
+    'name'         => "#{SERVER}-snapshot-#{Time.now}",
+    'volume_id'    => vol_id,
+    'organization' => ORG_ID
+  }
+  json = api_call('snapshots','POST',body)
+
+  json['snapshot']['id']
+end
+
+# Main execution
+
+# Loop over each volume for selected server
+volumes = get_volume_ids(SERVER)
+volumes.each do |vol|
+  # Start by creating the snap and waiting for it to complete, takes about 10 min
+  puts "Creating snapshot of #{vol}"
+  snap_id = create_snapshot(vol, SERVER)
+  puts "#{snap_id} in progress"
+  begin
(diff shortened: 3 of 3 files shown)


ehelms:
Given this a cron job, if the script is erroring we won't get any notifications I assume? Will there be a log of the script to debug if something seems amiss?

puppet/modules/scaleway/templates/manage_snapshots.erb
```diff
@@ -0,0 +1,120 @@
...
+# ensure that recent snapshots of a volume are taken,
+# and that only the most recent snapshots are kept
+
+# args
+#
+# name of the host to snapshot
+# api_key for auth
+# org_id is required by the api too
+
+SERVER  = <%= @name    -%>
```
ekohl: Should we make this a static file and read JSON/YAML files?

puppet/modules/scaleway/manifests/cron.pp
```diff
@@ -0,0 +1,15 @@
+# Deploys a cronjob which calls the Scaleway API to
+# snapshot a server. Does not need to be on the Scaleway host.
+#
+define scaleway::cron (
+  String $api_key = 'NA',
+  String $org_id  = 'NA'
+) {
+  file { "/etc/cron.daily/scaleway-snapshot-${name}":
```
ekohl: The downside is that there is no purging of unmanaged crons now. You probably want to make ensure a parameter.

GregSutcliffe:
@ehelms can you re-review?

puppet/modules/scaleway/manifests/cron.pp
```diff
@@ -0,0 +1,15 @@
+# Deploys a cronjob which calls the Scaleway API to
+# snapshot a server. Does not need to be on the Scaleway host.
+#
+define scaleway::cron (
+  String $api_key = 'NA',
```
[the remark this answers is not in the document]
GregSutcliffe: Good point, I'll drop the default.

puppet/modules/scaleway/templates/manage_snapshots.erb
GregSutcliffe: We could. Server could be $1 in the cronjob I guess too

puppet/modules/scaleway/manifests/cron.pp
GregSutcliffe: sounds good, will do

puppet/modules/scaleway/templates/manage_snapshots.erb
ekohl: That makes sense. The API key and org ID are probably good to not pass on the command line but server does make sense.

GregSutcliffe:
@ekohl updated. Left the default in and made ensure a param instead, that's cleaner. Left the server name in the script for now, else I have to deploy the generic script to /usr/bin *and* a line or file to cron to call it ...

puppet/modules/scaleway/manifests/cron.pp
```diff
@@ -0,0 +1,24 @@
...
+  String $org_id  = 'NA',
+  Boolean $ensure = present
```
ekohl: nit: we like trailing commas. Also think this should be `Enum['present', 'absent']` rather than a `Boolean`.

puppet/modules/scaleway/manifests/init.pp
```diff
@@ -0,0 +1,17 @@
+# Class to deploy snapshot management for Scaleway hosts
+#
+class scaleway(
+  Hash $servers = {},
```
ekohl: Can be `Hash[String, Hash]` to be even more specific.

puppet/modules/scaleway/templates/manage_snapshots.erb
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/ruby
+#
+# Copyright Foreman Maintainers 2018
+# GPL-3.0
+#
+# This ruby script use the Scaleway API key to
+# ensure that recent snapshots of a volume are taken,
+# and that only the most recent snapshots are kept
+
+SERVER          = <%= @name    -%>
```
ekohl: Should `SERVER` be a `ARGV` thing?

puppet/modules/scaleway/templates/manage_snapshots.erb
```diff
@@ -0,0 +1,116 @@
...
+SETTINGS        = YAML.load_file("/etc/scaleway/<%= @name -%>.yaml")
```
ekohl: I think this should be `/etc/scaleway/#{SERVER}.yaml`

puppet/modules/scaleway/templates/manage_snapshots.erb
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/ruby
```
ekohl: This file should be deployed in `init.pp`. It can probably also become a static file rather than a template.

GregSutcliffe:
Deploying this now, will add the required params to Foreman once it's imported.

puppet/modules/scaleway/manifests/cron.pp
GregSutcliffe: Show how long it's been since I wrote any serious Puppet. On it.

puppet/modules/scaleway/templates/manage_snapshots.erb
GregSutcliffe: Yeah ok, that makes sense

puppet/modules/scaleway/templates/manage_snapshots.erb
GregSutcliffe: Yeah, done

puppet/modules/scaleway/templates/manage_snapshots.erb
GregSutcliffe: Yeah ok, that makes sense

puppet/modules/scaleway/templates/manage_snapshots.erb
GregSutcliffe: Done

GregSutcliffe:
@ekohl updated, since the script is now static I've deployed it to bin and added a separate cronjob to call it.

ekohl:
> @ekohl updated, since the script is now static I've deployed it to bin and added a separate cronjob to call it.

Did you forgot to push?

GregSutcliffe:
Bah, pushed to greg:master instead of this branch. Updated now.

puppet/modules/scaleway/manifests/cron.pp
```diff
@@ -0,0 +1,24 @@
...
+) {
+  file { "/etc/scaleway/${name}.yaml":
+    ensure  => $ensure,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0700', #root only, contains API keys
+    content => "api_key: <%= @api_key -%>\norg_id: <%= @org_id  -%>\n",
+  }
+
+  cron { "scaleway-snapshot-${name}":
```
ekohl: This probably also needs `ensure => $ensure` so we can purge old ones.

puppet/modules/scaleway/manifests/init.pp
```diff
@@ -0,0 +1,25 @@
...
+  # To hold the API config
+  file { '/etc/scaleway':
+    ensure => 'directory',
+    owner  => 'root',
+    group  => 'root',
+    mode   => 0700,
+  }
+
+  file { '/usr/bin/scaleway-snapshot':
+    ensure  => $ensure,
```
ekohl: double space before `=>`

puppet/modules/scaleway/manifests/cron.pp
GregSutcliffe: good catch, ty

puppet/modules/scaleway/manifests/init.pp
GregSutcliffe: vim automation fail, fixed :)

GregSutcliffe:
back to you @ekohl :)

puppet/modules/scaleway/manifests/init.pp
```diff
@@ -0,0 +1,25 @@
...
+  # To hold the API config
+  file { '/etc/scaleway':
+    ensure => 'directory',
+    owner  => 'root',
+    group  => 'root',
+    mode   => 0700,
+  }
+
+  file { '/usr/bin/scaleway-snapshot':
+    ensure => $ensure,
```
ekohl: I don't think `$ensure` is defined here.

GregSutcliffe: dammit. Added an Enum for that :)

puppet/modules/scaleway/manifests/init.pp
```diff
@@ -0,0 +1,26 @@
...
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
+    source => 'puppet:///modules/scaleway/manage_snapshots.rb',
+  }
+
+  # takes a hash like: { 'server' => { 'api_key' => 'abcde', ord_id => '12345' }, ...}
+  create_resources(scaleway::cron, $servers)
```
ekohl: Technically you could add a third parameter that passes the default value of `$ensure` to each cron resource.

GregSutcliffe: See, now you're just *looking* for ways to block me :)

Wouldn't it be better to be able to delete specific cron jobs? Or is this just a default that can be overridden from the $servers hash? Either way, feels like we're splitting hairs now...

GregSutcliffe: @ekohl The scare with Scaleway's routers on Friday has reminded me that we should merge this sooner rather than later. Do you have objections to getting this merged and then iterating on it? If so, I'd like to work on this today, so let me know how you'd like to see it move forward

ekohl: You didn't notice that I did ack it - it's just that this would complete it. Personally I would have been to lazy to add an ensure parameter to this class and just hardcoded it to ensure present :)

The third parameter to `create_resources` is indeed the defaults for `$servers` and if that has an ensure parameter that takes precedence.

GregSutcliffe: You are correct, I did not notice. Thanks for the ACK, I'll get this rolled out now.

PR #647, opened 2018-05-25
address JJB warnings

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/properties/tfm-build-discarder.yaml
```diff
@@ -0,0 +1,9 @@
+- property:
+    name: tfm-build-discarder
```
ehelms: Switching as the default for all to this means we are now storing more data than we were on our master. Is that something we want to do given not all jobs have the same requirements or run patterns? For example, this sets `num-to-keep` to 30. If we have high pull request churn, some PRs will lose their PR data leaving reviewers unclear the state of a PR and having to re-run the tests again.

komidore64: I can understand the need for a couple different discarders. Maybe a standard and pull request discarder?

komidore64:
@ehelms reworked

PR #648, opened 2018-05-27
change some URLs to https

mmoll:


komidore64:
This looks sane to me!

Have you tested any of these `https` changes? I've run into GHPRB being very finicky with URLs so it would be nice to know that those work happily after setting `https`.

ekohl:
Thanks!

PR #649, opened 2018-05-28
Add forklift scm to release_test

xprazak2:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/builders/systest.yaml
```diff
@@ -46,7 +46,8 @@
               variables:
                 bats_environment:
                 - FOREMAN_EXPECTED_VERSION: ${{expected_version}}
-                ${{args:+foreman_installer_options:${{args}}}}
+                foreman_installer_options:
+                - '${{args}}'
```
komidore64: should this be intented over one?

```yaml
foreman_installer_options:
  - '${{args}}'
```

xprazak2: Perhaps, I wanted to keep it consistent with line 48, so I'll indent that as well.

PR #650, opened 2018-05-28
Run Redmine on HTTPS only with a redirect

ekohl:
This is untested. Probably best to roll it out with noop to see the diff before applying it.

### DIFF
--- a/puppet/modules/redmine/manifests/init.pp
+++ b/puppet/modules/redmine/manifests/init.pp
@@ -151,17 +151,12 @@ class redmine (
   }
 
   apache::vhost { $servername:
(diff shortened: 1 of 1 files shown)


ekohl:
@GregSutcliffe @mmoll I'd like to roll this out soonish (not today due to other priorities). Could you have a look?

GregSutcliffe:
LGTM. I was planning to roll out the snapshot script later on today as well, I can add this to my list.

GregSutcliffe:
Deploying now, hang tight :)

GregSutcliffe:
All good apart from #696 which is also merged. Ty @ekohl

PR #654, opened 2018-05-31
don't build core package on armv7 for 1.18+

mmoll:


mmoll:
By leaving armv7 in for the dependency job and the the fact that -installer and -proxy are `Architecture: all`, this still allows proxy installs on ARMv7.

mmoll:
let's see if that filter is really correct ;)

PR #658, opened 2018-05-31
start tracking jenkins plugins

komidore64:
I grabbed all the plugins from our jenkins instance and slapped them into a puppetfile.

I'm looking for guidance on what else i'd need to make this work properly.


ehelms:
I know this parities the `slave` module, but I might opt that we go tthe `jenkins_master` naming route to be explicitly clear (and not confuse it with something like puppet master).

I think my biggest question is how does this manifest know to talk to our Jenkins? Anything configuration wise that needs to be set?

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,722 @@
+class () {
+
+  jenkins::plugin { 'build-metrics':
```
ekohl: You can write this shorter:
```puppet
$plugins = {
  'build-metrics' => '1.3',
  'htmlpublisher' => '1.16',
}

$plugins.each |$plugin, $version| {
  jenkins::plugin { $plugin:
    version => $version,
  }
}
```

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,722 @@
+class () {
```
ekohl: This needs to be `class jenkins_master {`

komidore64:
ah okay, looks like i need to have set a [`node`](https://github.com/voxpupuli/puppet-jenkins#using-puppet-jenkins)

puppet/modules/jenkins_master/manifests/init.pp
komidore64: oh that looks much nicer!

ekohl:
> ah okay, looks like i need to set a node

Notes are set in manifests but since we use Foreman as the ENC that's not needed.

Can you have a look at the loop I suggested? The module is a lot shorter then

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,197 @@
+class jenkins_master {
+
+  jenkins::config { 'config_hash':
+    'JENKINS_JAVA_OPTIONS' => {
+      'value' => '-Djava.awt.headless=true -Djenkins.install.runSetupWizard=false -Xms2048m -Xmx2048m',
```
mmoll: iirc this is coming right now from foreman (ENC)

komidore64: Yes it is.

My goal is to get as much of our Jenkins into source control and code as possible to ease rebuilding if necessary in the future. I want our Jenkins to be [cattle instead of a pet](https://www.engineyard.com/blog/pets-vs-cattle)

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,197 @@
+class jenkins_master {
+
+  jenkins::config { 'config_hash':
```
ehelms: Note that this class has no parameters (https://github.com/voxpupuli/puppet-jenkins/blob/master/manifests/config.pp#L3) and gets the hash by configuring it on the main class:

https://github.com/voxpupuli/puppet-jenkins/blob/master/manifests/config.pp#L13
https://github.com/voxpupuli/puppet-jenkins/blob/master/manifests/init.pp#L74

komidore64:
> A things I noticed we should take note of:
> 
> 1. https://github.com/voxpupuli/puppet-jenkins#verifying -- should we include checksums?
> 2. https://github.com/voxpupuli/puppet-jenkins#plugin-dependencies -- Jenkins through its update manager does this, but sadly the API around this is not as spec'd out it appears

I'm not opposed to including checksums. Would you prefer we did?

ehelms:
The checksums seem fairly intensive to compute so I am OK skipping that for now. It'd be worth following p to figure out why puppet-jenkins does not download via HTTPS and correct that instead.

komidore64:
I'm not really assure what even you're wanting to address with (2).

komidore64:
@ehelms I included all the plugins that are currently installed on our jenkins instance, so I don't think dependencies will be an issue with our current plugin list?

ehelms:
@komidore64 ACK... I was more worried about what we do in the future when adding a plugin, do you think we should cross that bridge later?

komidore64:
@ehelms we could write a comment blurb above the plugins remind folks to be mindful of dependencies?

komidore64:
yeah, i'm in favor of crossing that bridge when we get there

ehelms:
This still shows WIP -- should that be dropped now or is there more work coming?

@mmoll How do you feel about this change given you do a lot of Jenkins management?

mmoll:
If it works, I'm all in. As always, we can patch the manifests on the puppet master locally and test before merging.

ehelms:
@mmoll Sounds like a good plan, but not a workflow I am familiar or comfortable enough with to try.

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,199 @@
+class jenkins_master {
+
+  jenkins { 'jenkins_master':
```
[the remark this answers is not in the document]
komidore64: updated!

ehelms:
@mmoll or @ekohl Either you feel up to testing this locally? Or outlining how to in more detail and I can give it a go?

ekohl:
@ehelms I would be willing to have a look but given my current backlog it might not happen soon. I think in general what you could do is:
* Put puppet on jenkins in noop
* Apply this patch locally on the puppet master
* See what would be changed
* Change files locally on the master till you get the desired result
* Remove noop and actually apply
* Change till you get good results
* Update the PR with the resulting changes

It's not pretty, but often it's the best way. An intermediate step could be to spin up a vagrant box and apply it there.

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,201 @@
...
```
ehelms: I think this needs to be:

```
class { 'jenkins':

}
```

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -0,0 +1,201 @@
...
+    'cloudbees-folder'                   => '6.4',
+    'cobertura'                          => '1.12.1',
+    'command-launcher'                   => '1.2',
+    'conditional-buildstep'              => '1.3.6',
+    'config-file-provider'               => '2.18',
+    'configurationslicing'               => '1.47',
+    'copy-to-slave'                      => '1.4.4',
+    'copyartifact'                       => '1.39.1',
+    'create-fingerprint'                 => '1.2',
+    'credentials'                        => '2.1.16',
```
ehelms: This plugin is included by default in the jenkins module and causes a conflict so we can omit it here or comment it out with a comment. Take a look at:

https://github.com/voxpupuli/puppet-jenkins/blob/master/manifests/config.pp#L9
https://github.com/voxpupuli/puppet-jenkins/blob/ef41da82f100a10f2dcfd04c6286d504452d3988/manifests/params.pp#L29-L31 (note you can't look at this in master cause we don't use master of this module)

komidore64: updated

puppet/modules/jenkins_master/manifests/init.pp
komidore64: updated

ehelms:
@mmoll @ekohl I tested this, and have updated (I think) the host group properly on Foreman. Going to merge, so please keep an eye out for any weirdness or lack of this actually working.

PR #660, opened 2018-05-31
don't use ruby's safe navigation operator

komidore64:
the machine where jenkins currently lives is using ruby 2.0. the
safe navigation operator wasn't introduced until ruby 2.3.


ehelms:
Why can't we? The commit message is a bit mis-leading or could be more informative for future devs looking back.

PR #662, opened 2018-05-31
Add RPM lint check on packaging PRs

ehelms:


ehelms:
I doubt any actually have 0 warnings, but I'd be surprised if any had actual errors.

PR #664, opened 2018-06-02
Handle the case where there are no unmanaged jobs

ekohl:
This still doesn't make it idempotent, but it shouldn't bork when
unmanaged_jobs.rb returns an empty string because there are no unmanaged
jobs.

Note this is untested and there's probably a much cleaner way to do this.

### DIFF
--- a/puppet/modules/jenkins_job_builder/manifests/config.pp
+++ b/puppet/modules/jenkins_job_builder/manifests/config.pp
@@ -41,7 +41,7 @@ define jenkins_job_builder::config (
   }
 
   exec { "remove_unmanaged_jobs-${config_name}":
-    command => "jenkins-jobs --conf ${inifile} delete --jobs-only $(ruby ${directory}/${config_name}/unmanaged_jobs.rb ${inifile})",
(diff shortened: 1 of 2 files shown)


puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -41,7 +41,7 @@
   }
 
   exec { "remove_unmanaged_jobs-${config_name}":
-    command => "jenkins-jobs --conf ${inifile} delete --jobs-only $(ruby ${directory}/${config_name}/unmanaged_jobs.rb ${inifile})",
+    command => "${directory}/${config_name}/unmanaged_jobs.rb ${inifile}",
```
xprazak2: shouldn't it be `ruby ${directory}/...` ?

ekohl: You might be right, not sure if the file is marked as executable.

ekohl: Updated

PR #665, opened 2018-06-04
Fix groovy syntax

ekohl:
3f9a3f015bbd06b2695cb14508968bc87230ce4e introduced a regression in the groovy syntax.


mmoll:
merged, bedankt @ekohl!

PR #666, opened 2018-06-04
Clean the workspace when releasing tarballs

ekohl:
This ensures we have no files like Gemfile.lock left behind from any previous run.


mmoll:
merged, bedankt @ekohl!

PR #667, opened 2018-06-04
Add a job to build 3.7 sources from git and release RPM

ehelms:


ehelms:
Updated with version info

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/version_map.groovy
```diff
@@ -0,0 +1,5 @@
+def versions = [
```
ekohl: We have a similar mapping in the katello job. Can we unify that in 1 file so we can't forget one?

ehelms:
Updated -- please check the logic introduced for that update

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.7/rpmRelease.groovy
```diff
@@ -0,0 +1,76 @@
...
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Build Source') {
+            parallel {
+                stage('build-katello-installer-tarball') {
+                    steps {
+                        dir('katello-installer') {
+                            configureRVM('2.2')
```
ekohl: Should we just make this 2.4? Also in a variable so it's only in 1 place?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -16,7 +10,7 @@ pipeline {
                 ghprb_git_checkout()
 
                 dir('foreman') {
-                    git url: "https://github.com/theforeman/foreman", branch: katello_versions[ghprbTargetBranch]
+                    git url: "https://github.com/theforeman/foreman", branch: foremanVersion(ghprbTargetBranch)
```
ekohl: This doesn't append the `-stable` part. The job definition probably also needs to include `version_map.groovy`

ehelms:
Updated

ekohl:
Given we already pushed an RC1, is this still wise to do?

ehelms:
Wise? Absolutely. This should be valid for every release. The technological solution is the tough part.

ekohl:
Given 3.7 is now EOL, should we close this or update it to the current version?

PR #671, opened 2018-06-05
Use the correct cleanup method in tarballsRelease

ekohl:
The clearWorkspace didn't work. This uses the CleanCheckout extension
which does appear to work.

We also match the exact git URLs for consistency.


dLobatog:
http://ci.theforeman.org/blue/organizations/jenkins/tarballs-release/detail/tarballs-release/38/pipeline

it's working gif

mmoll:
merged. bedankt @ekohl!

PR #673, opened 2018-06-05
long options for rsync makes for easier reading

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/deploy_web.sh
```diff
@@ -29,4 +29,4 @@ TARGET_PATH="website@theforeman.org:rsync_cache/"
 # Export this to avoid quoting issues
 export RSYNC_RSH="ssh -i /var/lib/workspace/workspace/rsync_web_key"
 
-/usr/bin/rsync -acvxz --delete-after ./_site/ $TARGET_PATH
+/usr/bin/rsync --archive --checksum --verbose --one-file-system --compress --stats --progress --delete-after ./_site/ $TARGET_PATH
```
ehelms: I count 5 options in the original and 7 here. Were options also added?

komidore64: `--stats` and `--progress` were added.

PR #674, opened 2018-06-05
only delete unmanaged jobs at midnight

komidore64:
this'll hopefully give people a bit more time to fiddle with test jobs before they get garbage collected

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/unmanaged_jobs.rb
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/unmanaged_jobs.rb
@@ -48,8 +48,8 @@ def main
 
   unmanaged_jobs = find_unmanaged_jobs_from_payload(payload)
 
-  unless unmanaged_jobs.empty?
-    `jenkins-jobs --conf #{inifile} delete --jobs-only #{format_jobs_for_output(unmanaged_jobs)}`
+  if Time.now.hour.zero? && !unmanaged_jobs.empty?
+    %x[jenkins-jobs --conf #{inifile} delete --jobs-only #{format_jobs_for_output(unmanaged_jobs)}]
   end
 end
 


puppet/modules/jenkins_job_builder/files/theforeman.org/unmanaged_jobs.rb
```diff
@@ -48,8 +48,8 @@ def main
 
   unmanaged_jobs = find_unmanaged_jobs_from_payload(payload)
 
-  unless unmanaged_jobs.empty?
-    `jenkins-jobs --conf #{inifile} delete --jobs-only #{format_jobs_for_output(unmanaged_jobs)}`
+  if Time.now.hour.zero? && !unmanaged_jobs.empty?
+    %x[jenkins-jobs --conf #{inifile} delete --jobs-only #{format_jobs_for_output(unmanaged_jobs)}]
```
ehelms: What was the reason for changing the invocation to `%x`?

komidore64: That's just a [style thing](http://rubocop.readthedocs.io/en/latest/cops_style/#enforcedstyle-percent_x) that rubocop usually enforces.

puppet/modules/jenkins_job_builder/files/theforeman.org/unmanaged_jobs.rb
```diff
@@ -48,8 +48,8 @@ def main
 
   unmanaged_jobs = find_unmanaged_jobs_from_payload(payload)
 
-  unless unmanaged_jobs.empty?
-    `jenkins-jobs --conf #{inifile} delete --jobs-only #{format_jobs_for_output(unmanaged_jobs)}`
+  if Time.now.hour.zero? && !unmanaged_jobs.empty?
```
ekohl: I think we should avoid doing anything earlier. So not do an API request.

PR #675, opened 2018-06-05
let's not make an unnecessary API requests if we don't delete jobs

komidore64:


ekohl:
Should we maybe change this into a cronjob rather than running it via puppet? Other than that I think this is an improvement.

ehelms:
This feels more traceable than a cronjob for me.

PR #678, opened 2018-06-06
Run RPM lint in parallel

ehelms:


ehelms:
An idea I had to get cleaner failure output when looking at the job, this should then print out a section for each package being linted to more clearly see which ones failed. I was thinking of applying something similar to RPM build part of the job too if this worked.

ehelms:
@ekohl Exactly, the ole do we let Jenkins parallelize or Ansible. And as you said, the nice part of Jenkins is we get a visual representation.

PR #681, opened 2018-06-07
Ensure obal tito extensions are available

ehelms:


xprazak2:
Solves the issue for me.

PR #684, opened 2018-06-08
Test asset precompile for Foreman core

ehelms:


ehelms:
See https://github.com/theforeman/foreman/pull/5671

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,15 @@ bundle install --without=development --jobs=5 --retry=5
 # we need to install node modules for integration tests (which only run on postgresql)
 if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
   npm install npm@'<6.0.0' # first upgrade to newer npm
+
+  # Test asset precompile
+  NODE_ENV=production $APP_ROOT/node_modules/.bin/npm install
+  bundle exec rake assets:precompile RAILS_ENV=production
+  bundle exec rake webpack:compile RAILS_ENV=production
+
+  rm -rf public/assets public/webpack
```
mmoll: Could you add package-lock.json here, just to be sure?

ehelms:
Good call -- updated

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,15 @@ bundle install --without=development --jobs=5 --retry=5
...
   npm install npm@'<6.0.0' # first upgrade to newer npm
+
+  # Test asset precompile
+  NODE_ENV=production $APP_ROOT/node_modules/.bin/npm install
+  bundle exec rake assets:precompile RAILS_ENV=production
+  bundle exec rake webpack:compile RAILS_ENV=production
+
+  rm -rf public/assets public/webpack package-lock.json
+
+  # Now install all production and dev dependencies
```
tbrisker: Do we actually need the dev dependencies on jenkins after the proposed change?

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,15 @@ bundle install --without=development --jobs=5 --retry=5
 # we need to install node modules for integration tests (which only run on postgresql)
 if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
   npm install npm@'<6.0.0' # first upgrade to newer npm
+
+  # Test asset precompile
+  NODE_ENV=production $APP_ROOT/node_modules/.bin/npm install
+  bundle exec rake assets:precompile RAILS_ENV=production
+  bundle exec rake webpack:compile RAILS_ENV=production
```
tbrisker: We already run this step without the production option in the jenkins:integration task, would it make sense to remove it from there and just run once, having jenkins run the tests on the production version of the bundle?

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
ehelms: Are there not JS unit tests that they are required to run for?

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
ehelms: I think you still want that in the rake task so that the rask task is independently runnable (e.g. locally) ?

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
tbrisker: I may be mistaken, but I think we only run js unit tests on travis.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,6 +24,15 @@ bundle install --without=development --jobs=5 --retry=5
 # we need to install node modules for integration tests (which only run on postgresql)
 if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
   npm install npm@'<6.0.0' # first upgrade to newer npm
+
+  # Test asset precompile
+  NODE_ENV=production $APP_ROOT/node_modules/.bin/npm install
```
sharvit: I am not sure I understand this line,
I never seen projects running npm from `./node_modules/.bin/npm`.

Does it mean we are running a local `npm` that we install using the global `npm`?
Why not to use the global `npm`?

Is this the line that install the `npm` locally with `npm`? `npm install npm@'<6.0.0'`?
Can we change it to `npm install --global npm@'<6.0.0'`?

Sorry for all the questions, I just want to make sure I understand.

ekohl: We generally don't want to use any global because it can interfere with other runs. The same jenkins slave could run some other job in parallel that installs another npm version. This provides us with some isolation. webpack-rails [does the same](https://github.com/mipearson/webpack-rails/blob/83a5d4c8af82e1f33116e3550d3bc3e8ecff977f/lib/webpack/railtie.rb#L18).

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
dLobatog: We still need them, as integration tests will indirectly check the JS works.

ehelms:
@mmoll I updated this based on the outcome from the Foreman PR, so I think this needs a re-review

mmoll:
merged, thanks @ehelms!

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -24,7 +24,11 @@ bundle install --without=development --jobs=5 --retry=5
 # we need to install node modules for integration tests (which only run on postgresql)
 if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
   npm install npm@'<6.0.0' # first upgrade to newer npm
+
+  # Test asset precompile
   $APP_ROOT/node_modules/.bin/npm install
```
ohadlevy: you could make this faster with --no-optional

PR #685, opened 2018-06-08
remove test_katello and children from jenkins

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/build/build_rpm_rubygem-katello_nightly.yaml
```diff
@@ -0,0 +1,32 @@
+- job:
+    name: build_rpm_rubygem-katello_nightly
+    node: fast
+    properties:
+      - tfm-build-discarder
+    triggers:
+      - timed: '0 18 * * *'
```
ehelms: What about keeping https://github.com/theforeman/foreman-infra/pull/685/files#diff-0141dbd04e4380140b3d3ea37281a53eL24 so that this builds a new RPM every time a change is introduced (PR is merged) ?

komidore64: yeah, we can do that

komidore64: updated

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/build/build_rpm_rubygem-katello_nightly.yaml
```diff
@@ -0,0 +1,32 @@
...
+    wrappers:
+      - abort_build:
+          timeout_min: 60
+      - ansicolor
+    builders:
+      - shell: !include-raw: scripts/build_rpm_rubygem-katello_nightly.sh
+    publishers:
+      - archive:
+          artifacts: 'pkg/*'
+      - trigger-parameterized-builds:
```
ehelms: With the change to this being a publisher, this job will now look like it succeed if all it did was generate the gem successfully which I think is confusing based upon the name of the job. This is partially the reason that this was a builder before and not a publisher.

komidore64: I set it as a publisher due to how `packaging_build_rpm` looks for the `lastSuccessfulBuild` of a referenced job to get the source. I do understand that a more accurate name could mitigate confusion.

ehelms: You can keep this as a builder but adopting the same strategy used in the newer hammer-cli and katello-installer build jobs. Take a look at:

https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRelease.groovy#L24-L30

komidore64: updated!

komidore64:
renamed `build_rpm_rubygem-katello-nightly` to `rubygem-katello-nightly-release`.

PR #689, opened 2018-06-15
convert rubygem-katello-nightly-release to groovy (baby)

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemKatelloNightlyRelease.groovy
```diff
@@ -0,0 +1,36 @@
...
+        timeout(time: 60, unit: 'MINUTES')
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage ("Build Gem") {
+            steps {
+                dir('katello') {
+                    configureRVM('2.2')
```
ekohl: I generally like magic numbers to be in variables, especially if they're being reused :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemKatelloNightlyRelease.groovy
```diff
@@ -0,0 +1,36 @@
...
+                    git url: "https://github.com/Katello/katello", branch: 'master'
+                    withRVM(["gem build katello.gemspec"], '2.2')
+                    archiveArtifacts artifacts: '*.gem'
+                }
+            }
+        }
+
+        stage('Trigger RPM Build') {
+            steps {
+                build job: 'packaging_build_rpm', propagate: true, parameters: [
```
ekohl: I wonder if we could convert this job to a groovy step so you see everything inside a single job. We could reuse it in every nightly pipeline.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemKatelloNightlyRelease.groovy
komidore64: magic number has been variablized

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemKatelloNightlyRelease.groovy
komidore64: I would like to do that, but I think that'd outside the scope of this pull request. I will make a note that this is what we want to do down the road.

PR #693, opened 2018-06-18
fix hammer-cli-katello-nightly-release job

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRelease.groovy
```diff
@@ -1,9 +1,13 @@
 pipeline {
     agent { label 'admin' }
 
-    option {
+    environment {
+        RUBY = '2.2'
```
mmoll: IMHO this should be 2.4

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRelease.groovy
```diff
@@ -12,19 +16,21 @@ pipeline {
         stage ("Build Gem") {
             steps {
                 dir('hammer_cli_katello') {
+                    configureRVM(env.RUBY)
                     git url: "https://github.com/Katello/hammer-cli-katello", branch: 'master'
-                    sh "gem build hammer_cli_katello.gemspec"
-                    archiveArtifacts artifacts '*.gem'
+                    withRVM(["gem build hammer_cli_katello.gemspec"], env.RUBY)
```
ehelms: Does this need an RVM environment configured? Just curious if that's overhead we are incurring or required to `gem build`

komidore64: I found that without configuring an RVM environment, I didn't have access to the `gem` command.

ehelms: Odd, the nodes should have Ruby installed. And that was working if I look randomly at http://ci.theforeman.org/job/hammer-cli-katello-nightly-release/66/console

komidore64:
@ehelms updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/hammerCliKatelloNightlyRelease.groovy
ekohl: Perhaps they have the `ruby` package but not always `rubygems`? We can just ensure that via the puppet `slave` module with `ensure_packages`.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/hammer-cli-katello-nightly-release.yaml
```diff
@@ -3,7 +3,7 @@
     project-type: pipeline
     sandbox: false
     triggers:
-      - timed: '0 22 * * *'
+      - github
```
ehelms: I think for this to work we need to specify something like -- 

```
    properties:
      - github:
          url: https://github.com/Katello/hammer-cli-katello
```

komidore64: That isn't necessary. The [docs](https://docs.openstack.org/infra/jenkins-job-builder/triggers.html#triggers.github) don't say anything about setting a `github` property.

As a more applicable example [rubygem-katello-nightly-release](http://ci.theforeman.org/job/rubygem-katello-nightly-release/) has been running as expected since its groovy conversion [occurred](https://github.com/theforeman/foreman-infra/pull/689) on 16 June.

There have been five builds triggered since the 16th.

ehelms: Fair, the Github plugin docs make me think it's required -- https://wiki.jenkins.io/display/JENKINS/GitHub+Plugin (Hyperlinks Between Pages section) but mus tbe just for that feature.

PR #696, opened 2018-06-19
Fix Apache header syntax

ekohl:
Rules require an action, in this case set.


puppet/modules/redmine/manifests/init.pp
```diff
@@ -176,7 +176,7 @@
       ssl_chain               => "/etc/letsencrypt/live/${servername}/chain.pem",
       ssl_key                 => "/etc/letsencrypt/live/${servername}/privkey.pem",
       headers                 => [
-        'Strict-Transport-Security: max-age=86400;',
+        'set Strict-Transport-Security: max-age=86400;',
```
ekohl: Btw, I initially set this to 1 day in case of issues. If this works well, we should increase it to 6 months or a year.

GregSutcliffe: +1 to that, will you add it to your todo?

PR #697, opened 2018-06-19
remove rvm from rubygem-katello-nightly-release

komidore64:
it is not necessary


ehelms:
I think the commi tmessage is a bit misleading? Should it be `from rubygem-katello-nightly-release` ?

komidore64:
i can't read apparently!

komidore64:
@ehelms there we go!

PR #698, opened 2018-06-19
Add jenkins job playbook for running a job on a different jenkins

zjhuntin:


puppet/modules/jenkins_job_builder/files/centos.org/ansible/jenkins_job.yml
```diff
@@ -0,0 +1,85 @@
...
+        method: GET
+        return_content: yes
+        timeout: 3
+        user: "{{ jenkins_username }}"
+        password: "{{ jenkins_password }}"
+        url: "{{ jenkins_host }}/job/{{ jenkins_job_name }}/{{ poll_result.json.executable.number }}/api/json"
+      register: build_job_result
+      until: build_job_result.json.result == 'SUCCESS' or build_job_result.json.result == 'FAILURE'
+      retries: 280
+      delay: 20
```
ehelms: I think this comes out to about 4.5 hours that it will wait? This should run in under 2 hours, so maybe we should drop the retries closer to that?

zjhuntin: Sure, that works for me, I was going for overkill at the time. Having it act as a limiter is probably a wiser idea.

ehelms:
@ekohl I figured we could do in the future as well. For now, we see if debugging is more or less difficult and avoid incurring the bandwidth to transfer back, time to do so and potential to fail at that point.

PR #687, opened 2018-06-14
Add katello-foreman-nightly test

zjhuntin:

PR #699, opened 2018-06-19
Adjust katello-nightly-release to use ci.centos.org when testing inst…

zjhuntin:
…allation tests

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloRelease.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloRelease.groovy
@@ -59,33 +59,12 @@ pipeline {
         stage('Install Test') {
             agent { label 'el && ipv6' }
 
-            environment {
-                VAGRANT_DEFAULT_PROVIDER = 'openstack'
-            }
             steps {
 
-                git url: 'https://github.com/theforeman/forklift'
+                git url: 'https://github.com/theforeman/foreman-infra'
 
-                sh "cp -f vagrant/boxes.d/99-local.yaml.example vagrant/boxes.d/99-local.yaml"
-                sh "vagrant up centos7-katello-bats-ci"
-
-            }
-            post {
-                always {
-                    sh "mkdir debug"
-                    sh "vagrant ssh-config centos7-katello-bats-ci > ssh_config"
-
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/root/bats_results/*.tap debug/ || true"
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/root/last_logs debug/ || true"
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/root/sosreport* debug/ || true"
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/root/foreman-debug.tar.xz debug/ || true"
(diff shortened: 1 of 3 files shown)


zjhuntin:
Requires: https://github.com/theforeman/foreman-infra/pull/698

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katelloRelease.groovy
```diff
@@ -60,33 +60,12 @@ pipeline {
...
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/root/foreman-debug.tar.xz debug/ || true"
-                    sh "scp -F ssh_config centos7-katello-bats-ci:/var/log/foreman-installer/katello.log debug/ || true"
-
-                    sh "vagrant destroy centos7-katello-bats-ci"
-
-                    archive "debug/*"
-                    script { step([$class: "TapPublisher", testResults: "debug/*.tap"]) }
-                    deleteDir()
+                withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'jenkins-centos', passwordVariable: 'PASSWORD', usernameVariable: 'USERNAME']]) {
+                    runPlaybook('ci/centos.org/ansible/run_jenkins_job.yml', 'localhost', "[jenkins_job_name=foreman-nightly-test, jenkins_username=${env.USERNAME}, jenkins_password=${env.PASSWORD}]"
```
ekohl: The name here is `foreman-nightly-test` but the pipeline is for katello. That doesn't sound right.

zjhuntin:
- Do we archive the debug logs somewhere?

Yes, they will be archived on the centos job this will kick off. It will archive a sosreport, foreman debug, and bats tests results. I was unsure how to link the centos job with this outside of looking in the console for it, but we probably can have some way to reach it directly from a job.

- Do we publish the test results somewhere?

The tests will be freely available to see on the centos jenkins job that this will start, so that anyone can see it. The job being this one: https://github.com/theforeman/foreman-infra/pull/687, once it has been published.

- How would I replicate this to debug the issue? (either as packager or developer)

We are not changing the pipeline that is being run, more trying to outsource it to hardware that will be able to run it without any memory issues. A user can see from the connected job what it is being run. I admit that it abstracted, but they would be able to follow the instructions from the centos job.

zjhuntin:
I agree that it is difficult to see when a job fails and then following a workflow to replicate it. I think that perhaps we should take a different strategy with this then. Instead of trying to have them read the job from any Jenkins to try and follow the workflow, which is by necessity going to do some things you will not have to do locally, to instead write out how these can be replicated locally. This could be through solved through a couple of ways off the top of my head.

- Write out how to replicate these important jobs locally using forklift. We could host these on theforeman.org and then link each job to their page.

- Write docker compose files to run these locally, the biggest advantage here being they just have to get docker working, and not have have libvirt, vagrant etc to run it.

ehelms:
For me, this is the same as previous from a developer perspective except what has changed is instead of `vagrant up centos7-katello-bats-ci` it is now 'ansible-playbook pipelines/pipeline_katello_nightly.yml`. That aspect we can communicate more clearly through Forklift docs and potentially infra docs.

ekohl:
@zjhuntin a documented way of debugging these issues on theforeman.org sounds like great idea.

ehelms:
I thnk thats a good idea, but not a blocker given this does not change the status quo with respect to debugging.

PR #701, opened 2018-06-19
seperate out asset precompile test

mmoll:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -44,3 +43,10 @@ bundle exec rake db:drop db:create db:migrate --trace
 tasks="pkg:generate_source jenkins:unit"
 [ ${database} = postgresql ] && tasks="$tasks jenkins:integration"
 bundle exec rake $tasks TESTOPTS="-v" --trace
+
+# Test asset precompile
+if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
+  bundle exec rake db:drop db:create db:migrate RAILS_ENV=production
```
ekohl: In the past we've seen you can't do this in 1 line. I think the `db:migrate` needs to be a separate rake invocation due to caching.

ehelms: We may need to explicitly drop the production database here https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/gemset_cleanup.sh#L20

mmoll:
updated.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -44,3 +43,11 @@ bundle exec rake db:drop db:create db:migrate --trace
 tasks="pkg:generate_source jenkins:unit"
 [ ${database} = postgresql ] && tasks="$tasks jenkins:integration"
 bundle exec rake $tasks TESTOPTS="-v" --trace
+
+# Test asset precompile
+if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
+  bundle exec rake db:drop db:create RAILS_ENV=production
+  bundle exec rake db:migrate RAILS_ENV=production
+  bundle exec rake assets:precompile RAILS_ENV=production
+  bundle exec rake webpack:compile RAILS_ENV=production
```
tbrisker: Should this also be `NODE_ENV=production` so we get the production bundle built?

ekohl: Yes, https://github.com/mipearson/webpack-rails/blob/a6faf5d2fc1a76481ddf640fb799245d719a7951/lib/tasks/webpack.rake#L5 does this.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
mmoll: @ehelms so I should add the db:drop with RAILS_ENV=production there, correct?

ekohl: I'd prefer `gemset_cleanup` not to do any rails tasks. Currently the kafo job is already broken because of it (needs `PUPPET_VERSION` to be set for bundle exec to work and the `|| true` doesn't appear to work). Can we introduce a `rails_cleanup` instead and use that where needed?

ehelms:
The production database that's created still needs to get cleaned up if success or failure I think.

mmoll:
I added the drop to gemset_cleanup.sh for now, to unbreak things and refactor later...

lzap:
Hello guys, this step is breaking discovery plugin test for some reason. 

https://ci.theforeman.org/job/test_plugin_matrix/653/database=postgresql,ruby=2.3,slave=fast/consoleFull

mmoll:
`12:08:07 /usr/local/rvm/gems/ruby-2.3.5@test_plugin_matrix-1/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `require': /home/jenkins/workspace/test_plugin_matrix/database/postgresql/ruby/2.3/slave/fast/plugin/test/unit/discovery_attribute_set_test.rb:52: syntax error, unexpected keyword_ensure, expecting keyword_end (SyntaxError)`

This requires Ruby 2.5.

PR #702, opened 2018-06-19
[RFC] Document Jenkins Job Naming conventions

ekohl:
This is very much a work in progress and I know I forgot many conventions. Please suggest those I forgot :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,15 @@ This repo contains puppet modules that are used to manage infrastructure used by
 
 ### Updates
 For more information what's currently being worked on, see the [Infrastructure Updates](https://projects.theforeman.org/projects/foreman/wiki/Infrastructure_Updates) page in the Foreman wiki.
+
+### Jenkins Job Naming conventions
+
+We're starting to implement some some job naming conventions.
+
+**Note** Because `centos.org` is a shared environment all jobs are prefixed by `foreman-` to denote they're ours.
+
+| **Name**                | **Convention**                        | **Example 1**            | **Example 2**                     |
+|-------------------------|---------------------------------------|--------------------------|-----------------------------------|
+| Nightly Package Builder | {git-repo}-{git-branch}-release       | foreman-develop-release  | hammer-cli-katello-master-release |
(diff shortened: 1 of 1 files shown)


ehelms:
Just checking if this should go in here in stead -- https://github.com/theforeman/foreman-infra/tree/master/puppet/modules/jenkins_job_builder

ehelms:
PR tests follow:

```
{git-repo}-{optional concern}-pr-test
```

Where {optional-concern} is for things like packaging repo that may have a job for rpms and another for deb

ehelms:
You could mention also that general rule of thumb is specific -> general (going left to right) with naming conventions. Where the left hand side generally denotes a project of some type and the right hand a categorization.

komidore64:
`{git-repo}-{optional-concern}-pr-test` is ideal for pull requests?

I've written a handful of jobs that are `{git-repo}-pull-request` so I guess those will need to be changed :)

ehelms:
Well, do we want to standardize on pr-test ? Since its shorter? I think we should add what we know here and merge it. Then we can iterate without leaving this open.

ekohl:
I'd prefer `pr-test`.

komidore64:
i'd prefer `*-pr` only because we may want to run other things than just tests on a pull-request, but that's a nitpick.

I won't be upset if i'm overruled for `*-pr-test`.

ehelms:
That's fair, `-test` is perhaps a bit extraneous unless we started doing things like:

```
blah-pr-test
blah-pr-rubocop
```

Maybe?

ehelms:
I say lets add the pr one and merge this.

komidore64:
:+1:

ekohl:
> You could mention also that general rule of thumb is specific -> general (going left to right) with naming conventions. Where the left hand side generally denotes a project of some type and the right hand a categorization.

Following this logic it would be `{repo}-{concern}-{type}-{pr}` - i.e. `foreman-test-pr` and `foreman-rubocop-pr`, right?

ekohl:
Added PR testing and moved it to the main README since that's not really big anyway and guarantees visibility.

PR #704, opened 2018-06-21
Bind the katello test step to ipv6

ekohl:
This ensures it runs correctly on a node with IPv6 which we need for
forklift.


mmoll:
merged, bedankt @ekohl!

PR #707, opened 2018-06-22
Correct vagrant in foreman-discovery-image-publish

ekohl:
This now uses the openstack provider which we've switched to a while
back. We drop the --debug in provision because that leaks API keys.

I haven't tested this yet, but it should be enough.


lzap:
Does not work yet: http://ci.theforeman.org/job/foreman-discovery-image-publish/8/console

If I get openstack credentials I can test this locally. Or is it the time to start building ISOs in koji? If we actually garbage collect ISO files carefully, this could be the way to build them.

ekohl:
@lzap I did have a look at it and we probably need to update the image name we use. If it can be built in koji, I wouldn't mind that either but I'm assuming that it's better to fix the job now to get the release out. AFAIK you have jenkins slave access. I'd recommend logging into one and taking the credentials from `~jenkins/.vagrant.d/Vagrantfile`

PR #708, opened 2018-06-25
convert rubygem-katello-nightly-release and hammer-cli-katello-nightl…

komidore64:
…y-release to groovy pipelines


komidore64:
theforeman/foreman-packaging#2689 is related to this one.

hammer-cli-katello won't build probably without both PRs.

komidore64:
I'm not particularly married to seeing variables with parameters. If people would rather, i could define groovy variables instead. lemme know.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
```diff
@@ -0,0 +1,31 @@
+pipeline {
```
ehelms: A nit, to capitalize RPM in the file name since its an acronym.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
```diff
@@ -0,0 +1,31 @@
...
+        timeout(time: 2, unit: 'HOURS')
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage ("Build Gem") {
+            steps {
+                dir('gem-build') {
+                    git(url: "https://github.com/${env.OWNER_REPO}", branch: env.BRANCH)
```
ehelms: Are these reference as `env.` or `params.` ?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/build/rubygem-katello-nightly-release.yaml
```diff
@@ -4,6 +4,20 @@
...
+          default: Katello/katello
+      - string:
+          name: BRANCH
+          default: master
+      - string:
+          name: GEMSPEC
+          default: katello.gemspec
+      - string:
+          name: PACKAGE_NAME
+          default: rubygem-katello
```
ehelms: The weird part about parameters is, you can change them per job invocation which seems odd for most of these to do.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
```diff
@@ -0,0 +1,31 @@
...
+                    archiveArtifacts(artifacts: '*.gem')
+                }
+            }
+        }
+
+        stage('Build RPM') {
+            steps {
+                dir('foreman-packaging') {
+                    git(url: 'https://github.com/theforeman/foreman-packaging', branch: 'rpm/develop')
+                    obal(action: "release", extraVars: ['build_package_tito_releaser_args': ["--arg jenkins_job=${env.JOB_NAME}", "--arg jenkins_job_id=${env.BUILD_ID}"]], packages: env.PACKAGE_NAME)
```
ekohl: Can we also set the releaser via variables here? It's what the current RPM build job does and I think that unless we have a specific `obal` command it's the quickest way to add more nightlies.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
komidore64: `env.OWNER_REPO` is accessing the `OWNER_REPO` environment variable.

komidore64: oh! `${params.OWNER_REPO}` is also a thing that would work! i didn't realize that!

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
komidore64: renamed to `rubygemRPMNightlyRelease.groovy`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRpmNightlyRelease.groovy
komidore64: It looks like obal may have an undocumented `build_package_tito_releasers` variable we can set. I will play with that a bit and get back to you.

komidore64: my bad, looks like it's just `releasers`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/rubygem-katello-nightly-release.groovy
```diff
@@ -0,0 +1,4 @@
+def owner_repo = 'Katello/katello'
```
ehelms: Would we be better off naming these files `rubygem-katello.groovy` since the data here is not specific to a nightly-release and could be re-used? Except I suppose maybe the branch.

ekohl: I'm still against the rubygem prefix because that's a RH-thing. This will be weird on packages that are also built on Debian. IMHO we should be using the source repo name rather than the result. Perhaps we should also be using the GH org but that's unlikely to be giving name conflicts. The branch could be a good thing though if we do start releasing "nightly" versions for released branches (1.18/3.7).

komidore64: I gave it the same name as this job to denote it's currently only related to the `rubygem-katello-nightly-release` job. But I do agree that if we decide to reuse it in the future, it could be renamed to something more generic.

ehelms: Last question, would it be cleaner then to just put these into the yaml itself?

ehelms: Eh, I retract that question as I don't think I like it after mulling.

ehelms:
To @ekohl point, and I could go either way, we have an inconsistency now. We have `rubygem-katello-nightly-release` and `hammer-cli-katello-nightly-release` with the former clashing with `katello-nightly-release` the pipeline. Sounds like we should rename the latter and then decide if we are using some sort of clarification for the other jobs?

ekohl:
https://github.com/theforeman/foreman-infra/pull/702 :)

ehelms:
@komidore64 With convetions merged, I think we can update this now?

ekohl:
I realized that the big difference between this hammer plugin and others is that this one has no foremandist tag in the release. That means we have to tag it instead of cherry picking it. This can break if we switch Ruby stacks.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy
```diff
@@ -0,0 +1,31 @@
...
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage ("Build Gem") {
+            steps {
+                dir('gem-build') {
+                    git(url: "https://github.com/${owner_repo}", branch: branch)
+                    sh("gem build ${gemspec}")
+                    archiveArtifacts(artifacts: '*.gem')
```
ekohl: It's important we clean any old gems or we can archive multiple gems this way. Do I miss the cleanup step?

ehelms: Agreed, we should have a post {} section to deleteDir at the end

komidore64:
rebased

komidore64:
@ehelms @ekohl changed job names to new conventions

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy
komidore64: can do!

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy
```diff
@@ -0,0 +1,31 @@
...
+                }
+            }
+        }
+
+        stage('Build RPM') {
+            steps {
+                dir('foreman-packaging') {
+                    git(url: 'https://github.com/theforeman/foreman-packaging', branch: 'rpm/develop')
+                    obal(action: "release", extraVars: ['build_package_tito_releaser_args': ["--arg jenkins_job=${env.JOB_NAME}", "--arg jenkins_job_id=${env.BUILD_ID}"]], packages: package_name)
```
ekohl: One problem is that we can't really rely on `releasers` to be set to the right one but then there are also 3 jenkins releasers. Perhaps https://github.com/theforeman/foreman-packaging/pull/2732 can help but I really think we need a obal nightly action.

komidore64:
@ekohl i've moved nightly args and releasers into the job. next is to remove them from package_manifest in fpkg.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy
```diff
@@ -0,0 +1,57 @@
...
+ * package_name
+ */
+
+def obalExtraVars = [
+    'build_package_tito_releaser_args': [
+        "--arg jenkins_job=${env.JOB_NAME}",
+        "--arg jenkins_job_id=${env.BUILD_ID}"
+    ],
+    'releasers': [
+        'koji-katello-jenkins'
```
ekohl: This could be good enough for the first version since we're only on katello anyway, but perhaps this should also be a variable?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-katello-master-release.groovy
```diff
@@ -0,0 +1,4 @@
+def owner_repo = 'Katello/hammer-cli-katello'
```
ekohl: Not needed right now, but we've renamed various repos (foreman_docker) to match their gem name. Probably good to have a similar look at hammer git repos. It'd allow us to introduce a `gem_name` variable and derive `gemspec` and `package_name` based on it. Again, future improvement :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy
komidore64: yeah it could be moved into a variable in the future

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-katello-master-release.groovy
komidore64: yeah, hammer-cli-* bits could benefit from having their repos named to be the same as their packages/gems.

PR #709, opened 2018-06-25
Add close inactive task to crontab

tbrisker:


puppet/modules/prprocessor/manifests/init.pp
```diff
@@ -65,6 +64,14 @@
     require     => Package[$packages],
   }
 
+  cron { 'close inactive':
```
ehelms: General question, cron is quite untrackable. Feels like whenever I see them we should be preferring something like a Jenkins job that can be monitored for failures. Thoughts?

tbrisker:
@ekohl addressed your comments; regarding monitoring - i agree this isn't ideal but hopefully this script specifically woulnd't need much monitoring, it was already running with cron on the old openshift infra.

ehelms:
@ekohl Harder because of SSH access requirement?

ekohl:
@ehelms yes. Because we have community slaves that means they have pretty much full access to our infra. The PR processor has some API tokens with a fair amount of access. While it's true that they already have access via other ways, we should be closing those ways down rather than opening up more ways.

ehelms:
I thought we:

 a) restricted ssh'ing to only be run on the master
 b) locked down what scripts can be run on the remote boxes via ssh for the jenkins user

puppet/modules/prprocessor/manifests/init.pp
```diff
@@ -65,6 +64,15 @@
...
+    command     => "cd ${app_root} && bundle exec ruby scripts/close_inactive.rb",
```
ohadlevy: shouldn't it be bundle exec script/close_inactive.rb ?

puppet/modules/prprocessor/manifests/init.pp
```diff
@@ -65,6 +64,15 @@
...
+    user        => $username,
+    environment => ["HOME=${app_root}"],
+    hour        => 1,
+    minute      => 23,
+    require     => Exec['install prprocessor']
```
ohadlevy: dont you need the ENV=GITHUB_OAUTH_TOKEN too?

ekohl:
@tbrisker could you have a look at the comments?

puppet/modules/prprocessor/manifests/init.pp
tbrisker: I'm not sure, where would i get the token from?

ekohl: You could do `GITHUB_OAUTH_TOKEN='${github_oauth_token}' bundle exec`. I think cron files per user can be considered secure since you can't read other users' crontabs.

tbrisker:
Pushed an updated version that hopefully addresses all the comments. IIRC, there were some issues with the script itself that @ohadlevy was looking into, i'm not sure if they're fixed yet or not - if not we should wait to fix those first.

PR #712, opened 2018-07-02
Allow db:drop to fail

ekohl:
When the database doesn't exist, the rake task will fail. Because that
means we've reached the desired state, this is actually not a problem
for us.

cc @tbrisker


ekohl:
Should fix https://ci.theforeman.org/job/test_1_17_stable/40/database=postgresql,ruby=2.4,slave=fast/console

tbrisker:
Thanks @ekohl !

PR #714, opened 2018-07-09
Verify version and release for RPM PR test

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -17,3 +17,67 @@ def find_changed_packages(diff_range) {
...
+
+            for (int i = 0; i < packages.size(); i++) {
+                package_name = packages[i]
+                version_status = 'failed'
+                release_status = 'failed'
+
+                new_version = query_rpmspec("packages/${package_name}/*.spec", '%{VERSION}')
+                new_release = query_rpmspec("packages/${package_name}/*.spec", '%{RELEASE}')
+
+                sh "git checkout origin/${env.gitlabTargetBranch}"
```
ekohl: Probably needs to use the GHPRB var here.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -17,3 +17,67 @@ def find_changed_packages(diff_range) {
...
+                      script: "rpmdev-vercmp ${old_release} ${new_release}",
+                      returnStatus: true
+                    )
+
+                    compare_new_to_one = sh(
+                      script: "rpmdev-vercmp 1 ${new_release}",
+                      returnStatus: true
+                    )
+
+                    if (compare_version != VERCMP_EQUAL && (compare_new_to_one == VERCMP_OLDER || compare_new_to_one == VERCMP_EQUAL)) {
```
ekohl: Edge case, but what if the version actually goes backwards? This can happen if a PR was merged in the mean time and an incorrect rebase.

ehelms: Agreed, edge case not sure worth taking into account.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
```diff
@@ -53,6 +56,63 @@ pipeline {
...
+                              script: "rpmdev-vercmp ${old_release} ${new_release}",
+                              returnStatus: true
+                            )
+
+                            compare_new_to_one = sh(
+                              script: "rpmdev-vercmp 1 ${new_release}",
+                              returnStatus: true
+                            )
+
+                            if (compare_version != VERCMP_EQUAL && (compare_new_to_one == VERCMP_OLDER || compare_new_to_one == VERCMP_EQUAL)) {
```
ekohl: The `compare_new_to_one == VERCMP_OLDER` is to catch the release is set to 0.1?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
```diff
@@ -1,5 +1,8 @@
 def packages_to_build
 def packages = [:]
+def VERCMP_NEWER = 12
```
ekohl: Should these be defined in `rpm_packaging.groovy` so every job that uses this can use them?

ehelms: If this functionality were re-factored out I'd lean towards yes, but for now keeping them close to the code that uses them.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
ehelms: I believe so.

PR #715, opened 2018-07-09
Centralize the groovy library files into theforeman.org

zjhuntin:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -1,32 +1,20 @@
...
-        if (config.extraVars) {
-            extraVars = defaultVars + config.extraVars
-        } else {
-            extraVars = defaultVars
-        }
+    if (options) {
+        command += options
+    }
 
-        ansiblePlaybook(
```
ekohl: any reason this can no longer be used?

zjhuntin: No, but this way feels easier and faster way of using this instead of the class command.

ekohl: Faster in which way? It feels like this function is now reinventing this class command.

zjhuntin: I made a poor assumption, it's not for any of those reasons. It turns out that ci.theforeman.org has the plugin that allows `ansiblePlaybook` to be run. ci.cenos.org does not have that plugin and this was the way @ehelms got around not having it.

PR #718, opened 2018-07-11
Make sure URI queries have json defined before checking other conditionals

zjhuntin:


puppet/modules/jenkins_job_builder/files/centos.org/ansible/jenkins_job.yml
```diff
@@ -53,7 +53,7 @@
         password: "{{ jenkins_password }}"
         url: "{{ start_job_result.location }}api/json"
       register: poll_result
-      until: poll_result.json.executable.number | default(0) | int > 0
+      until: poll_result.json is defined and (poll_result.json.executable.number | default(0) | int > 0)
```
komidore64: optional:

it would look more uniform to surround all conditional sections with parentheses.

`(poll_result.json is defined) and (poll_result.json.executable.number | default(0) | int > 0)`

PR #720, opened 2018-07-13
Add tfm-ror52 release job

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+        disableConcurrentBuilds()
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
+
+                checkout([
```
zjhuntin: Not saying it should change but what is the advantage of using this over git url: <git_url>, branch <branch_name> ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+                script {
+
+                    if (fileExists('commit')) {
+                        commit = readFile(file: 'commit').trim()
+
+                        if (commit) {
+                            packages_to_build = find_changed_packages("${commit}..HEAD")
+                        }
+                    } else {
+                        packages_to_build = 'all'
```
zjhuntin: Why build all in this case? Wouldn't we want this to be something more along the lines of none? Or is it more that Obal will safely ignore the packages that are already built, but will try to build anything that still needs building?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror52-release.yaml
```diff
@@ -0,0 +1,16 @@
...
+- job:
+    name: tfm-ror52-release
+    project-type: pipeline
+    properties:
+      - github:
+          url: https://github.com/theforeman/rails-packaging
+    triggers:
+        - github
+        - pollscm:
+            cron: "* * * * *"
```
zjhuntin: This just means it's polling the repo to see if there is a new commit that was merged every minute?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
ehelms: This form actually restricts this to run only on the branch specified via the internals.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
ehelms: Yea, it would try to fall back to build everything which has a check for existence in Koji.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror52-release.yaml
ehelms: ACK

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
zjhuntin: Ahh okay, good with me.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror52-release.yaml
ehelms: I thought as I understood things in the docs, this is required for a trigger on merge. I could be wrong though and remove it to try it out.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+        ansiColor('xterm')
+        buildDiscarder(logRotator(numToKeepStr: '15'))
+        disableConcurrentBuilds()
+    }
+
+    stages {
+        stage('Clone Packaging') {
+            steps {
+
+                deleteDir()
```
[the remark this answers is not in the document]
ehelms: Define "clean the git repo" ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+                    action: 'release',
+                    packages: packages_to_build
+                )
+
+            }
+        }
+    }
+
+    post {
+        always {
```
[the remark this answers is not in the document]
ehelms: Good catch, I did mean this to only be on success.

ehelms:
Updated with recent comments.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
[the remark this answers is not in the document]
ekohl: https://github.com/theforeman/foreman-infra/blob/9948b6a1b7488351b706852c5eaae933ea1bace7/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy#L62-L67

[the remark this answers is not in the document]
ehelms: I suppose we could, however, if there are lingering artifacts not in git that won't cover cleaning those out?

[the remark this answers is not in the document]
ekohl: That's exactly what it's cleaning out in tarballsRelease. There we had some extra tarballs remaining from the previous run and adding that fixed the issue. It's equivalent to `git clean` (IIRC that's what it calls under the hood).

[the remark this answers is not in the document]
ehelms: So turning it around, what's the advantage to git clean vs. straight deleteDir to cleanup everything?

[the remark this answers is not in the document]
evgeni: it does a `git clean -fdx`, which wipes everything that is either unknown or changed, so that should be sufficient

https://github.com/jenkinsci/git-client-plugin/blob/027d2a0b2f8726d2f08fb747e219e497b06e856c/src/main/java/org/jenkinsci/plugins/gitclient/CliGitAPIImpl.java#L804-L806

given this would save time (the clone is not thrown away, just updated cleanly), I'd +1 that change

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+                        [url: 'https://github.com/theforeman/rails-packaging']
+                    ]
+                ])
+
+            }
+        }
+        stage('Find packages') {
+            steps {
+
+                copyArtifacts(projectName: env.JOB_NAME, optional: true)
```
evgeni: Should this have

```
buildSelector {
                latestSuccessful(true)
            }
```

or similar to get the commit of the last successful job?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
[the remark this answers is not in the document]
ekohl: A git clone can take a long time on some repositories. Cleaning + updating might be a lot faster.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+                    packages: packages_to_build
+                )
+
+            }
+        }
+    }
+
+    post {
+        success {
+            sh "git rev-parse HEAD > commit"
```
evgeni: so the idea here is that after we run a build, we write the sha1 of `HEAD` to `commit` and archive that file, and the next run will fetch it again via `copyArtifacts` and know what changed? deserves a comment ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+                            packages_to_build = find_changed_packages("${commit}..HEAD")
+                        }
+                    } else {
+                        packages_to_build = 'all'
+                    }
+                }
+            }
+        }
+        stage('Release Build Packages') {
+            steps {
```
evgeni: I think `find_changed_packages` can return `[]`, would this step then need a condition not to run? like we do here?

https://github.com/theforeman/foreman-infra/blob/5656e2708758b9f7761b8839d6725ea49ece841c/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/tfmROR51.groovy#L33-L35

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror52-release.yaml
```diff
@@ -0,0 +1,15 @@
+---
+- job:
+    name: tfm-ror52-release
+    project-type: pipeline
+    concurrent: false
```
evgeni: duplicate with the `        disableConcurrentBuilds()` above?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,67 @@
...
+
+                copyArtifacts(projectName: env.JOB_NAME, optional: true)
+
+                script {
+
+                    if (fileExists('commit')) {
+                        commit = readFile(file: 'commit').trim()
+
+                        if (commit) {
```
evgeni: in which case is this false? when the file exists, but empty? should we build `all` then?

ekohl:
> while I still think the github plugin should just export the "previous commit" information when a merge happens, it does not as of today.

Actually I like this approach more because it can handle multiple merges and can resume on its own.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
ehelms: By default, copyArtifacts will look for `lastSuccessfulBuild`

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
ehelms: That's the idea, can comment

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
evgeni: oh, ok.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/tfm-ror52-release.yaml
ehelms: This version of it is more robust than using in the pipeline so I removed that reference.

ehelms:
@evgeni @ekohl Updated and addressed comments

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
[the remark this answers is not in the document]
ekohl: If you've added `CleanCheckout` then `deleteDir()` should no longer be needed.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,68 @@
...
+                    packages: packages_to_build
+                )
+
+            }
+        }
+    }
+
+    post {
+        success {
+            # Save current commit hash for use by the next release run
```
evgeni: I don't think you can use `#` for comments, only `//` and `/* … */`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tfmROR52.groovy
```diff
@@ -0,0 +1,68 @@
...
+            sh "git rev-parse HEAD > commit"
+            archive 'commit'
```
komidore64: i believe this should be `archiveArtifacts`.

> As of Jenkins 2.x, this step is deprecated in favor of the more configurable archiveArtifacts.

ehelms:
@komidore64 Updated based on your comment

PR #721, opened 2018-07-14
Split webpack to separate Katello job

ehelms:
Exploring the concept of splitting out to separate jobs to allow faster iteration on UI code that does not need to run full Ruby test suite.

@waldenraines @jlsherrill

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-webpack-pr-test.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-webpack-pr-test.yaml
@@ -1,5 +1,5 @@
 - job:
-    name: katello-pr-test
+    name: katello-webpack-pr-test
     project-type: pipeline
     concurrent: true
     properties:
@@ -7,15 +7,16 @@
           url: https://github.com/Katello/katello
       - tfm-pull-request-build-discarder
     triggers:
-      - github_pr_pipeline:
-          context: 'katello'
-          excluded_regions:
+      - github_pr_katello:
+          context: 'katello/webpack'
+          included_regions:
             - ^webpack\/.+
+            - ^engines\/bastion_katello\/.+
             - ^app\/assets\/.+
             - package.json
     dsl:
       !include-raw:
-        - pipelines/test/testKatello.groovy
+        - pipelines/test/testKatelloWebpack.groovy
         - pipelines/lib/rvm.groovy
         - pipelines/lib/git.groovy
         - pipelines/lib/foreman.groovy
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
@@ -66,16 +66,6 @@ pipeline {
                         }
                     }
                 }
-                stage('react-ui') {
-                    when {
-                        expression { fileExists('package.json') }
-                    }
-                    steps {
-                        sh "npm install npm"
-                        sh "node_modules/.bin/npm install"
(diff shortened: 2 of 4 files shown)


waldenraines:
I don't know enough about foreman-infa to approve this but I fully support this initiative, thanks for the quick turnaround on this @ehelms!

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-pr-test.yaml
```diff
@@ -9,6 +9,10 @@
     triggers:
       - github_pr_pipeline:
           context: 'katello'
+          excluded_regions:
+            - ^webpack\/.+
+            - ^app\/assets\/.+
+            - package.json
```
ehelms: @jlsherrill @waldenraines -- This represents folders that if they are the only things changed will not trigger a build. In this case, the idea being these do not trigger the re-designed katello-pr-test for ruby tests. Can you check accuracy?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-webpack-pr-test.yaml
```diff
@@ -0,0 +1,22 @@
...
+          url: https://github.com/Katello/katello
+      - tfm-pull-request-build-discarder
+    triggers:
+      - github_pr_katello:
+          context: 'katello/webpack'
+          included_regions:
+            - ^webpack\/.+
+            - ^engines\/bastion_katello\/.+
+            - ^app\/assets\/.+
+            - package.json
```
ehelms: @jlsherrill @waldenraines -- This represents folders that if they are the only things changed will trigger a build. In this case, the idea being these trigger the new katello-webpack-pr-test for ruby tests. Can you check accuracy?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/tests/katello-pr-test.yaml
jlsherrill: One question, with the recent change we made that requires the katello job to run, will this cause unmergable prs?

ehelms: That's a fair question, testing this out it appears that the way the included/excluded work is by not triggering a job which would break that required logic. Which kinda sucks. The alternatives are:

 1) Single katello-pr-test job that doe sits own checking for git changes and runs subsets of test tasks based on those changes thus reporting status still just on different aspects
 2) Split pr jobs that do their own checking for git changes and either report success if nothing to run or run the tests and let them dictate status

Option #1 would mean things stay as is, with [test katello] and it runs whatever is needed each time. Option #2 means more status granularity and ability to individually re-run a status.

ehelms: I don't think we do. @waldenraines Can you comment towards the nature of the webpack Ui tests?

waldenraines: We don't have any API integration tests, all of the API calls are mocked.  If you changed the API without updating the UI, no tests would fail.  This is likely something we should look at adding in the future.  At the moment we're just relying on the dev to realize that they also need to update the UI if they change the API.

jlsherrill: i guess I have a slight preference with 1) as it would potentially reduce some re-test time if there were updates in both places.

ekohl:
Closing this for now due to its age and merge conflicts.

PR #725, opened 2018-07-18
Update release_packages job for 1.18+ style

ekohl:
This breaks the 1.17 release but it's hard to support both. With 1.19 being branched we are more likely to use this style than the old one.


ekohl:
cc @xprazak2

xprazak2:
LGTM

PR #728, opened 2018-07-19
Add pcreech ssh-key

pcreech:


zjhuntin:
ACK

ehelms:
@pcreech Can you send a discourse thread with this request?

pcreech:
https://community.theforeman.org/t/request-to-add-pcreech-to-foreman-infra-ssh-keys/10523

evgeni:
the thread got 6 likes in a week, so we should be good?

zjhuntin:
I think we are good to go here.

ekohl:
Thanks, will add the user in Foreman and discuss the password offline.

PR #730, opened 2018-07-23
Test releases on PostgreSQL

ekohl:
We've always tested releases with PostgreSQL but in c57d43f0af47cecc867bf6d65b1ce8ae8c82cb6c this was changed to MySQL.  Probably unintentional.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
@@ -17,7 +17,7 @@
           repo_environment: staging
           pl_puppet: ''
           run_hammer_tests: false
-          db_type: mysql
+          db_type: ''
           umask: ''
           expected_version: '${version}'
     axes:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_test.yaml
```diff
@@ -17,7 +17,7 @@
           repo_environment: staging
           pl_puppet: ''
           run_hammer_tests: false
-          db_type: mysql
+          db_type: ''
```
zjhuntin: We don't want to specify it as postgresql?

ekohl: Maybe the message wasn't totally correct, but I'd prefer to test with the defaults. This happens to be postgres. In a release test I think we want to test with defaults where possible.

ekohl: I think @komidore64 told us that we need to specify every parameter. That's why `pl_puppet` is there as well while being empty.

PR #732, opened 2018-07-24
Ensure Jenkins user part of docker group

ehelms:
To enable Jenkins slave to run docker

### DIFF
--- a/puppet/modules/slave/manifests/docker.pp
+++ b/puppet/modules/slave/manifests/docker.pp
@@ -17,4 +17,10 @@ class slave::docker {
       default  => 'docker',
     };
   }
+
+  group { 'docker':
+    ensure => 'present',
+  }
+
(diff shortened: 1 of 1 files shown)


puppet/modules/slave/manifests/docker.pp
```diff
@@ -17,4 +17,10 @@
       default  => 'docker',
     };
   }
+
+  group { 'docker':
+    ensure => 'present',
+  }
+
+  User<|title == jenkins|>{groups +> ['docker']}
```
zjhuntin: Does this add the jenkins user to the docker group if it's already not in it?

ekohl: It should. You may need to quote `jenkins` here.

ehelms:
Updated to put the jenkins user in quotes

PR #733, opened 2018-07-25
Add Katello 3.8 pipeline

ehelms:


puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/katello/3.8.groovy
```diff
@@ -0,0 +1,42 @@
+pipeline {
+    agent { label 'foreman' }
+
+    stages {
+        stage('Setup Environment') {
```
ekohl: Now that we're copying this between pipelines, can we introduce a helper for this?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.8.groovy
```diff
@@ -0,0 +1,109 @@
...
+        stage('Mash Koji Repositories') {
+            agent { label 'sshkey' }
+
+            steps {
+
+                sh "ssh -o 'BatchMode yes' root@koji.katello.org katello-mash-split-3.8.py"
+
+            }
+        }
+        stage('Client Repoclosure') {
```
ekohl: I thought we also had el5 as a client

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/katello/3.8.groovy
ehelms: @zjhuntin and I discussed this and he is going to look at re-factoring but outside the scope of this change since that would hol dup this needed pipeline.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.8.groovy
ehelms: We do, but it's repoclosure is tricky due to it's age and so even if oyu look at nightly we dropped el5 repoclosure.

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/katello/3.8.groovy
ekohl: Would https://github.com/theforeman/foreman-infra/pull/736 be a better place?

ehelms: Yup -- thats where he plans to test re-factoring out.

PR #739, opened 2018-07-26
Ensure cron mail ends up at sysadmins

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #741, opened 2018-07-31
Allow version and release check to work more broadly

ehelms:
Find the spec file explicitly rather than relying on wildcard paths
that don't work for all packaging release workflows.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
@@ -67,14 +67,15 @@ pipeline {
                         def index = i
                         packages[packages_to_build[index]] = {
                             package_name = packages_to_build[index]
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
```diff
@@ -67,14 +67,15 @@ pipeline {
                         def index = i
                         packages[packages_to_build[index]] = {
                             package_name = packages_to_build[index]
+                            spec_path = sh(returnStdout: true, script: "find -name \"${package_name}\\.spec\"").trim()
 
-                            new_version = query_rpmspec("packages/**/${package_name}/*.spec", '%{VERSION}')
-                            new_release = query_rpmspec("packages/**/${package_name}/*.spec", '%{RELEASE}')
+                            new_version = query_rpmspec("${spec_path}", '%{VERSION}')
```
ekohl: Do you need to interpolate the string here or could you simply us `spec_path` here?

ehelms:
Fair point @ekohl -- dropped the interpolation

PR #744, opened 2018-08-01
Fix folderChanged logic

ehelms:
This was breaking after an update to ci.centos.org infrastructure which makes me think the Jenkins upgrade they did changed slightly how the groovy worked with the empty check. This makes it more robust against potential usage of empty on different data types.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/folderChanged.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/folderChanged.groovy
@@ -5,7 +5,8 @@ def folderChanged(folders) {
     files_changed = readFile('files_changed').split('\n')
 
     for (int i = 0; i < folders.size(); i++) {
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/folderChanged.groovy
```diff
@@ -5,7 +5,8 @@ def folderChanged(folders) {
     files_changed = readFile('files_changed').split('\n')
 
     for (int i = 0; i < folders.size(); i++) {
-        changed = changed || !files_changed.findAll { it =~ folders[i] }.empty
```
zjhuntin: does findAll {}.empty not work not return false if 0 or found?

ehelms: It seems to be returning an array of values. For example:

```
[false, false, false, false]
```

PR #745, opened 2018-08-02
Add foreman-packaging release job

ehelms:


ekohl:
If a release fails (for example because the build already exists) then this job won't recover from it and get stuck in a loop. How do we recover from this?

ehelms:
We can turn on diff package in obal to avoid the build already exists error. As for other errors, if we can't enumerate them then I am not sure they are worth consideration.

ekohl:
If a build failure is introduced (i.e. something was merged with errors) then it blocks the release pipeline. IMHO that's not an issue per se.

Another I can think of is addition and removal of packages. AFAIK `obal release` doesn't change koji tags so release will fail. Removal of packages shouldn't trigger an `obal release` either.

We probably need to update `foreman-packaging`'s README as well.

Other than that I think I very much like this change. I do wonder if we could update the PRs with a link to the release job but that's optional.

ehelms:
Looking at bit further, setting status is gonna be slightly harder than I originally thought due to how handling merges is for Github today. For one, this strategy does not specifically call out a single merge because in the event of a failure, you could be building a backlog of items.

My proposal to start would be an IRC notification that the job failed. @ekohl thoughts?

ekohl:
We already have some jobs that report to IRC but the experience is that failing jobs often still go unnoticed. I don't mind IRC notifications but I'm leaning to email notifications as well.

ehelms:
I added an email notification to the job's post section. Currently it just emails me to reduce spam. My belief is that this job will need to be a bit experimental at first. Implemented, but not the full source of truth as we get used to it and figure out edge cases surrounding it. So we should in my view (I applied this same workflow to tfm-ror52 job):

 1) Merge a packaging PR
 2) Go to this job, monitor what it does
 3) Make changes as it errors or does anything odd
 4) Release locally if we need to

ehelms:
Thanks for review @ekohl -- this will require https://github.com/theforeman/foreman-packaging/pull/2834 before it can be truly usable

PR #753, opened 2018-08-07
Expect 1.20.0-develop in deb systest

tbrisker:


mmoll:
merged, thanks @tbrisker!

PR #754, opened 2018-08-07
Set the Redmine API key for close_inactive.rb

ekohl:
The script wants to close issues so it needs the API key.


mmoll:
merged, bedankt @ekohl!

PR #755, opened 2018-08-07
Clean up inactive users

tbrisker:


ekohl:
The whole block is no longer needed because we pass it in. It should be rewritten to just have the create_resources block

tbrisker:
Updated to remove the whole block.

ekohl:
Can you update the parameter to `Hash[String, Hash] $users = {}` as well?

tbrisker:
you got it

PR #760, opened 2018-08-08
Switch yaml to debug for Ansible on duffy

ehelms:
Ansible 2.4 on Duffy does not support yaml output yet


zjhuntin:
ACK

zjhuntin:
What needs to happen to duffy in order for this to work?

zjhuntin:
Never mind, Ansible 2.6, which we can wait till becomes more readily available on Duffy.

PR #761, opened 2018-08-09
Do not add deb prefix automatically

xprazak2:
PRs opened against debians require prefix,
so we do not need to add it second time.


xprazak2:
To avoid https://ci.theforeman.org/job/packaging_build_deb_coreproject/5742/console

PR #762, opened 2018-08-09
submit koji status as koji/<package> not koji/<taskid>

evgeni:

PR #763, opened 2018-08-10
only list the koji task id in the description

evgeni:
the package name is already in the "context" and this makes the description better readable

requested by @ekohl in #762


ekohl:
Thanks!

PR #771, opened 2018-08-15
add "true" for db:drop

mmoll:


mmoll:
I guess because it was not needed back then ;)

PR #773, opened 2018-08-15
Add Foreman client mash and repoclosure to Foreman nightly

ehelms:


ehelms:
@ekohl I debated internally whether to make the client repositories their own pipeline independent of the core push. While not currently present, the aspects I wondered about were adding tests of client packages and if that should be independent of the release pipeline for core.

ekohl:
I think it'll depend on if we move hammer over as well but I think that does make sense. Client repos should be usable without the core repository so should be promoteable independently.

ehelms:
I've updated this to be it's own client release job.

ehelms:
This will need https://github.com/theforeman/foreman-packaging/pull/2821 before it can hope to succeed

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanClientRelease.groovy
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+    }
+}
+
+void repoclosure(repo, dist) {
+    obal(
+        action: 'repoclosure',
+        packages: "${repo}-repoclosure-${dist}"
+    )
```
ehelms: Switching to try out using our obal repoclosure config and functionality.

PR #774, opened 2018-08-15
Add update-job for ci.centos.org jobs

zjhuntin:
This job will update the Foreman jobs for ci.centos.org every fifteen minutes.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
@@ -1,0 +1,40 @@
+pipeline {
+    agent { label 'fast' }
+
+    stages {
+        stage('Setup workspace') {
+            steps {
+                checkout([
+                    $class : 'GitSCM',
+                    branches : [[name: 'update_jobs']],
+                    extensions: [[$class: 'CleanCheckout']],
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/theforeman/foreman-infra.git']
+                    ]
+                ])
+
+                virtEnv('./ci', 'pip install jenkins-job-builder')
+            }
+        }
+
+        stage('Update ci.centos.org jobs') {
+            steps {
+                withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'jenkins-centos', passwordVariable: 'PASSWORD', usernameVariable: 'USERNAME']]) {
+                    virtEnv('./ci', "cd ./centos.org && jenkins-jobs --conf ./centos_jenkins.ini --user '${env.USERNAME}' --password '${env.PASSWORD}' update -r ./jobs")
+                }
+            }
+        }
+    }
+
+    post {
+        always {
+            script {
+                if(fileExists('ci/update_jobs')) {
+                    dir('ci/update_jobs') {
+                        deleteDir()
+                    }
+                }
+            }
+        }
+    }
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
@@ -1,0 +1,8 @@
+- job:
+    name: foreman-infra-jenkins-jobs-update
+    project-type: 'pipeline'
+    sandbox: true
+    dsl:
+      !include-raw:
+        - pipelines/infra/updateJobs.groovy
+        - pipelines/lib/virtualEnv.groovy
--- a/puppet/modules/jenkins_job_builder/files/centos.org/centos_jenkins.ini
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/centos_jenkins.ini
@@ -1,0 +1,10 @@
+[job_builder]
+ignore_cache=True
+keep_descriptions=False
+include_path=.
+recursive=True
+allow_duplicates=False
+
+[jenkins]
+url=https://ci.centos.org
+query_plugins_info=False
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/virtualEnv.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/virtualEnv.groovy
@@ -1,0 +1,13 @@
+def virtEnv(path, command) {
+    dir(path) {
+        if(!fileExists('venv')) {
+            sh "virtualenv venv"
+        }
+
+        sh """
+        source venv/bin/activate
+        ${command}
+        deactivate
+        """
+    }
+}


puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-update-jobs.yml
```diff
@@ -0,0 +1,10 @@
+---
+- job:
+    name: 'foreman-update-jobs'
+    project-type: 'pipeline'
+    sandbox: true
+    triggers:
+      - timed: '*/15 * * * *'
```
ehelms: Can this be triggered on PR merge to foreman-infra?

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-update-jobs.yml
```diff
@@ -0,0 +1,10 @@
+---
+- job:
+    name: 'foreman-update-jobs'
```
ehelms: Thinking something more descriptive here perhaps, in the sense of adding knowledge that it's related to foreman-infra or jenkins-jobs

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,20 @@
+pipeline {
+    agent { label 'foreman' }
+
+    stages {
+        stage('Update Jenkins Jobs') {
+            steps {
+                deleteDir()
+                git url: 'https://github.com/theforeman/foreman-infra.git'
```
ehelms: A few folks have indicated that the strategy to follow is more efficient than a deleteDir and re-clone:

```
                checkout([
                    $class : 'GitSCM',
                    branches : [[name: '*/tfm-ror52']],
                    extensions: [[$class: 'CleanCheckout']],
                    userRemoteConfigs: [
                        [url: 'https://github.com/theforeman/rails-packaging']
                    ]
                ])
```

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,20 @@
...
+                dir("${env.WORKSPACE}/ci/centos.org") {
+                    sh """
+                    virtualenv update_jobs
+                    source update_jobs/bin/activate
+                    pip install jenkins-job-builder
+                    jenkins-jobs --conf ~/jenkins_jobs.ini update --delete-old -r ./jobs
+                    """
```
ehelms: A post section that cleans up the virtualenv would be nice.

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,20 @@
+pipeline {
+    agent { label 'foreman' }
+
+    stages {
+        stage('Update Jenkins Jobs') {
+            steps {
+                deleteDir()
+                git url: 'https://github.com/theforeman/foreman-infra.git'
+                dir("${env.WORKSPACE}/ci/centos.org") {
```
ehelms: Is the `env.WORKSPACE` required since oyu should be at the root of the project?

zjhuntin:
@ehelms Updated, working on figuring out how we could trigger the job on merge to foreman-infra.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-infra-update-jenkins-jobs.yml
```diff
@@ -0,0 +1,10 @@
+---
+- job:
+    name: 'foreman-infra-update-jenkins-jobs'
```
ehelms: I guess this should be slightly reversed, we've tended to put the action at the end and the specifc -> generic at the front

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,39 @@
+pipeline {
+    agent { label 'foreman' }
+
+    stages {
+        stage('Update Jenkins Jobs') {
```
ehelms: Some of my comments got lost in the restructure but I think splitting this into two stages would be nice.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,39 @@
...
+                checkout([
+                    $class : 'GitSCM',
+                    branches : [[name: 'master']],
+                    extensions: [[$class: 'CleanCheckout']],
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/theforeman/foreman-infra.git']
+                    ]
+                ])
+
+                dir("./ci") {
```
ehelms: Should this be specific to the `centos.org` folder?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/updateJobs.groovy
```diff
@@ -0,0 +1,39 @@
...
+                    sh """
+                    virtualenv update_jobs
+                    source update_jobs/bin/activate
+                    pip install jenkins-job-builder
+                    jenkins-jobs --conf ~/jenkins_jobs.ini update --delete-old -r .
```
ehelms: Given shift in idea, how is this being handled now? The credential side of things.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-infra-update-jenkins-jobs.yml
```diff
@@ -0,0 +1,10 @@
+---
+- job:
+    name: 'foreman-infra-update-jenkins-jobs'
```
ehelms: I think this name would need to be something more like `foreman-infra-jenkins-jobs-update` to follow our naming scheme

zjhuntin:
@ekohl @ehelms and I discussed that and he made a good point in that if we want we can later come back to this and add updating ci.theforeman.org with this job. However there are enough other factors involved with updating ci.theforeman.org that it becomes a much more involved PR. For now I think this one is doing everything it's meant to. I will however update a card we have related to updating ci.theforeman.org via Jenkins as well.

ekohl:
On the other hand, if you introduce this now you have 2 systems touching these resources. That can make debugging very hard. I'd prefer to have 1 place to find things.

ehelms:
I am not following the two systems touching these resources. Currently no resource touches ci.centos.org jobs other than a developer. I could see the job name being a bit mis-leading; however, I do think @zjhuntin taking an iterative approach in smaller chunks is beneficial.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+                dir("./ci") {
+                    sh """
+                    virtualenv update_jobs
+                    source update_jobs/bin/activate
+                    pip install jenkins-job-builder
+                    deactivate
+                    """
+                }
+              }
```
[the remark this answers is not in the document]
zjhuntin: I'm unsure what you mean here.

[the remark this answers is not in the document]
ehelms: This chunk of code you moved down below.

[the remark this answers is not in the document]
zjhuntin: I'm unsure that the virtualenv will carry over to the other Shell command. Which is why I am activating and deactivating on my shell commands.

ekohl:
An alternative is to simply add a Puppet config to also run the updates on centos.org and keep both CI systems unified.

I strongly dislike us having 2 ways of updating the same tools. Experience tells us that we often don't end up fixing the technical debt we build up in a short term so IMHO we should either deploy both CI infras using Puppet or both using Jenkins. Jenkins does have the advantage of visibility.

ehelms:
That *is* the plan. However, I do not necessarily agree that *this* PR should have to do both. I think smaller, dedicated PRs are better than trying to re-do everything all at once. I have in the past tended to leave gaps of technical debt. We are planning to address theforeman.org case. One benefit of this method is it allows creating a job, testing that it works on one infra, working out the kinks and then coming back and updating for the other infra.

zjhuntin:
I like this PR as is.

> Experience tells us that we often don't end up fixing the technical debt we build up in a short term
I want to eventually move to using this job to update ci.theforeman.org. @ekohl I understand the fear of having two setup, and then we won't move to something in a reasonable amount of time. I want to update this, but I don't think it goes into this PR. This is something that I will get to next week though. I will add to this current job and create a stage that will update ci.theforeman.org. This will do what the puppet config is doing now, and then the PR can also remove the puppet config for ci.theforeman.org and we will be at one system.

That is how I would like to go moving forward.

puppet/modules/jenkins_job_builder/files/centos.org/centos_jenkins.ini.example
```diff
@@ -0,0 +1,10 @@
+[job_builder]
```
ekohl: Is there a reason this is named `.ini.example`? Why not just `.ini`?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        stage('Update ci.centos.org jobs') {
+            steps {
+                dir("./ci") {
+                    withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'jenkins-centos', passwordVariable: 'PASSWORD', usernameVariable: 'USERNAME']]) {
+                        sh """
+                        source update_jobs/bin/activate
+                        mv ./centos.org/centos_jenkins.ini.example ./centos.org/jenkins_jobs.ini
```
ekohl: If you're going to stick with `.example`, I think this is also setup code and not actual updating.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,54 @@
...
+                    branches : [[name: 'master']],
+                    extensions: [[$class: 'CleanCheckout']],
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/theforeman/foreman-infra.git']
+                    ]
+                ])
+
+                dir("./ci") {
+                    sh """
+                    virtualenv update_jobs
```
ekohl: I'd either use a self-describing name like `venv` or store it as constant.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,54 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    post {
+        always {
+            script {
+                if(fileExists('ci/update_jobs')) {
```
ekohl: Just a thought: alternative to deleting the venv I wonder if it makes sense to install an explicit JJB version. We'd know what JJB we get and it would be a bit more efficient.

puppet/modules/jenkins_job_builder/files/centos.org/centos_jenkins.ini.example
zjhuntin: I did it because we wanted to avoid having it before (see we have a .gitignore with .ini files). I think it doesn't make sense to keep it and we can instead just have a .ini file.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
zjhuntin: Would this be along the lines of having a virtual env intstalled on the jenkins system itself?

puppet/modules/jenkins_job_builder/files/centos.org/centos_jenkins.ini.example
ekohl: That is only `jenkins_job.ini` though I'm not sure why @ehelms added that in 9fd0ec78e9072f7fa3d1afd0340695d3f5c6191a.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
ekohl: No, but a workspace could reuse virtualenvs. On the other hand, it could be a source of problems as well so I'd be fine with leaving that optimization out.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        stage('Update ci.centos.org jobs') {
+            steps {
+                dir("./ci") {
+                    withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'jenkins-centos', passwordVariable: 'PASSWORD', usernameVariable: 'USERNAME']]) {
+                        sh """
+                        source update_jobs/bin/activate
+                        mv ./centos.org/centos_jenkins.ini.example ./centos.org/jenkins_jobs.ini
+                        jenkins-jobs --conf ./centos.org/jenkins_jobs.ini --user ${env.USERNAME} --password ${env.PASSWORD} update --delete-old -r ./centos.org
```
evgeni: this will include stuff from `centos.org/ansible` and die in flames. you want to only scan `centos.org/jobs` for jobs :)
also, for the relative paths in the job definitions to work, you need to execute this from the `centos.org` folder

zjhuntin: I tested it locally with the test command and nothing broke, but it's still probably a better practice.

evgeni: interesting, test is what exploded for me. but hey, now it's the way it did not explode ;) thanks

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,53 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    post {
+        always {
+            script {
+                if(fileExists('ci/venv')) {
```
evgeni: do we need this safeguard?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
```diff
@@ -0,0 +1,12 @@
...
+    name: 'foreman-infra-jenkins-jobs-update'
+    project-type: 'pipeline'
+    sandbox: true
+    scm:
+      - foreman-infra
+    triggers:
+      - pollscm: ''
+    dsl:
+      !include-raw:
+        pipelines/infra/updateJobs.groovy
```
evgeni: would that need to be

```
- pipelines/infra/updateJobs.groovy
```

or is jjb not too picky about lists?

zjhuntin: I've built it from the yaml, so I don't think it's too picky.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
zjhuntin: It's probably not strictly needed as if it errored out and the file isn't there it won't matter if it fails here. But I figured it wouldn't hurt to do it a "better" way.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,53 @@
...
+                    branches : [[name: 'master']],
+                    extensions: [[$class: 'CleanCheckout']],
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/theforeman/foreman-infra.git']
+                    ]
+                ])
+
+                dir("./ci") {
+                    sh """
+                    virtualenv venv
```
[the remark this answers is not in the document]
zjhuntin: Sure, I'll make a lib file for virtual envs.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
evgeni: `dir('blah')` creates `blah` if it does not exist, so it won't error :)

zjhuntin:
[test]

zjhuntin:
Requires: https://github.com/theforeman/foreman-infra/pull/801

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,40 @@
...
+
+    stages {
+        stage('Setup workspace') {
+            steps {
+                checkout([
+                    $class : 'GitSCM',
+                    branches : [[name: 'update_jobs']],
+                    extensions: [[$class: 'CleanCheckout']],
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/zjhuntin/foreman-infra.git']
```
ehelms: Noting this will need update before merging

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -0,0 +1,40 @@
...
+                    ]
+                ])
+
+                withVenv('./ci', 'pip install jenkins-job-builder')
```
ehelms: Maybe `VirtEnv` ? I worry that `Venv` won't be obvious when reading down the road.

chris1984:
[test]

zjhuntin:
Okay, I think this is ready to go.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
```diff
@@ -0,0 +1,10 @@
+- job:
+    name: foreman-infra-jenkins-jobs-update
+    project-type: 'pipeline'
+    sandbox: true
+    scm:
```
ehelms: Is this needed if cloning inside the pipeline?

zjhuntin: Probably not, went ahead and removed.

PR #776, opened 2018-08-16
Add a clone_foreman_infra function

ehelms:


puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foremanNightlyTest.groovy
```diff
@@ -10,9 +10,7 @@ pipeline {
         }
         stage('Provision Node') {
             steps {
-                dir('foreman-infra') {
-                    git url: 'https://github.com/theforeman/foreman-infra.git'
-                }
+                clone_foreman_infra(dir: 'foreman-infra')
```
komidore64: I feel like this is more obscure than what was originally there.

komidore64:
Where are you defining this method?

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foremanNightlyTest.groovy
komidore64: `git_clone_foreman_infra()` seems clearer though

ehelms:
Name changed and file added that has the function in it.

PR #780, opened 2018-08-17
convert timed jenkins job runs to cron

komidore64:


puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -22,36 +22,30 @@
...
+  cron { "jenkins-jobs-update-${config_name}-delete-old":
+    command     => "jenkins-jobs --conf ${inifile} update --delete-old ${directory}/${config_name} > /var/cache/jjb.xml",
+    hour        => 0,
+    minute      => 0,
+    environment => 'PATH=/bin:/usr/bin:/usr/sbin',
+    require     => File[$inifile],
   }
-  $cmd = "jenkins-jobs --conf ${inifile} ${subcmd} ${directory}/${config_name} > /var/cache/jjb.xml"
 
-  exec { "jenkins_jobs_update-${config_name}":
```
ekohl: I'm not sure we need to change this from an `exec` to a cron. The downside is that the time till update (on average) goes up. Right now it's running twice an hour. with an hourly cron it can go up to a lot. Worst case it's merged just after the puppet run has started (say 9:31). Then the next Puppet run is at 10:01 which means you'll have to wait for the 11:00 cronjob to update. I'd just update the removal of unmanaged to a daily cron.

komidore64:
@ekohl updated based on comments

PR #781, opened 2018-08-20
add foreman-ansible-modules job

sean797:
I need to make some changes to `foreman-ansible-modules` to before we merge this so initially, I've just got it testing the foreman_location module until I can show the pipeline is working. How can we test this before merging?

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
@@ -23,3 +23,24 @@ def runPlaybook(args) {
         sh "${command.join(' ')}"
     }
 }
+
+def containerPlaybook(playbook) {
+    dir('containers') {
+        runPlaybook(
+            playbook: playbook,
+            inventory: cico_inventory('../'),
+            extraVars: ['@vars/remote.yml'],
+            options: ['-b']
+        )
+    }
+}
+
+def ansibleModulesPlaybook(playbook, route) {
+    dir('ansible-modules') {
+        runPlaybook(
+            playbook: playbook,
+            inventory: cico_inventory('../'),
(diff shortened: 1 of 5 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -23,3 +23,24 @@ def runPlaybook(args) {
         sh "${command.join(' ')}"
     }
 }
+
+def containerPlaybook(playbook) {
```
ehelms: Can ya give a little more breakout and why move this into Ansible library since it's quite container specific?

ehelms:
There are a few things needed and some thoughts, I'll start with the former.

 1) I'll need to update FAM git repo so PRs can be tested on ci.centos.org. Fairly simple but will require some action on the ci.centos.org admin which is good to do ahead of time.
 2) You may want to consider adding some logic to do testing based on some set of criteria depending on what changed. If module X changes, run playbook Y (or playbook with a similar name scheme) to avoid the whole test suite on every change. If module_utils changes run everything to ensure no breakages across.

I see that you are using the containerized deployment to run the tests against. I think that's totally rad as it adds more testing against the containerized deployment. I do want to caution and ensure that you are OK with the instability that may come from this choice. There is still some flux in the container design, with https://github.com/theforeman/forklift/pull/790 probably being the last major change followed by more small changes. This testing would also be uni-directional to start (in that containers could break FAM testing). I would not be opposed to exploring the idea of running FAM tests against container deployment testing.

sean797:
Thanks @ehelms I might have been a little eager in raising this, the discussion over in FAM seems like it might be taking a different course. Were hold off on this for now.

ehelms:
@sean797 can ya link the discussion in FAM?

sean797:
https://github.com/theforeman/foreman-ansible-modules/issues/154

PR #782, opened 2018-08-21
Refactor git_clone_foreman_infra function to not require args map

zjhuntin:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -1,6 +1,4 @@
-void git_clone_foreman_infra(args) {
-    target_dir = args.target_dir ?: ''
-
+void git_clone_foreman_infra(target_dir = '') {
```
ehelms: This signature shouldn't need changing.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/duffy.groovy
```diff
@@ -1,6 +1,6 @@
 def provision() {
     fix_ansible_config()
-    git_clone_foreman_infra(dir: 'foreman-infra')
+    git_clone_foreman_infra(target_dir = 'foreman-infra')
```
ehelms: I think this should just be `target_dir:`

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
zjhuntin: @ehelms  [here](https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanRelease.groovy#L33) it is failing at this point at the moment. It doesn't seem to recognize that args.target_dir should exist and fails out. Is there another way to not have it do that?

[This](http://ci.theforeman.org/job/foreman-nightly-release/326/) is an example failure

ehelms: Oh maybe we need to set a default value for args, something like `args = {}` in the parameters.

tbrisker:
Is it possible this didn't propagate to centos ci yet? I'm still seeing nightly failures with the same error.

ekohl:
So the deploy job is failing to deploy its own fix. Nice!

PR #785, opened 2018-08-22
don't make git_clone_foreman_infra void

evgeni:


ehelms:
What is the difference between the two? Or what is the failure?

evgeni:
To be honest, I don't really know.

For variables, `def foo` is same as `Object foo` (and well, `void` is `void`).

For methods, I'd expect this only to define the return type (which is irrelevant here), but w/o that change, stuff fails: https://ci.theforeman.org/job/katello-nightly-release/404/console
And with it, it works: https://ci.theforeman.org/job/katello-nightly-release/405/console

evgeni:
this does not happen in my local groovy shell, so it's kinda weird

evgeni:
okay. that's not it.

PR #789, opened 2018-08-23
Fix ci.centos.org race condition

ehelms:


evgeni:
as much as I hate sleeps, this is probably the cheapest workaround we can have now.

ACK if you filed an issue with ci.centos.org about it :)

ehelms:
Already filed by other users -- https://bugs.centos.org/view.php?id=14920

mmoll:
merged, thanks @ehelms!

PR #793, opened 2018-08-27
Switch to testing Katello on Ruby 2.5

ehelms:
Merge should wait on Katello code being updated to handle Ruby 2.5


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,10 +1,24 @@
...
+    },
+    'KATELLO-3.6': {
+        'foreman': '1.17-stable',
+        'ruby': '2.4'
+    },
+    'KATELLO-3.5': {
+        'foreman': '1.16-stable',
+        'ruby': '2.4'
+    }
 ]
```
evgeni: I think I'd prefer having

```groovy
def ruby = katello_versions[ghprbTargetBranch]['ruby']
def foreman_branch = katello_versions[ghprbTargetBranch]['foreman']
```

here and then using these names afterwards.

ehelms:
Updated to your request @evgeni

evgeni:
ACK, as the trello card is marked "WIP", not merging just yet.

PR #795, opened 2018-08-28
archive ci.centos.org data for easier job failure analysis

evgeni:
sometimes jobs fail because the cico data we got is incomplete/empty, this makes it easier to spot such failures

example failure, where it *seems* we got an empty response: https://ci.centos.org/job/foreman-katello-nightly-test/76/


evgeni:
@ehelms do you know if that data contains any secrects/tokens that should *not* be archived?

PR #796, opened 2018-08-28
Rename release jobs to pipeline per naming conventions

ehelms:


ehelms:
Note this will result in jobs being deleted and re-created unfortunately.

evgeni:
does this need changes to webhooks/triggers in github?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-nightly-pipeline.yaml
```diff
@@ -1,5 +1,5 @@
 - job:
-    name: foreman-nightly-release
+    name: foreman-nightly-pipeline
```
ekohl: I was thinking that we may want to rename this to `foreman-nightly-rpm-pipeline`. This would allow us to add a `foreman-nightly-deb-pipeline` as well. Same applies for most other jobs.

ehelms: We have always tended to treat each OS stream as independent for nightly but dependent for major releases right?

ekohl: This solution gives a certain bias to RPMs which is not how I'd like to view it. Explicit naming or unification can remove that impression. I don't know how feasible it is to unify them because my knowledge of Jenkins pipelines is too limited. If it is feasible and we can get to it in a reasonable timeframe then I don't mind this naming. Unification would actually have my preference because you need to watch fewer jobs to monitor the process.

ehelms: I view it less about mechanics of the job and more about our workflow design. Do we view RPM and Deb as separate release paths that should not gate each other? Should nightly as a whole be gated on both? Or should each release independent of the other?

ekohl: I think that repository promotion doesn't have to block each other. Should a Deb build issue hold back RPMs or vice versa? It looks like Jenkins now allows complex steps which means we could do:
```
setup → (prep el7 → test el7 → push el7)
        (prep d9 x86 → test d9 x86 → push d9 x86)
        (prep d9 arm → test d9 arm → push d9 arm)
....
```
Note that I don't know the exact Debian steps. For el7 prep would be mashing for example. Testing could also involve multiple (parallel) steps, like testing installation vs upgrade or even install on postgres, install on mysql and upgrade.

Previously I assumed this would be too complex for Jenkins pipelines but if this is technically possible I'd be happy to have it in a single job for visibility.

A counter argument could be efficiency: you still want to be able to repush el7 if you made an infra fix there but repushing all of the debs could be too much. Could we have partial restarts as well?

ehelms:
@ekohl I updated foreman-nightly-pipeline to foreman-nightly-rpm-pipeline

ehelms:
If approved, I'll try to rename in the UI before merging.

ehelms:
Updated to include `rpm` in Foreman job titles. I think the amount of effort to write API or cli to do here and there renames will cost more time than me simply doing it inside the UI.

PR #800, opened 2018-08-29
Move foreman plugins release to multi-stage parallel

ehelms:
Provides better visual for internal stages inside parallel steps:

screenshot from 2018-08-29 11-54-34


komidore64:
well ... maybe see if we can abbreviate those names, so there isn't gross wordwrap?

komidore64:
`repo-cl-1.18-el7` or something?

ekohl:
Perhaps just `Repoclosure 1.18-el7`? Then it uses 2 lines but not such an ugly one.

ehelms:
I tried that and it did not help. The UI appears to treat characters all equally and simply wrap after a certain number of them.

komidore64:
@ehelms understood. thanks for checking! :+1:

PR #801, opened 2018-08-30
Add virtualenv as a required package for theforeman.org Jenkins nodes

zjhuntin:


puppet/modules/slave/manifests/init.pp
```diff
@@ -143,6 +143,8 @@
       ensure => present;
     'ansible':
       ensure => latest;
+    'virtualenv':
```
ehelms: Did you check that this package exists and you can install it by this name on centos box?

ekohl: Both Debian and EPEL call it `python-virtualenv`

mmoll:
merged, thanks @zjhuntin!

PR #804, opened 2018-09-01
Refs #24780 - use native npm

mmoll:


ohadlevy:
3.10 is fairly old, and i am worried this is a major deviation from what developers are actually using, i would prefer to reduce (and even limit) the amount of npm/node combinations, in fact, I would be in favor of creating a simple docker image that builds assets, it can be reused across all the various distros, and its output can be simply added to the resulting tar/package...?

mmoll:
NPM 3.10 and NodeJS 6 are the versions coming from EPEL and thus are used to build the RPMs. The "npm install npm" dance was only a workaround and I had hoped that EPEL would somewhen provide a newer version for performance reasons, however the current npm package set seems to install and work just fine with npm 3 again.

Also, this combination is still "supported" until 04/2019.

> I would be in favor of creating a simple docker image

I would be strongly against that (at this moment). If we need newer node/npm versions, we should just tap the nodesource.org RPM repos, like we do for the DEB build, as this fits in much better in all the build processes we use currently.

PR #805, opened 2018-09-03
try harder to get a node from cico and fail properly

evgeni:
sometimes, cico returns

    {"message": "Requested servers successfully", "hosts": {}, "ssid": "012345679"}

with the hosts list being empty. This makes the data useless for further
processing. Try to get a list of hosts from cico again and if that
returns no hosts too, properly fail inside the playbook, instead of
writing an empty inventory and then failing in the next steps of the
pipeline.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
@@ -3,11 +3,24 @@
   vars:
     api_key: "{{ lookup('file', '~/duffy.key') }}"
   tasks:
-    - name: 'Get list of nodes'
+    - name: 'Get a new node'
       cico:
         action: get
         api_key: "{{ api_key }}"
       register: cico_data
+
+    - name: 'Retry geting list of nodes'
+      cico:
+        action: list
+        api_key: "{{ api_key }}"
+        ssid: "{{ cico_data.results.ssid }}"
+      register: cico_data
+      when: len(cico_data.results.hosts) == 0
+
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
```diff
@@ -3,12 +3,25 @@
...
     api_key: "{{ lookup('file', '~/duffy.key') }}"
   tasks:
-    - name: 'Get list of nodes'
+    - name: 'Get a new node'
       cico:
         action: get
         api_key: "{{ api_key }}"
       register: cico_data
 
+    - name: 'Retry geting list of nodes'
```
ehelms: Is the idea that the host was actually provisioned, but the return data did not provide useful information? Or that the host was never actually provisioned?

evgeni: I've seen jobs where the `cico` module returns success, but the hosts list is empty. I don't know if that's a race condition, and the host is provisioned later, or if the API wrongly returned success (which is more likely, but I didn't have the chance to poke at the API enough yet).

My idea was to at least retry the listing of hosts and fail loudly if that also fails.

The fail is actually more important here, as it makes the Jenkins job fail nicely, instead of a weird "no hosts matched" warning from ansible.

ehelms: OK, sounds like we just need to keep an eye on things after this.

PR #806, opened 2018-09-03
add rubocop to test_develop jobs

mmoll:


mmoll:
This is just a kludgy stopgap to check more thoroughly than hound does and to abort CI jobs with style violations before consuming further resources with tests.

PR #808, opened 2018-09-04
Redirect stats dashboard http to https

GregSutcliffe:
Cleaned up the params a bit too.

### DIFF
--- a/puppet/modules/web/manifests/stats.pp
+++ b/puppet/modules/web/manifests/stats.pp
@@ -1,7 +1,17 @@
+# Class for configuring the Apache proxy in front of Shiny Server
+#
+# === Parameters:
+#
+# $hostname::       FQDN for LetsEncrypt
+#
+# $webroot::        Location of the DocumentRoot (htdocs etc) folder
+#
+# $https::          Whether to enable the https vhost. Also redirects traffic from http
+#
 class web::stats(
-  $hostname = 'stats.theforeman.org',
-  $webroot = '/var/www/vhosts/shiny/htdocs',
-  $https = false,
+  String $hostname              = 'stats.theforeman.org',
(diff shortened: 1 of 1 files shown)


puppet/modules/web/manifests/stats.pp
```diff
@@ -19,13 +32,24 @@
...
   }
 
-  apache::vhost { 'shiny_server':
-    port          => '80',
-    servername    => $hostname,
-    docroot       => $webroot,
-    docroot_owner => $::apache::user,
-    docroot_group => $::apache::group,
-    proxy_pass    => $proxy_pass,
+  if $https_only {
```
ekohl: Maybe we should unify this with `$https`? I don't see a reason to allow `https` but not run `https_only` these days.

GregSutcliffe: Yeah, fair point. Fixed.

GregSutcliffe:
I noticed Redmine is using this code without an exception for LE, and it seems to be working?

ekohl:
That sounds correct. Looks like it prefers the actual files if they're present and we have the docroot set correctly.

puppet/modules/web/manifests/stats.pp
```diff
@@ -19,16 +29,15 @@
...
-  }
-
   if $https {
+    apache::vhost { 'shiny-server':
+      port          => '80',
+      servername    => $hostname,
+      docroot       => $webroot,
+      docroot_owner => $::apache::user,
+      docroot_group => $::apache::group,
+      redirect_dest => "https://${servername}/",
```
ekohl: 0710256ead6a01821ad5c0589e7eb19b49ddeee0 was needed to set this to `$hostname`

PR #809, opened 2018-09-04
Fetch list again always just to be safe from duffy

ehelms:


ehelms:
In Ansible, the filter length has to be used (`|length`) instead of `len()`. Further, if you register a variable in a task with a when conditional, the variable gets filled by the value of the conditional when it's false. Thus I swapped this to just always fetch the list again.

puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
```diff
@@ -15,12 +15,14 @@
         api_key: "{{ api_key }}"
         ssid: "{{ cico_data.results.ssid }}"
       register: cico_data
-      when: len(cico_data.results.hosts) == 0
```
zjhuntin: Don't you still want this here just as `when: cico_data.results.hosts|length == 0` like below?

zjhuntin: Never mind, realized what you were saying int he description now.

PR #811, opened 2018-09-05
Update `stable` manual version to `1.19`

alexjfisher:
There is no copy for 1.20 yet, so I've left `latest` and `next` as is.


mmoll:
merged, thanks @alexjfisher!

PR #812, opened 2018-09-05
use deb.debian.org as Debian mirror

mmoll:


mmoll:
@evgeni This should be the best canonical option, correct?

ekohl:
Should the ARM builders also be changed?

evgeni:
better than random ftp.de and ftp.us anyways ;)

we only have stretch and newer, right? older (until jessie) apt does not support SRV DNS records, and you end up with a few more RTTs for HTTP redirects. but the speed is still better I guess.

PR #813, opened 2018-09-06
Use Ruby 2.5 more often

mmoll:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -11,7 +11,7 @@ pipeline {
     environment {
         version = env.getProperty('version')
         major_version = env.getProperty('major_version')
-        ruby_ver = '2.4.0'
+        ruby_ver = '2.5.1'
```
ekohl: Is there a reason to prefer 2.5.1 over 2.5?

mmoll: I guess not.

mmoll:
OK, I removed the kafo related commit for now

PR #814, opened 2018-09-06
shift testing focus from Ruby 2.4 to 2.5

mmoll:


mmoll:
The Rails 5.2 SCL is Ruby 2.5 based, so this should get reflected here.

PR #815, opened 2018-09-09
Fix pbuilder cleaning

ekohl:
The subdirectories of result might not exist. This causes an error when cleaning. The mindepth ensures the directory itself isn't removed.


mmoll:
merged, bedankt @ekohl!

PR #820, opened 2018-09-12
Update obal everytime if it exists

ehelms:


ehelms:
If we feel comfortable swapping to a release strategy I can be ok with it. @evgeni thoughts?

PR #826, opened 2018-09-18
set 120 mins timeout for katello-pr-test

mmoll:


mmoll:
I hope this also works for the groovy jobs...

komidore64:
I think for pipeline jobs it is cleaner to keep as much configuration in groovy as possible.

```groovy
pipeline {
    [...]
    options {
        timestamps()
        timeout(time: 2, unit: 'HOURS')
        ansiColor('xterm')
        buildDiscarder(logRotator(numToKeepStr: '15'))
    }
    [...]

}
```

komidore64:
Or would it make even more sense to have a default timeout of 120 mins for all jobs? Those that need more can override.

mmoll:
Seems this would only be possible with a workaround... https://issues.jenkins-ci.org/browse/JENKINS-26778

mmoll:
edited the grovvy script now.

PR #828, opened 2018-09-25
Call setup_obal explicitly

ehelms:
The git checkouts on each call to obal() will fail if within a
parallel block. Let's make this explicit until we switch to the
Obal package.


zjhuntin:
The only other place that might also need this called explicitly is `tfmROR51.groovy`

ehelms:
Updated

PR #830, opened 2018-10-04
Drop Katello client repoclosure from nightly

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy
```diff
@@ -112,8 +101,6 @@ void repoclosure(repo, dist, additions = []) {
             "-l ${dist}-scl",
             "-l ${dist}-scl-sclo",
```
ekohl: I think we can also drop this since it no longer exists in `yum_el7.conf`

PR #831, opened 2018-10-04
Add foreman-client-release symlink

ehelms:


jturel:
I don't think there's a feasible way for me to test this but it looks sane!

PR #833, opened 2018-10-04
drop i386 builds

mmoll:


tbrisker:
don't we also need to remove where we set $arch to i386? or is that done in jenkins somewhere?

mmoll:
For x86 is set in Jenkins and in the two modified scripts, xxx64 is amd64 and xxx32 is i386.

PR #836, opened 2018-10-09
Send email to discourse on pipeline failure

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy
```diff
@@ -92,6 +92,15 @@ pipeline {
             }
         }
     }
+    post {
+        failure {
+            emailext(
+                subject: "${env.JOB_NAME} failed",
+                to: 'ci@community.theforeman.org',
+                body: "Foreman nightly pipeline failed: \n\n${env.BUILD_URL}"
```
GregSutcliffe: Should this string be unique from the one above, since it's from a different build?

ehelms: Doh, good catch :)

GregSutcliffe:
:+1:

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanRelease.groovy
```diff
@@ -60,6 +60,15 @@ pipeline {
             }
         }
     }
+    post {
+        failure {
+            emailext(
+                subject: "${env.JOB_NAME} failed",
```
ekohl: Should this include the job ID to make it unique?

ehelms:
We could explore IRC notifications, but I do want to keep this focused on the plan for discourse. Also, updated the subject.

PR #838, opened 2018-10-11
Update web class for 1.20 branching

tbrisker:
in preparation for branching


ekohl:
Depends on https://github.com/theforeman/theforeman.org/pull/1204

PR #839, opened 2018-10-11
Change plugin pipeline to work around repoclosure concurrency

ehelms:


ehelms:
@ekohl Not the prettiest code but works, can see example run: http://ci.theforeman.org/blue/organizations/jenkins/foreman-plugins-rpm-pipeline/detail/foreman-plugins-rpm-pipeline/386/pipeline

PR #842, opened 2018-10-18
Add 1.20 client RPM pipeline

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/client/1.20.groovy
```diff
@@ -0,0 +1,77 @@
...
+            post {
+                always {
+                    deleteDir()
+                }
+            }
+        }
+    }
+}
+
+void push_rpms(version, distro) {
```
evgeni: should these two be in a `lib` somewhere?

ehelms:
@evgeni Attempted DRY, let me know what ya think

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/repoclosure.groovy
```diff
@@ -0,0 +1 @@
+
```
evgeni: this file is empty

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/client/1.20.groovy
```diff
@@ -0,0 +1,66 @@
...
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Mash Koji Repositories') {
+            agent { label 'sshkey' }
+
+            steps {
+
+                mash("foreman-client-mash-split-1.20.py")
```
ekohl: Could we put `1.20` in a variable?

PR #844, opened 2018-10-22
archive the extra vars file

evgeni:
this can have valuable information for debugging


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -10,6 +10,7 @@ def obal(args) {
     }
 
     writeYaml file: extra_vars_file, data: extra_vars
+    archiveArtifacts artifacts: extra_vars_file
```
ehelms: Could this ever contain passwords?

evgeni: technically, in the future, sure.
but as of today, we don't pass (or use) any passwords when calling `obal`.

PR #854, opened 2018-10-30
set obal to be relative to the job's workspace rather than pwd()

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -13,7 +13,7 @@ def obal(args) {
     archiveArtifacts artifacts: extra_vars_file
 
     wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
-        withEnv(['ANSIBLE_FORCE_COLOR=true', "PYTHONPATH=${pwd()}/obal"]) {
+        withEnv(['ANSIBLE_FORCE_COLOR=true', "PYTHONPATH=${env.WORKSPACE}/obal"]) {
```
evgeni: `setup_obal()` will create an `obal` subdir *wherever* it is called from, so both `pwd()` and `end.WORKSPACE` feel a bit wrong.

shall we make `setup_obal()` create it at `env.WORKSPACE` in the first place? :)

PR #855, opened 2018-10-31
release_nightly_build_rpm - don't rely on packaging_build_rpm anymore

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/release-nightly-build-rpm.groovy
```diff
@@ -0,0 +1,38 @@
...
+            }
+        }
+
+        stage("Obal Nightly Build RPM") {
+            steps {
+                script {
+                    if (env.jenkins_job) {
+                        releaserArgs << "--arg jenkins_job=${env.jenkins_job}"
+                    }
+                    if (env.jenkins_job_id) {
```
ekohl: You could argue that if either of these parameters are empty it's actually a misconfiguration and you could fail on it before git clone.

komidore64: The fact that the `jenkins_job` parameter defaults to nothing makes me lean towards no failing the job if the parameter is left empty.

ekohl: Hmm, you're probably right since the manifest includes the job name already it may not be needed.

ehelms:
Job my memory of the design... this is another intermediary step until individual jobs are created utilizing library functionality instead of a central job?

komidore64:
Correct.

PR #867, opened 2018-11-19
Move build_description setting to job stage due to different pipeline…

ehelms:
… agents

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
@@ -15,5 +15,9 @@ def set_job_build_description() {
        build_description += "<a href=\"${link}\">${files_list[i]}</a><br/>"
     }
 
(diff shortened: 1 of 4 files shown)


ehelms:
Testing this first

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -15,5 +15,5 @@ def set_job_build_description() {
        build_description += "<a href=\"${link}\">${files_list[i]}</a><br/>"
     }
 
-    currentBuild.description = build_description
+    currentBuild.description += build_description
```
ekohl: I wonder if description is ever null

ehelms: You wonder correctly, and I had been testing an update to handle that.

ehelms:
@ekohl check the second commit to handle the root ownership issues we've been seeing

PR #868, opened 2018-11-20
add foreman-proxy-develop-test and foreman-proxy-develop-release

komidore64:
and refactor `foreman-installer-develop-release` into a generic job that can be used for other nightly build jobs.


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,30 @@
+def setup_nightly_build_environment(args) {
+    def commit_hash = ''
+
+    dir(args.package_name) {
+        git(url: "https://github.com/${args.github_repo}.git", branch: 'develop')
```
evgeni: make the branch configurable, as some projects use `master` instead of `develop`

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,30 @@
...
+        commit_hash = archive_git_hash()
+    }
+    dir('foreman-packaging') { git(url: 'https://github.com/theforeman/foreman-packaging.git', branch: 'rpm/develop') }
```
evgeni: linebreaks, please :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,30 @@
...
+    configureRVM(args.ruby_version ?: env.ruby_version)
+
+    return commit_hash
+}
+
+def generate_nightly_sourcefiles(args) {
+    def sourcefile_paths = []
+
+    dir(args.package_name) {
+        withRVM(["bundle install --jobs 5 --retry 5"], env.ruby_version)
```
evgeni: the previous step allows `args.ruby_version`, this not anymore. intentional?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,30 @@
...
+        withRVM(["bundle exec rake pkg:generate_source"], env.ruby_version)
+        archiveArtifacts(artifacts: 'pkg/*')
+        dir('pkg') {
+            sourcefile_paths = sh(script: "ls -1", returnStdout: true).trim().tokenize('\n').collect {
```
evgeni: we have similar code in
```
puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy:    def files_list = sh(script: "ls jobs/ -1", returnStdout: true).trim().split()
puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/koji.groovy:        tasks = sh(returnStdout: true, script: "ls kojilogs -1 |grep -o '[0-9]*\$'").trim().split()
```
should we make this more generic and a library function?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: Yeah, that's a good point. I was going to add this feature down the line but might as well do it now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: I thought this looked cleaner in-line being that it was just a single `git()` call, but I'm okay wit uniformity :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: Good idea

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: writer error.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,36 @@
+def setup_nightly_build_environment(args) {
+    def commit_hash = ''
+    def package_name = args.package_name
```
ekohl: I'm not too happy about package_name. For the examples you have here they are the same for every distro, but if we start building nightlies for plugins, then they are suddenly different. Perhaps `project_name` would be more accurate?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -17,24 +17,21 @@ pipeline {
     stages {
```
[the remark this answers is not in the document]
komidore64: https://github.com/theforeman/foreman-infra/blob/456cf74f8bb9ce3428a118ab48b5456534e868d3/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy#L14-L18

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,36 @@
...
+    def git_url = args.git_url
+    def branch = args.branch
+    def ruby_version = args.ruby_version ?: env.ruby_version
+
+    dir(package_name) {
+        git(url: git_url, branch: branch)
+        commit_hash = archive_git_hash()
```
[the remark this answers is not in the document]
komidore64: https://github.com/theforeman/foreman-infra/blob/456cf74f8bb9ce3428a118ab48b5456534e868d3/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/git.groovy#L14-L18

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: I chose `package_name` because it is the name of the package that we're asking obal to build.

It differs for `foreman-proxy`. The repo/project is titled `smart-proxy` whereas the package built by obal is titled `foreman-proxy`.

Or maybe I'm not entirely understanding why you aren't happy with the variable name.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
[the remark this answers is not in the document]
komidore64: Well, all it's building right now are `foreman-installer` and `foreman-proxy` which is why I've called it `foreman-x`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -45,10 +42,10 @@ pipeline {
...
                             obal(
                                 action: 'nightly',
-                                packages: 'foreman-installer',
+                                packages: package_name,
                                 extraVars: [
-                                    'releasers': [ 'koji-foreman' ], // TODO: remove releasers once foreman-installer is set to `{}` in foreman-packaging
-                                    'nightly_sourcefiles': [ "${env.WORKSPACE}/foreman-installer/pkg/${sourcefile}" ],
+                                    'releasers': [ 'koji-foreman' ],
+                                    'nightly_sourcefiles': sourcefile_paths,
                                     'nightly_githash': commit_hash
```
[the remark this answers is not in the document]
komidore64: I'm curious which parts aren't obvious.

Looking at the `setup_nightly_build_environment` method definition, it contains `return commit_hash`. I'm assigning `setup_nightly_build_environment(...)`'s return value to a variable titled `commit_hash`.

If you're argument is that they're a bit conceptually disjointed, I agree with you there. I could probably take a second crack at not having the commit hash be a psuedo side-effect of `setup_nightly_build_environment`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
ekohl: If you have a plugin it's `rubygem-foreman_memcache` for RPMs but `ruby-foreman-memcache` on Debian. Now I know obal doesn't handle debs right now, but if it's aligning to obal, perhaps `obal_name` would be clearer?

On a side note: longer term I would like obal to understand git repos and how to build archives as well so you can locally point it to a git checkout and let it build a package like you can do with the current `koji-*-jenkins` releaser.

komidore64: `project_name` does seem clearer for the current state of `foreman-infra` nightly building. Down the road I'm okay with renaming things to reflect concepts once `obal` can do more.

ekohl:
I still disagree with moving the building of the artifact to the release job. A lesser concern is that we check out an additional git repo in the release job where we previously didn't have to. My main concern is race conditions:
* Developer merges commit x to develop
* CI runs test_develop on commit x
* Developer merges commit y to develop
* test_develop from commit x triggers release_develop
* release_develop checks out develop and builds an archive of commit y

Now we are running untested code in the release where previously we had a guarantee that tests actually passed for the code we pushed to nightlies.

komidore64:
Building an artifact is part of the release process, not the testing process.

While it is technically possible for the scenario you described to occur, I believe the likelihood if it doing so is low. I would however would like to explicitly prevent that scenario from happening in the future.

ehelms:
@komidore64 You had an idea to trigger the release job with the hash from test_develop right? Which would achieve both concerns of keeping source generation in a release job (not a test job) and ensuring source is built from a valid hash?

I think we are debating some philosophical differences here in how we think about post merge activities. One thought I had was that maybe `test_develop` is misleading and what we are really intending is a `foreman-develop-source-release`. With the idea that the job:

 * verifies the develop branch via tests
 * generates a verified source
 * archives the source and commit hash

While this doesn't change the overall workflow that much, it does change the perception of what we actually care about workflow wise. Which I think drills down to:

 * Developer merges X to develop
 * Verified source is generated
 * Nightly RPM and Deb packages are built from the verified source
 * Integration tests are run to verify packages (aka nightly test pipeline) per distro

I do realize some of this discussion strays from the design you put together @komidore64 and that we reviewed and ACK'd. So I do apologize for the churn that is happening as we see the work live an in person. I think the changes being made are important. And I think it's good to talk about the goals and workflow we see ourselves wanting for how to take a change (aka a commit) and get it into the release repositories.

ekohl:
@komidore64 is correct that testing and releasing are separate concerns. I had the same thought that I wanted to post here, but @ehelms beat me to it. Archiving the commit hash in the test job achieves the guarantee we have now while still separating the concerns.

komidore64:
I appreciate y'all continuing the discussion and helping me make our Jenkins jobs better both functionally and conceptually.

Since Opening this pull-request, I've also opened #872 which includes changes from this one. On top of that I've since started refactoring `foreman-proxy-develop-test` into a pipeline for the purpose of passing the git hash along to `foreman-proxy-develop-release`.

I am going to close this pull-request and #872, to open a single pull-request with all the changes I've made so far.

PR #869, opened 2018-11-21
always use extra vars from a file

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foremanRelease.groovy
```diff
@@ -38,10 +38,10 @@ pipeline {
                     runPlaybook(
                         playbook: 'ci/centos.org/ansible/jenkins_job.yml',
                         extraVars: [
-                            "jenkins_job_name=foreman-nightly-test",
-                            "jenkins_username=${env.USERNAME}",
-                            "jenkins_password=${env.PASSWORD}",
-                            "jenkins_job_link_file=${env.WORKSPACE}/jobs/foreman-nightly-test"
+                            "jenkins_job_name": "foreman-nightly-test",
+                            "jenkins_username": "${env.USERNAME}",
+                            "jenkins_password": "${env.PASSWORD}",
```
evgeni: As `runPlaybook` now uses `writeExtraVars` and that always runs `archiveArtifacts` on the generated `extra_vars_file`, this will leak the Jenkins password.

I see multiple possible solutions here, which I'd like to discuss:

1. pass `-e "jenkins_password=${env.PASSWORD}"` as `options` instead of `extraVars`, which will use the same protection by Jenkins (replacing known passwords with `********`) as we used before.
1. add a `hiddenExtraVars` option, that will generate a second `extra_vars_file` that is *not* archived
1. make `writeExtraVars` not archive the file by default

I think I am leaning to `hiddenExtraVars`, even if it's more code.

ekohl: I'm leaning to 2 as well, perhaps named `sensitiveExtraVars`.

ehelms: Is #1 do-able without having to get to hacky in knowing when its a password or username?

evgeni: @ehelms when you are in a `withCredentials` block, Jenkins automatically replaces all credential strings in the output with asterisks (that's what we've been using in before this PR too), so we don't have to do anything special for that to work.

evgeni: I have implemented #2 for now, let me know what y'all think

PR #875, opened 2018-11-30
refactor foreman-proxy, foreman-installer, and foreman-proxy nightly tests and releases

komidore64:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
...
+            userRemoteConfigs: [[url: git_url]]
+        ])
+    }
+    dir('foreman-packaging') {
+        git(url: 'https://github.com/theforeman/foreman-packaging.git', branch: 'rpm/develop')
+    }
+    setup_obal()
+    configureRVM(ruby_version)
+
+    return commit_hash
```
ehelms: Guessing this can be dropped now?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+            }
+        }
+        stage("test ruby-2.5 & puppet-5.3.3") {
+            steps {
+                run_test(ruby: '2.5', puppet: '5.3.3')
+            }
+        }
+        stage("Release Nightly Package") {
+            steps {
+                build(
```
ehelms: This will fail the job if the RPM or Debian don't build IIRC. I've always debated internally if that was the behavior we wanted. Given this is the test job that produces a tested and valid commit hash. My thinking is we don't want this job to be marked as failed just because the release failed. I think that would mean either marking this as don't fail if this build fails or to reverse the triggering (i.e. put the trigger on the release job).

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
...
+}
+
+def generate_nightly_sourcefiles(args) {
```
ekohl: Just some bikeshedding: this isn't just for nightly so `generate_sourcefiles`?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
+def setup_nightly_build_environment(args) {
```
ekohl: This is my limited knowledge of groovy: why do you use args instead of named arguments here?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
...
+        // the full checkout() is necessary for checking out specific commits,
+        // as opposed to just branches with git()
+        checkout([
+            $class : 'GitSCM',
+            branches : [[name: git_ref]],
+            extensions: [[$class: 'CleanCheckout']],
+            userRemoteConfigs: [[url: git_url]]
+        ])
+    }
+    dir('foreman-packaging') {
```
ekohl: This and obal is only needed when building RPMs

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
...
+        git(url: 'https://github.com/theforeman/foreman-packaging.git', branch: 'rpm/develop')
+    }
+    setup_obal()
+    configureRVM(ruby_version)
```
ekohl: Only needed when building source files so I wonder if this shouldn't be part of that function.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,42 @@
+def setup_nightly_build_environment(args) {
+    def commit_hash = ''
```
ekohl: This is unused now

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+    stages {
+        stage("Collect Git Hash") {
+            steps {
+                git(url: 'https://github.com/theforeman/smart-proxy', branch: 'develop')
+                script {
+                    commit_hash = archive_git_hash()
+                }
+            }
+        }
+        stage("test ruby-2.1 & puppet-4.10.9") {
```
ekohl: Previously all testing was done in parallel, possibly on different slaves. Is it possible to do the same in pipelines? Otherwise the tests will take a lot longer.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+pipeline {
+    agent any
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
```
ekohl: Where did Ruby 2.0 go? EL7 on production still runs Ruby 2.0 so it's still a very important target.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+            steps {
+                run_test(ruby: '2.3', puppet: '4.10.9')
+            }
+        }
+        stage("test ruby-2.4 & puppet-4.10.9") {
+            steps {
+                run_test(ruby: '2.4', puppet: '4.10.9')
+            }
+        }
+        stage("test ruby-2.4 & puppet-5.3.3") {
```
ekohl: If we're cleaning up, this can be removed IMHO since we'll never really run this in production. We'll always use the REST API with Puppet > 4.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+            steps {
+                run_test(ruby: '2.1', puppet: '4.10.9')
+            }
+        }
+        stage("test ruby-2.2 & puppet-4.10.9") {
+            steps {
+                run_test(ruby: '2.2', puppet: '4.10.9')
+            }
+        }
+        stage("test ruby-2.3 & puppet-4.10.9") {
```
ekohl: Testing with Puppet 3.8.7 would actually make more sense here since that's what Xenial runs out of the box. With Puppet 4 we don't use the actual gem anyway.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
```diff
@@ -0,0 +1,89 @@
...
+            steps {
+                run_test(ruby: '2.4', puppet: '4.10.9')
+            }
+        }
+        stage("test ruby-2.4 & puppet-5.3.3") {
+            steps {
+                run_test(ruby: '2.4', puppet: '5.3.3')
+            }
+        }
+        stage("test ruby-2.5 & puppet-4.10.9") {
```
ekohl: Could also go for the above reasons

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: It allows you to pass in key-value pairs when calling.

```groovy
setup_nightly_build_environment(
    git_url: 'https://github.com/theforeman/smart-proxy.git',
    git_ref: 'develop',
    project_name: 'foreman-proxy'
)

I mostly stick with it to keep in line with most other helper methods I've seen in `foreman-infra`.

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy (hunk 5 above)
Comment (komidore64) [reply puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy]: Sure, but currently all of our nightly packages build an RPM.

**Comment by komidore64** [comment]
@ekohl I am sticking with the same rubies and puppets that are currently used in `test_proxy`. Previously it sounded like you were suggesting to get rid of puppet 3 testing entirely in theforeman/smart-proxy#624 but now asking about where ruby 2.0 went makes me think that wasn't what you were saying.

Discussing which rubies and/or puppets should be added to or removed from smart-proxy testing is outside the scope of this PR.

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy (hunk 8 above)
Comment (komidore64) [reply puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy]: `foreman-proxy-develop-test` took about 16 minutes to run when I tried it just now. That time includes waiting for `foreman-proxy-develop-release` to finish.

I don't consider this a significant negative.

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy (hunk 4 above)
Comment (ehelms) [reply puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy]: This makes the calling of the method clearer to read but the method signature itself less clearer. We've err'd on the side of making the calling of the method clearer to read to make reading pipelines more obvious.

**Comment by komidore64** [comment]
@ekohl ping. this has been updated.

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy (hunk 5 above)
Comment (komidore64) [reply puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy]: We are currently only building RPMs with this pipeline.

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy (hunk 6 above)
Comment (komidore64) [reply puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy]: All sources build in this PR are ruby based (or require a ruby) to build the source.

**Comment by ekohl** [comment]
> @ekohl I am sticking with the same rubies and puppets that are currently used in `test_proxy`. Previously it sounded like you were suggesting to get rid of puppet 3 testing entirely in [theforeman/smart-proxy#624](https://github.com/theforeman/smart-proxy/pull/624) but now asking about where ruby 2.0 went makes me think that wasn't what you were saying.

Correct. The Ruby versions are still important and all used in production setups, but the Puppet 3 gem is no longer supported. So rather than a matrix (ruby * puppet) it can be simplified to just a vector (ruby). I'm also fine with doing that as a separate step since it should also be done for the PR testing version of the proxy.

**Comment by komidore64** [comment]
I decided to drop foreman from this PR, and we'll look at the ruby * puppet matrix later.

@ekohl is there anything blocking this from being merged?

File: puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy [hunk 13]
```diff
@@ -17,51 +17,59 @@ pipeline {
     stages {
         stage('Setup Environment') {
             steps {
-                dir('foreman-installer') {
-                    git(url: 'https://github.com/theforeman/foreman-installer.git', branch: 'develop')
-                    script { commit_hash = archive_git_hash() }
+                setup_nightly_build_environment(
+                    git_url: git_url,
+                    git_ref: (env.git_ref ?: git_ref),
```
ekohl: If you rerun the job (in case it fails) you could be rebuilding the wrong version. Could we error out if no git_ref was given? That's probably enough to prevent human error.

komidore64: I need to look at Jenkins to confirm, but I think it copies the same parameters if you rebuild, yes?

komidore64: If you build the job without handing in a `git_ref`, the job just assumes `develop`. Is that not desired?

ehelms: HEAD would not be desired given we want to build the last successfully tested commit. A human could enter HEAD but that should be a conscious choice which is what @ekohl is getting at.

komidore64:
rebased to see if that fixes strange travis errors

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
komidore64: Okay. I've refactored these jobs to error if there is no `git_ref` given.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -17,51 +17,64 @@ pipeline {
     stages {
         stage('Setup Environment') {
             steps {
-                dir('foreman-installer') {
-                    git(url: 'https://github.com/theforeman/foreman-installer.git', branch: 'develop')
-                    script { commit_hash = archive_git_hash() }
+                script {
+                    if ( env.git_ref?.isAllWhitespace() ) {
```
evgeni: is that like Ruby's `&` in `nil&.something`?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
evgeni: but you have `build_deb = true` for some jobs below, or am I missing something?

ekohl: True, but I'm always a fan of being explicit about what's there for which reason.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -0,0 +1,39 @@
...
+        // as opposed to just branches with git()
+        checkout([
+            $class : 'GitSCM',
+            branches : [[name: git_ref]],
+            extensions: [[$class: 'CleanCheckout']],
+            userRemoteConfigs: [[url: git_url]]
+        ])
+    }
+    dir('foreman-packaging') {
+        git(url: 'https://github.com/theforeman/foreman-packaging.git', branch: 'rpm/develop')
```
ekohl: I've noticed multiple jobs that have a git URL there will start to monitor it. For example, the installer is currently built on every merge to `rpm/develop` because the webhook is delivered. Would that also happen here?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
ekohl: The point I'm making is that I'm not sure a setup function is the right approach. In the UI there's a button that allows you to rerun a stage. Because we split setup and actual run into separate stages, that always fails because then you're missing the setup parts. Perhaps it's a bigger issue and doesn't have to block this PR.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
komidore64: `?` before a method call makes it nullsafe.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: To which job are you referring? If a job has polling setup, then it will poll for changes on _every_ repo that is contained in the pipeline.

ekohl: Maybe it was a misdiagnosis from my side. The `test_develop` and `test_proxy_develop` jobs currently trigger a lot of builds. I don't exactly where I saw it so I don't mind merging and seeing where the pieces land ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: Okay. I'd consider that a larger pipeline design issue/discussion because we have *lots* of pipelines with a setup task.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: > but you have `build_deb = true` for some jobs below, or am I missing something?

I probably could've been clearer. We are currently only building RPMs *within* the pipelines included in this PR. DEB builds are carried out in other jobs.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: It looks like both [`test_develop`](http://ci.theforeman.org/job/test_develop/changes) and [`test_proxy_develop`](http://ci.theforeman.org/job/test_proxy_develop/changes) are both triggering properly?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
ekohl: Agreed

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
ekohl: They are triggering, but the problem is that every part in the matrix triggers a RPM build job. On http://koji.katello.org/koji/packageinfo?packageID=996 you can see we build every git hash 4 or 5 times. @ehelms thought it was something with the interaction between matrix jobs and pipeline jobs.

ehelms:
What are the list of current concerns to merging this code and seeing how it operates?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
komidore64: Oh oh! I am following you now. I believe that a pipeline rather than a matrix would solve this problem as long as there is only one stage that triggers the build job after all the test stages have completed.

ekohl: https://github.com/theforeman/foreman-infra/pull/890

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-proxy-develop-test.groovy
ekohl: @ehelms suggested that if you use `agent` you can actually run them on separate nodes and in parallel but I guess we can evaluate this at a later stage.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_1_17_stable.yaml
```diff
@@ -30,7 +30,7 @@
     execution-strategy:
```
ekohl: This reminds me: we deprecated 1.17 so we should remove its jobs.

ehelms:
@komidore64 needs a rebase

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
evgeni: gotcha

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -17,51 +17,64 @@ pipeline {
...
+                    dir(project_name) {
+                        commit_hash = archive_git_hash()
                     }
                 }
             }
         }
+        stage('Build and Archive Source') {
+            steps {
+                script {
+                    sourcefile_paths = generate_sourcefiles(project_name: project_name)
```
ekohl: I remember a reason why we previously generated the archive as a separate step: it's not idempotent for the installer. It retrieves the nightly modules from the internet so running it twice can yield different results.

ehelms:
The trigger issue is not in play here as far as I can tell. Except for the installer, all other nightly packaging jobs are kicked off by a test job explicitly rather than polling the Git repository. And the test jobs while kicked off by polling the Git repository, only clone the git repository that matters. Therefore, the test job should get kicked off only when changes are made to the source repository and if successful kickoff a packaging job.

The problem with the installer is that we do not have a separate test job. The installer merges directly trigger packaging job.

The general trigger problem is in part due to how we clone tooling repositories and we should talk about switching them. Either to start using Obal installed from pip or an Obal container to avoid the git clone. As for handling the dual clone of the source repository and the packaging repository, that is trickier. There is possibly still some combination of things that disable polling properly for a git repository, I have not found it yet and its hard to test. This means we could try to build some master job to handle triggering or build it into the processor. Or fix the Git/Github plugin if in fact thats the issue.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
ehelms: Which is ok for the installer since it has no 'merged to master test job' and merges kickoff this job so source is only generated once?

If after going this route, we are still not satisfied with re-building the source then may we need to re-think back to the old strategy but instead of using a Jenkins job as the source of truth, the source is uploaded to a static server for consumption.

ekohl: You're right. Doing this in a single job does mean you get a chicken and egg problem when it's broken. If you need a fix in packaging but also need a recent release artifact for it you can no longer test it in CI. You must do it manually or blindly merge it. Currently the building of a release artifact is one phase and continues as long as the source repository is working. Releasing that into packages is a separate concern. That means our CI can rely on the latest release artifact and test our packaging PRs appropriately. In the new situation we lose this ability. This tight coupling means that any platform issue (EL7, Debian 9, Ubuntu 16.04, Ubuntu 18.04) issue will limit our testing ability. That is my major concern with this design.

I do see that a single pipeline provides a lot more visibility for us humans and would have greatly increased the chance we'd have our recent continuous nightly failure. Because that's more important I'd be OK with proceeding with this, especially because there's another way to regain the testing ability in CI.

komidore64:
I rebased and added one more commit to set polling for `foreman-packaging` and `obal` to FALSE

ekohl:
I'm OK with merging this after squashing, but would like to focus on getting 1.21.0-RC1 out this week. Shall we aim at merging this Monday? Then we can monitor it during the week and see if there are any issues.

komidore64:
@ekohl ready when you are.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-proxy-develop-release.yaml
```diff
@@ -0,0 +1,24 @@
+- job:
+    name: foreman-proxy-develop-release
```
ekohl: Bike shedding: I thought we based the job names on the git names so should this be smart-proxy?

komidore64:
@ekohl updated.

PR #876, opened 2018-12-04
Drop become from jenkins job kickoffs

ehelms:


evgeni:
I had the very same idea while working on #869 :)
Thanks!

PR #877, opened 2018-12-04
Skip Ubuntu tests in release_nightly_test_deb

ekohl:
Currently nova provisioning is broken on rackspace for Ubuntu. It doesn't set the fqdn correctly which makes the installer fail. This means we haven't had nightlies for a long time. While this isn't pretty, it's better than having no nightlies at all.


mmoll:
@ekohl I guess the DEBs of bionic and xenial will be pushed, if the stretch builds are green then?

ekohl:
That is the current implication yes. I'm tracking the Ubuntu issue in a support ticket and will update once it's resolved. The last update is that Canonical is testing the new nova version.

mmoll:
merged, bedankt @ekohl!

PR #880, opened 2018-12-07
don't fail when we could not get the current time for extraVars

evgeni:
on ci.centos.org, we can't use java.util.Date.getTime() in post actions
due to sandboxing. let's not fail in this case.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
@@ -31,8 +31,14 @@ def runPlaybook(args) {
 }
 
 def writeExtraVars(args) {
-    def timestamp = new Date().getTime()
-    def extra_vars_file = 'extra_vars-' + timestamp.toString() + '.yaml'
+    def suffix
+    try {
+        def timestamp = new Date().getTime()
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -31,8 +31,14 @@ def runPlaybook(args) {
...
 
 def writeExtraVars(args) {
-    def timestamp = new Date().getTime()
-    def extra_vars_file = 'extra_vars-' + timestamp.toString() + '.yaml'
+    def suffix
+    try {
+        def timestamp = new Date().getTime()
+        suffix = timestamp.toString()
+    } catch(Exception ex) {
+        suffix = 'no_timestamp'
```
ekohl: Should we fall back to a shell implementation?

evgeni: mhhh. we could.

ekohl: Alternatives are finding a job ID but I don't know if you have access to env vars.

evgeni:
right, that is better, closing

PR #882, opened 2018-12-07
Use extra vars from command line for centos jobs

ehelms:
Tested and worked as a solution to limit to ci.centos use. There really isn't any sensitive information to worry about but I can update to the same pattern there if you want.


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -43,3 +49,23 @@ def writeExtraVars(args) {
 
     return extra_vars_file
 }
+
+def buildExtraVars(args) {
+    def timestamp = System.currentTimeMillis()
+    def extra_vars_file = 'extra_vars-' + timestamp.toString()
```
evgeni: given it's a string, do we need to write it to file at all? as the file won't be used and the output will have the data anyways

PR #883, opened 2018-12-10
Add foreman-release to release_packages

ekohl:
foreman-release should only be released for RPMs on 1.20+.

This is untested so I'd appreciate a careful review.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_packages.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_packages.yaml
@@ -20,6 +20,7 @@
             - foreman-proxy
             - foreman-selinux
             - foreman-installer
+            - foreman-release
     builders:
       - trigger-builds:
         - project: packaging_build_rpm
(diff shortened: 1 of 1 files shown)


ekohl:
This will be unable to release to client repos so still incomplete.

ehelms:
This should call `obal release` which would build it for the client repos at least. The client release job runs everyday so it would then pick it up within 24 hours.

ekohl:
The issue is that foreman-release is in both foreman-1.20-rhel7 and foreman-client-1.20-rhel7 tags and obal can't deal with that. In practice I've seen it not end up in foreman-client-1.20-rhel7 and obal exits with code 1. That means release_packages will exit with a failure and you need to manually resume the pipeline.

ehelms:
The issue you mention should only occur if the client repository tags are not set up yet right?

ekohl:
No, it always occurs. Pretty sure this is also an issue in nightly. As proof: http://koji.katello.org/koji/buildinfo?buildID=36330 is missing the client tag.

ehelms:
Ahh, this is true for all but EL7 because one build can exist at a time. Duh (thats a self duh). We need to either:

 a) Add code to obal that says 'if build already exists, attempt to tag build in'
 b) Make foreman-client-nightly-rhel7 inherit foreman-nightly-nonscl-rhel7 to pull in the package

Thoughts?

ekohl:
> a) Add code to obal that says 'if build already exists, attempt to tag build in'

This would be the preferred solution since it also helps us with cherry picks. It's also slightly harder.

ekohl:
For 1.21.0-RC1 I released it by hand.

ekohl:
Updated. Now that we've dropped 1.19 this is easier.

PR #888, opened 2018-12-17
Trigger Katello pipeline builds after Foreman

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/katello-master-release.yaml
```diff
@@ -4,6 +4,12 @@
     sandbox: false
     triggers:
       - github
+      - build-result:
```
ekohl: This job already triggers a build via the groovy script

ehelms: Where?

ekohl: https://github.com/theforeman/foreman-infra/blob/4eddf0759e4690b9dfa7b4f4cf339dae07886b74/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/rubygemRPMNightlyRelease.groovy#L50-L56

ekohl: It does mean that every nightly triggers it while I think it's only needed on a Foreman RPM, right?

ehelms: Correct! But given we use a generic job instead of a specific job, I can't decipher the Foreman specific build

ekohl: It looks like the parametrized trigger plugin does allow to filter on parameters:
> Restrict matrix execution to a subset

ehelms: I don't think that does what you think it does from looking around.

ehelms:
I have updated this to drop the rubygem-katello rebuild trigger for the time being and added a trigger to run the rpm-pipeline if the installer builds.

ehelms:
@ekohl Is the worry around use of resources?

ekohl:
Yes, but in the broader sense. If it fails and we do a lot of builds then we spam discourse. This is annoying for developers.

PR #871, opened 2018-11-22
Simplifying our forklift testing pipeline definitions

evgeni:
these are the required infra changes for https://github.com/theforeman/forklift/pull/880

PR #889, opened 2018-12-18
fix the names of katello nightly pipelines

evgeni:


evgeni:
this was missed in https://github.com/theforeman/foreman-infra/pull/871/

PR #890, opened 2018-12-19
Add a pipeline for test_develop

ehelms:
This still needs some work such as triggers setup, and addressing comments. However, I am opening this up as testing of it is going well for comments to replace `test_develop`.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
@@ -1,30 +1,12 @@
-def databaseFile (id) {
-    writeFile(file: 'config/database.yml', text: """
-test:
-  adapter: postgresql
-  database: ${id}-test
-  username: foreman
-  password: foreman
-  host: localhost
-  template: template0
-
-development:
-  adapter: postgresql
-  database: ${id}-development
-  username: foreman
-  password: foreman
-  host: localhost
-  template: template0
-
-production:
-  adapter: postgresql
-  database: ${id}-development
-  username: foreman
-  password: foreman
-  host: localhost
-  template: template0
-""")
-
+def databaseFile(id, database = 'postgresql') {
+    if (database == 'sqlite3') {
+        text = sqliteTemplate()
+    } else if (database == 'mysql') {
+        text = mysqlTemplate(id)
+    } else {
+        text = postgresqlTemplate(id)
+    }
+    writeFile(file: 'config/database.yml', text: text)
 }
 
 def addGem() {
@@ -35,21 +17,90 @@ def addSettings(settings) {
     sh "cp config/settings.yaml.example config/settings.yaml"
 }
 
-def configureDatabase(ruby) {
-    withRVM(['bundle install --jobs=5 --retry=5'], ruby)
-    withRVM(['bundle exec rake db:drop -q || true'], ruby)
-    withRVM(['bundle exec rake db:create -q'], ruby)
-    withRVM(['bundle exec rake db:migrate -q'], ruby)
+def configureDatabase(ruby, name = '') {
+    withRVM(['bundle install --without=development --jobs=5 --retry=5'], ruby, name)
+    withRVM(['bundle exec rake db:drop || true'], ruby, name)
+    withRVM(['bundle exec rake db:create --trace'], ruby, name)
+    withRVM(['bundle exec rake db:migrate --trace'], ruby, name)
 }
 
-def cleanup(ruby) {
+def cleanup(ruby, name = '') {
     try {
 
-        withRVM(['bundle exec rake db:drop DISABLE_DATABASE_ENVIRONMENT_CHECK=true || true'], ruby)
+        withRVM(['bundle exec rake db:drop DISABLE_DATABASE_ENVIRONMENT_CHECK=true || true'], ruby, name)
 
     } finally {
 
-        cleanupRVM(ruby)
+        cleanupRVM(ruby, name)
 
     }
+}
+
+def postgresqlTemplate(id) {
+  return """
+test:
+  adapter: postgresql
+  database: test-${id}-test
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+
+development:
+  adapter: postgresql
+  database: test-${id}-development
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+
+production:
+  adapter: postgresql
+  database: test-${id}-production
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+"""
+}
+
+def mysqlTemplate(id) {
+  return """
+test:
+  adapter: mysql2
+  database: test-${id}-test
+  username: foreman
+  password: foreman
+  host: localhost
+
+development:
+  adapter: mysql2
+  database: test-${id}-development
+  username: foreman
+  password: foreman
+  host: localhost
+
+production:
+  adapter: mysql2
+  database: test-${id}-production
+  username: foreman
+  password: foreman
+  host: localhost
+"""
+}
+
+def sqliteTemplate() {
+  return """
+test:
+  adapter: sqlite3
+  database: db/test.sqlite3
+  pool: 5
+  timeout: 5000
+
+development:
+  adapter: sqlite3
+  database: db/development.sqlite3
+  pool: 5
+  timeout: 5000
+"""
 }
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
@@ -1,0 +1,201 @@
+def commit_hash = ''
+foreman_branch = 'develop'
+
+pipeline {
+    agent { label 'fast' }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Test Matrix') {
+            parallel {
+                stage('ruby-2.5-postgres') {
+                    agent { label 'fast' }
+                    environment {
+                        RUBY = '2.5'
+                        GEMSET = 'ruby-2.5-postgres'
+                    }
+                    stages {
+                        stage("setup-2.5-postgres") {
+                            steps {
+                                git url: 'https://github.com/theforeman/foreman', branch: foreman_branch
+                                script {
+                                    commit_hash = archive_git_hash()
+                                }
+                                configureRVM(env.RUBY, env.GEMSET)
+                                databaseFile(gemset(env.GEMSET))
+                                configureDatabase(env.RUBY, env.GEMSET)
+                                withRVM(['npm install'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("unit-tests-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake jenkins:unit TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("integration-tests-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake jenkins:integration TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("assets-precompile-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake assets:precompile RAILS_ENV=production'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                    }
+                    post {
+                        always {
+                            cleanup(env.RUBY, env.GEMSET)
+                            deleteDir()
+                        }
+                    }
+                }
+                stage('ruby-2.4-postgres') {
+                    agent { label 'fast' }
+                    environment {
+                        RUBY = '2.4'
+                        GEMSET = 'ruby-2.4-postgres'
+                    }
+                    stages {
+                        stage("setup-2.4-postgres") {
+                            steps {
+                                git url: 'https://github.com/theforeman/foreman', branch: foreman_branch
+                                configureRVM(env.RUBY, env.GEMSET)
+                                databaseFile(gemset(env.GEMSET))
+                                configureDatabase(env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("unit-tests-2.4-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake jenkins:unit TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                    }
+                    post {
+                        always {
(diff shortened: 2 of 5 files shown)


ehelms:
This absolutely runs across multiple Jenkins nodes. The `agent` declaration guarantees this:

```
13:40:25 [ruby-2.5-postgres] Running on slave01.aws.theforeman.org in /var/lib/workspace/workspace/foreman-develop-test
[Pipeline] [ruby-2.4-postgres] node
13:40:25 [ruby-2.4-postgres] Running on slave01.aws.theforeman.org in /var/lib/workspace/workspace/foreman-develop-test@2
[Pipeline] [ruby-2.3-postgres] node
13:40:25 [ruby-2.3-postgres] Running on slave01.rackspace.theforeman.org in /var/lib/workspace/workspace/foreman-develop-test
[Pipeline] [ruby-2.5-mysql] node
13:40:25 [ruby-2.5-mysql] Running on slave01.rackspace.theforeman.org in /var/lib/workspace/workspace/foreman-develop-test@2
[Pipeline] [ruby-2.5-sqlite3] node
13:40:25 [ruby-2.5-sqlite3] Running on slave03.osuosl.theforeman.org in /var/lib/workspace/workspace/foreman-develop-test
[Pipeline] [ruby-2.5-postgres] {
```

This is a decent amount of duplication. There are some strategies for removing the duplication that in turn hurt the visual representation on Jenkins. And in some ways, I appreciate some of the duplication due to clearer readability about what is happening. I want to ensure this is easy for folks to be able to decipher what is happening and how ot make changes.

komidore64:
> This absolutely runs across multiple Jenkins nodes. The `agent` declaration guarantees this

Oh awesome!

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -23,41 +69,46 @@ production:
...
+  host: localhost
 
-def configureDatabase(ruby) {
-    withRVM(['bundle install --jobs=5 --retry=5'], ruby)
-    withRVM(['bundle exec rake db:drop -q || true'], ruby)
-    withRVM(['bundle exec rake db:create -q'], ruby)
-    withRVM(['bundle exec rake db:migrate -q'], ruby)
+production:
+  adapter: mysql2
+  database: ${id}-development
```
ekohl: This is `${gemset}-prod` in the old code

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -23,41 +69,46 @@ production:
...
+  password: foreman
+  host: localhost
 
-    if (settings['organizations']) {
-        sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
-    }
-}
+development:
+  adapter: mysql2
+  database: ${id}-development
```
ekohl: This is `${gemset}-dev` in the old code

ekohl:
The integration tests also compile webpack. I don't mind doing it twice, but perhaps we should compile before running tests. When compilation fails you know where to look rather than just the general test run. On the other hand, this is already an issue now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,201 @@
...
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Test Matrix') {
+            parallel {
+                stage('ruby-2.5-postgres') {
```
evgeni: is this really better readable than

```groovy
def matrix = [
  [ruby: '2.5', db: 'postgres', testtargets: 'jenkins:unit jenkins:integration pkg:generate_source', npm: true, assets: true, webpack: true],
  [ruby: '2.4', db: 'postgres', testtargets: 'jenkins:unit', npm: false, assets: false, webpack: false],
  [ruby: '2.3', db: 'postgres', testtargets: 'jenkins:unit', npm: false, assets: false, webpack: false],
  [ruby: '2.5', db: 'mysql',    testtargets: 'jenkins:unit', npm: false, assets: false, webpack: false],
  [ruby: '2.5', db: 'sqlite3',  testtargets: 'jenkins:unit', npm: false, assets: false, webpack: false]
]
```
(and probably we can drop `npm` there and compute it as `npm = assets or webpack`)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,30 +1,13 @@
...
+def databaseFile(id, database = 'postgresql') {
+    if (database == 'sqlite3') {
+        text = sqliteTemplate()
+    } else if (database == 'mysql') {
+        text = mysqlTemplate(id)
+    } else {
+        text = postgresqlTemplate(id)
+    }
+    writeFile(file: 'config/database.yml', text: text)
+    sh "cat config/database.yml"
```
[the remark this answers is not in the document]
ehelms: This was simply for debugging. I will drop it.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -43,21 +26,90 @@ def addSettings(settings) {
...
 
-        cleanupRVM(ruby)
+        cleanupRVM(ruby, name)
 
     }
 }
+
+def postgresqlTemplate(id) {
+  return """
+test:
```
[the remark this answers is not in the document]
ehelms: We could... but this is quite readable wouldn't you say?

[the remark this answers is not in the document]
evgeni: It is, yeah. Just throwing ideas :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,201 @@
...
+                            }
+                        }
+                        stage("npm-2.5-postgres") {
+                            steps {
+                                withRVM(['npm install'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("unit-tests-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake jenkins:unit jenkins:integration pkg:generate_source TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
```
tbrisker: Would this be a good opportunity to split away the integration tests from the unit tests? right now we have to wait for the slowest test to finish, which is always the integration tests since they run a lot longer - if we split them and run them in parallel, we could get results faster.

ehelms: I could give that a test while I await further reviews.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
evgeni: If my suggestion makes the visual representation ugly, I am good with the current approach.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: I made that change, and I can't say how much faster it made this pipeline but genreally speaking from what I can see: 

 * test_develop: avg of 1hr 10 minutes
 * foreman-develop-test: avg of 50 minutes

So I am seeing roughly 20 minute gain on this job.

tbrisker: that seems inline with the timings of the current tests, usually mysql and sqlite (which don't run integration tests) take around 50min and pg takes around 1:10. We can also get rid of any steps related to npm installation and asset compilation in other tests so we might see even a bit of extra improvement there (and save bandwidth on downloading npm packages).

ehelms:
I've pushed some updates to it based on discussion.

ehelms:
What else would folks like to see here?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,199 @@
...
+                                withRVM(['bundle exec rake jenkins:integration TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("assets-precompile-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake assets:precompile RAILS_ENV=production'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                    }
+                    post {
```
ekohl: Does this mean we generate the source if this particular stage (ruby-2.5-postgres) passes? In the past we only built it if *all* stages* passed. Though I guess this is also covered by not continuing to the packaging stage.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,199 @@
...
+                stage('ruby-2.5-postgres') {
+                    agent { label 'fast' }
+                    environment {
+                        RUBY = '2.5'
+                        GEMSET = 'ruby-2.5-postgres'
+                    }
+                    stages {
+                        stage("setup-2.5-postgres") {
+                            steps {
+                                git url: 'https://github.com/theforeman/foreman', branch: 'develop'
```
ekohl: We do repeat the branch and URL quite a few times. How about variables? Since we copy the pipeline on branching it'd make that part easier.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: That is correct.  I was trying to avoid having to create a brand new environment just to generate the source. And the current job generates the source for *all* job configurations (https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh#L44) but it only *archives* the source if all passed. However, workspaces play a role here. As you said, if these stages fail, the packaging won't kickoff.

Although, in this current design you did get me thinking that this may not work. As the packaging jobs are kicked off mid-pipeline and those packaging jobs need lastSuccessfulBuild which could have a race condition in current form. In @komidore64 proposal, a new packaging job would generate the source and avoid this. So perhaps we need to get that job online first.

ekohl: Looking at https://github.com/theforeman/foreman-packaging/blob/9293c65d55c59eec7a1a58b762f98ca48d12bb2f/rel-eng/custom/custom.py#L124-L130 I also wonder if this would break an assumption. Perhaps not because runs might not even show up in a pipeline build.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: Agreed on branch since that can be configured. The URL I am reticent since that shouldn't really change and this is just as or more readable than a variable.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -43,21 +25,90 @@ def addSettings(settings) {
...
+  adapter: postgresql
+  database: test-${id}-development
+  username: foreman
+  password: foreman
+  host: localhost
+  template: template0
+
+production:
+  adapter: postgresql
+  database: test-${id}-development
```
ekohl: Should this be production?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,199 @@
...
+                    propagate: false,
+                    wait: false,
+                    parameters: [
+                        string(name: 'project', value: 'foreman'),
+                        string(name: 'jenkins_job', value: env.JOB_NAME),
+                        string(name: 'jenkins_job_id', value: env.BUILD_ID)
+                    ]
+                )
+                build(
+                    job: 'release_nightly_build_rpm',
```
ekohl: We should have a dedicated release pipeline for this after JJB has applied the latest config (foreman-develop-release IIRC).

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -0,0 +1,199 @@
...
+                            steps {
+                                withRVM(['bundle exec rake jenkins:unit TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("integration-tests-2.5-postgres") {
+                            steps {
+                                withRVM(['bundle exec rake jenkins:integration TESTOPTS="-v" --trace'], env.RUBY, env.GEMSET)
+                            }
+                        }
+                        stage("assets-precompile-2.5-postgres") {
```
ekohl: IIRC `jenkins:integration` also calls `assets:precompile` so it might be good to swap the order (setup -> unit-tests -> assets-precompile -> integration-tests)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ekohl: I was wrong and that job doesn't exist. Disregard my comment.

ehelms: Will be once https://github.com/theforeman/foreman-infra/pull/923 is added,  I will update this job to reflect that.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: These are all in parallel and thus independent, ordering doesn't matter.

ehelms:
Also updated to reflect pointing at new foreman-develop-release job.

ehelms:
@tbrisker @ekohl When you get time, another final round of review would be appreciated here.

ehelms:
Merging as this doesn't remove test_develop and a little redundancy for a day or two only helps.

PR #893, opened 2018-12-25
Run test_katello seed in test env

tbrisker:
Otherwise seeds fail when trying to connect the default organization to
candlepin.

Also remove code for supporting katello<3.1 that is no longer needed.

This will hopefully fix core's katello test that still don't use the newer pipelines, similar to the fix in #892 for the pipeline.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_katello.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_katello.sh
@@ -73,14 +73,10 @@ bundle exec rake db:migrate --trace
 bundle exec rake jenkins:katello TESTOPTS="-v" --trace
 
 # Run the DB seeds to verify they work
-# Don't run DB seeds if the version of katello is less than 3.1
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_katello.sh
```diff
@@ -73,14 +73,10 @@ bundle exec rake db:migrate --trace
...
 # Run the DB seeds to verify they work
-# Don't run DB seeds if the version of katello is less than 3.1
-VERSION=$(grep -Po "(\d+\.)+\d+" ${PLUGIN_ROOT}/lib/katello/version.rb)
-
-if [ $(echo ${VERSION}$'\n3.1.0' | sort --version-sort --reverse | head -n1) != '3.1.0' ]; then
-  bundle exec rake db:drop || true
-  bundle exec rake db:create db:migrate --trace
-  bundle exec rake db:seed --trace
-fi
+bundle exec rake db:drop RAILS_ENV=test || true
```
ekohl: Would it be easier to `export RAILS_ENV=test` at the start?

PR #896, opened 2019-01-03
update RVM and gem version

mmoll:


mmoll:
this is needed to enable Ruby 2.6

PR #897, opened 2019-01-05
handle new RVM key

mmoll:


mmoll:
see https://github.com/maestrodev/puppet-rvm/issues/145

PR #898, opened 2019-01-06
Remove organization/location enabling in CI

tbrisker:
Organizations an Locations are now always enabled, no need to manually
enable them in the ci pipelines.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
@@ -33,14 +33,6 @@ def addGem() {
 
 def addSettings(settings) {
     sh "cp config/settings.yaml.example config/settings.yaml"
-
-    if (settings['locations']) {
-        sh "sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml"
-    }
-
-    if (settings['organizations']) {
-        sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
-    }
 }
 
 def configureDatabase(ruby) {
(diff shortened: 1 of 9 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -33,14 +33,6 @@ def addGem() {
...
 def addSettings(settings) {
     sh "cp config/settings.yaml.example config/settings.yaml"
-
-    if (settings['locations']) {
-        sh "sed -i 's/:locations_enabled: false/:locations_enabled: true/' config/settings.yaml"
-    }
-
-    if (settings['organizations']) {
-        sh "sed -i 's/:organizations_enabled: false/:organizations_enabled: true/' config/settings.yaml"
-    }
```
ehelms: The trouble is, this code is not isolated to nightly. That is, for example, this code is used on Katello branch PRs against older Foreman code. So we either leave this around for a few releases, or try to version lock it.

ehelms: The above is true for all places this is used.

tbrisker: only we never actually use this file when running tests thanks to your commit from 3 years ago https://github.com/theforeman/foreman/pull/2528 :smile_cat:

tbrisker: I think there may actually be on case where it does work, which is bats testing - i think in that case we run foreman with production environment rather than tests. That may also be the case for proxy tests but i'm not sure.

ehelms: The bats test shouldn't be using this groovy code at all. I am fine leaving it for now if we are not sure.

PR #899, opened 2019-01-07
Fix tfm-ror52 missing RVM and refactor push_rpms

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
```diff
@@ -1,5 +1,9 @@
-void push_rpms(repo_type, version, distro) {
-    withRVM(["cap yum repo:sync -S overwrite=true -S merge=false -S repo_source=foreman-${repo_type}-${version}/${distro} -S repo_dest=${repo_type}/${version}/${distro}"])
+void push_foreman_rpms(repo_type, version, distro) {
+    push_rpms("foreman-${repo_type}-${version}", repo_type, version, 'el7')
```
ekohl: Should `'el7'` be `distro` here?

ehelms: solid catch, updated

PR #900, opened 2019-01-08
Prepare for 1.21 manual

tbrisker:
This should only be merged once we have the 1.21 manual update in theforeman/theforeman.org prepared.


tbrisker:
https://github.com/theforeman/theforeman.org/pull/1276 is ready, both should be merged together

PR #901, opened 2019-01-08
rely on jekins discard settings for katello

jlsherrill:


jlsherrill:
15 to keep doesn't seem to last even a day most of the time

PR #907, opened 2019-01-11
Disable polling on some git repositories

ehelms:
According to the internet, this should disable polling for git pushes. IN theory that should solve our multi-triggered jobs issue.


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/obal.groovy
```diff
@@ -19,6 +19,7 @@ def setup_obal() {
     dir("${env.WORKSPACE}/obal") {
         checkout([
             $class : 'GitSCM',
+						poll: false,
```
ekohl: Whitespace :)

PR #909, opened 2019-01-13
Update stable branch jobs for 1.21

tbrisker:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_proxy_1_21_stable.yaml
```diff
@@ -1,12 +1,12 @@
 - job:
```
ekohl: In develop we also test this one with 2.5.

tbrisker: added.

PR #911, opened 2019-01-16
only add job link for the currently running job

evgeni:
when a pipeline runs multiple steps in parallel on the same node, each
step gets an unique workspace (jobname@XXX for all but the first run).
this means that when we write the remote job details to the workspace,
we might not overwrite any pre-existing files in there.

this leads to jobs with multiple links like seen in [1], which is
confusing.
let's be more explicit and only add one link at at time to the
description.

[1] https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/659/

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy
@@ -59,7 +59,7 @@ pipeline {
                     post {
                         always {
                             script {
-                                set_job_build_description()
+                                set_job_build_description("foreman-katello-nightly-test")
                             }
                         }
                     }
@@ -86,7 +86,7 @@ pipeline {
                     post {
                         always {
                             script {
-                                set_job_build_description()
+                                set_job_build_description("foreman-katello-upgrade-nightly-test")
                             }
                         }
                     }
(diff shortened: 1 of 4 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy
```diff
@@ -59,7 +59,7 @@ pipeline {
                     post {
                         always {
                             script {
-                                set_job_build_description()
+                                set_job_build_description("foreman-katello-nightly-test")
```
ehelms: Maybe set this as a environment variable to reference in the steps and post given we now have it in 3 places?

evgeni: like this?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/ci_centos_integration_test.groovy
```diff
@@ -42,7 +42,8 @@ pipeline {
             post {
                 always {
                     script {
-                        set_job_build_description()
+                        set_job_build_description("foreman-ci-centos-simple-test")
+                        set_job_build_description("foreman-ci-centos-simple-test-2")
```
ehelms: Mind doing it here too so this job mimics the pattern of 'best practices' ?

PR #915, opened 2019-01-18
Disable sudo for the jenkins user on non-Debian

ekohl:


ekohl:
Looking a bit closer I suspect most of this is because of pbuilder and chown

ekohl:
Updated to maintain sudo on Debian, but disable it elsewhere. Should be a first step.

mmoll:
let's try it...

mmoll:
merged, bedankt @ekohl!

PR #918, opened 2019-01-23
Explicitly clean the installer

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -80,6 +80,10 @@ void build_tarball(project, version, ruby_ver) {
             rake = "bundle exec rake"
         }
 
+        if (project == 'foreman-installer') {
+                withRVM(["${rake} clean"], ruby_ver, project)
```
ehelms: Indentation seems off

ekohl: Updated and added a `~/.vim/ftplugin/groovy.vim`:
```vim
set tabstop=4
set shiftwidth=4
set expandtab
set autoindent
set smartindent
```

PR #919, opened 2019-01-24
Disable concurrent installer builds

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-installer-develop-release.yaml
```diff
@@ -1,7 +1,7 @@
 - job:
     name: foreman-installer-develop-release
     project-type: pipeline
-    concurrent: true
+    concurrent: false
```
komidore64: Concurrency defaults to `false`, so you could just remove this line entirely.

ekohl: Looks like we do have it in two other places. Shall I remove those too?

komidore64: Well, now that I commented I also feel it can't hurt to be explicit. I've run into one or two weird situations with changing a boolean value in JJB from non-default to default (by removing the field entirely) and then the update action just sees "there's nothing to change on this field" and leaves it how it's currently set.

I take back my previous comment. Let's leave in `concurrent: false`.

PR #921, opened 2019-01-25
 use git diff instead of find to find the old and new specfiles

evgeni:
* introduce more wrappers for git diff and finding of changed files
* use git diff instead of find to find the old and new specfiles
  using find might find the wrong spec file as we sometimes have multiple packages with the same spec name
* move old vs new path check to the begining of the test

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
@@ -6,8 +6,20 @@ def update_build_description_from_packages(packages_to_build) {
     currentBuild.description = build_description
 }
 
+def diff_filter(range, filter, path) {
+    return sh(returnStdout: true, script: "git diff ${range} --name-only --diff-filter=${filter} -- '${path}'").trim()
+}
+
+def find_added_or_changed_files(diff_range, path) {
+    return diff_filter(diff_range, 'ACMRTUXB', path)
+}
+
+def find_deleted_files(diff_range, path) {
+    return diff_filter(diff_range, 'D', path)
+}
+
 def find_changed_packages(diff_range) {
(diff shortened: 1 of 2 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/rpm_packaging.groovy
```diff
@@ -6,8 +6,20 @@ def update_build_description_from_packages(packages_to_build) {
     currentBuild.description = build_description
 }
 
+def diff_filter(range, filter, path) {
+    return sh(returnStdout: true, script: "git diff ${range} --name-only --diff-filter=${filter} -- '${path}'").trim()
```
ehelms: One of these days, I think we should make this call structure a method since I have seen us use it quite a few times.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/rpm_packaging.groovy
```diff
@@ -65,22 +65,25 @@ pipeline {
 
                     for(int i = 0; i < packages_to_build.size(); i++) {
                         package_name = packages_to_build[i]
+                        spec_pattern = "packages/**/${package_name}.spec"
 
-                        sh "git checkout origin/${env.ghprbTargetBranch}"
+                        old_spec_path = find_deleted_files("origin/${env.ghprbTargetBranch}", spec_pattern)
+                        if (! old_spec_path) {
+                            old_spec_path = find_added_or_changed_files("origin/${env.ghprbTargetBranch}", spec_pattern)
+                        }
```
ehelms: Looks like you could drop the `if (old_spec_path) {` that follows this line and change it to an `else`

evgeni: No? I want:
```
old_spec_path = find_delted_files() || find_added_or_changed_files()
if (old_spec_path) …
```

Your suggestion would drop the `|| find_added_or_deleted` part?

PR #930, opened 2019-02-06
Drop test_develop in favor of foreman-develop-test

ehelms:


ehelms:
Holding on merge to ensure a few green runs of foreman-develop-test

ehelms:
The integration tests seems to be rather unstable if you look at the job and history of jobs:

http://ci.theforeman.org/blue/organizations/jenkins/foreman-develop-test/detail/foreman-develop-test/9

@tbrisker any thoughts here? test_develop doesn't seem to hit these instabilities, but this new job seems to uncover them more readily. I am unclear if this is a CI problem or an integration tests problem.

tbrisker:
This may be a bit of both. Perhaps because the tests run faster now or more in parallel we are hitting issues we didn't before; for example, the integration test that's failing seems to use a static uuid, perhaps changing it to a random one will fix that failure - https://github.com/theforeman/foreman/blob/develop/test/integration/host_js_test.rb#L529

tbrisker:
opened https://projects.theforeman.org/issues/26022 to fix at least the integration one

ehelms:
@tbrisker @ekohl 5 successful builds in a row....http://ci.theforeman.org/job/foreman-develop-test/

How are y'all feelin about merging this?

tbrisker:
Looks like there are still quite a lot of transient failures... 
is there a way of easily seeing which tests failed in the pipeline? the only way i could find is opening the failed step and scrolling through thousands of lines of output.

ekohl:
We can use a test publisher but currently that's not configured. The old job had a JUnit publisher on `jenkins/reports/unit/*.xml`.

ekohl:
> We can use a test publisher but currently that's not configured. The old job had a JUnit publisher on `jenkins/reports/unit/*.xml`.

https://github.com/theforeman/foreman-infra/pull/998 implements that.

PR #933, opened 2019-02-08
Add hammer-cli-develop-release job

ehelms:


ehelms:
Currently failing -- http://ci.theforeman.org/job/hammer-cli-master-release/1/consoleFull

ehelms:
Success -- http://ci.theforeman.org/job/hammer-cli-master-release/3/console

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
```diff
@@ -0,0 +1,6 @@
+def git_url = 'https://github.com/theforeman/hammer-cli.git'
+def git_ref = 'master'
+def package_name = 'rubygem-hammer_cli'
```
ekohl: I'm not too happy about this naming because on Debian it's named `ruby-hammer-cli`.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
```diff
@@ -0,0 +1,6 @@
...
+def project_name = 'hammer-cli'
+def build_rpm = true
+def build_deb = false
```
ekohl: We do ship hammer in debs, but we don't build nightly debs for hammer?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli.yaml
```diff
@@ -35,10 +35,5 @@
...
           health-scale-factor: 1.0
       - ircbot_freenode
-      - trigger-parameterized-builds:
-        - project: release_nightly_build_rpm
-          condition: SUCCESS
-          predefined-parameters: |
-            project=packages/foreman/rubygem-hammer_cli
-            jenkins_job=${JOB_NAME}
-            jenkins_job_id=${BUILD_NUMBER}
+      - trigger-release-job:
```
ekohl: If you now trigger the release job, shouldn't that take care of the archiving? That'd mean you could remove the archive publisher (and need a matching `foreman-packaging` manifest change PR).

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
evgeni: While the `package_name` isn't used in the Debian branch of the job, I'd also prefer to call this `rpm_package_name`

ekohl: It's also not really the rpm package name since that could be `tfm-rubygem-hammer_cli`. `obal_name` or `obal_manifest_name` perhaps?

ehelms: `packaging_folder_name` ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
ehelms: Correct -- https://github.com/theforeman/foreman-infra/pull/933/files#diff-fd91d137d8bc077e688a82182d8e3e29L38

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli.yaml
ehelms: Updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
ekohl: folder is typically a windows name where unix uses directory. `obal_package_name` perhaps? Currently this line needs an update anyway

ehelms:
Updated the parameter name

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
```diff
@@ -0,0 +1,6 @@
+def git_url = 'https://github.com/theforeman/hammer-cli.git'
```
ekohl: Side note: I've been wondering if we could use the github URL from the job properties here but for now I'm fine with hardcoding this.

ehelms:
Fair -- are you requesting the test job get re-factored alongside this? before this? or quickly there after this?

ekohl:
It looks like a rather straight forward conversion because the hammer job isn't that complicated. I'd prefer alongside this.

ehelms:
Added a hammer-cli-master-test job:

http://ci.theforeman.org/blue/organizations/jenkins/hammer-cli-master-test/detail/hammer-cli-master-test/3/pipeline/

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_hammer_cli.yaml
```diff
@@ -28,17 +28,10 @@
     builders:
```
ekohl: Do we want to drop this file already or do you keep them in parallel so we can monitor it for a bit?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/hammer-cli-master-release.groovy
komidore64: i thought hard about this too a few months back, then decided it wasn't worth all the clever string manipulation to avoid adding two extra variables.

ekohl: Out of interest: what string manipulation? You can `git clone https://github.com/theforeman/hammer-cli` so the URL is exactly the same.

ehelms:
Good call, I did miss the trigger. Added the trigger and dropped the old job.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -55,7 +55,7 @@ pipeline {
                         dir('foreman-packaging') {
                             obal(
                                 action: 'nightly',
-                                packages: project_name,
+                                packages: obal_package_name ?: project_name,
```
ekohl: This broke foreman-selinux: `groovy.lang.MissingPropertyException: No such property: obal_package_name for class: WorkflowScript`

PR #940, opened 2019-02-12
Correct the Debian PR tester jobs

ekohl:
These are the nightly jobs that are now responsible for building tarballs.


mmoll:
merged, bedankt @ekohl!

PR #941, opened 2019-02-13
don't serve freight's .refs folder via http and rsync

evgeni:
freight uses a .refs folder underneath `dists/<name>-<time>/` to keep
track of "still referenced" debs for housekeeping of `pool/`.
As the timestamp changes every time freight runs, simple mirroring tools
(like wget or rsync) will think these files are new and re-download them
every time, producing more traffic than needed.
These files are internal to freight and not used or referenced in the
repository metadata, so no real Debian client will ever try to access
them. Thus we can just safely exlude them from being transferred save a
lot of traffic.

This commit should save us about $600-700 of traffic each month.

### DIFF
--- a/puppet/modules/freight/manifests/user.pp
+++ b/puppet/modules/freight/manifests/user.pp
@@ -68,6 +68,10 @@ define freight::user (
       headers => 'Set Cache-Control "public, max-age=120"',
     },
     {
+      path    => "${webdir}/dists/*/.refs/",
+      deny    => 'from all',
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
:+1: looks good to me. @ekohl?

lzap:
Heh you deserve 10% of the annual price then, thanks :-)

mmoll:
merged, спасибо Женя!

PR #949, opened 2019-02-14
Rely on releasers from package_manifest when nightly releasing

ehelms:


komidore64:
:-1:

The end goal was to remove nightly specific content from `package_manifest.yaml`.

ekohl:
@komidore64 shouldn't this be redundant since we set `nightly_sourcefiles` and `nightly_githash`?

ehelms:
@komidore64 I assumed that this would remove nightly specific information as it would just use whatever releaser is configured for a given package within package_manifest ? For example, instead of using `koji-foreman-jenkins`, and configured it specifically for that, it would use `koji-foreman` (which it's already coded here todo).

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman-installer-develop-release.groovy
```diff
@@ -4,3 +4,5 @@ def project_name = 'foreman-installer'
 def obal_package_name = 'foreman-installer'
 def build_rpm = true
 def build_deb = true
+def source_type = 'rake'
```
ekohl: At some point I'd like to optimize here. In Foreman we can also use `rake -f Rakefile.dist pkg:generate_source` and that doesn't need a bundle install which makes it much faster. I'm thinking of introducing a `bundle-rake` source type vs `rake` but this is a future optimization.

komidore64:
> I assumed that this would remove nightly specific information as it would just use whatever releaser is configured for a given package within package_manifest ?

Yeah, okay that is correct. I _think_ I added that [`?:`](https://github.com/theforeman/foreman-infra/pull/949/files#diff-275e3819b8311b4a9a3247faf99f07c6L60) because one or two packages broke, so this was a stopgap fix.

PR #955, opened 2019-03-07
Modify foreman develop release timeout

tbrisker:
Deb releases sometimes take more than 2 hours, causing the pipeline to
timeout.


tbrisker:
For example: http://ci.theforeman.org/job/foreman-develop-release/142/

PR #956, opened 2019-03-07
adjust jenkins for new pipeline definitions

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/luna/nightly.groovy
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                }
+            }
+        }
+    }
+    post {
+        failure {
+            emailext(
+                subject: "${env.JOB_NAME} ${env.BUILD_ID} failed",
+                to: 'evgeni@redhat.com',
```
evgeni: will be switched to `ci@community` as soon as this is working fine

PR #959, opened 2019-03-11
Add Ruby 2.5 & Puppet 6 to the Kafo PR testing

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
```diff
@@ -25,16 +25,18 @@
...
           values:
           - '4.10'
           - '5.0'
+          - '6.0'
     execution-strategy:
-      combination-filter: '!(ruby ==~ /^2\.[0-3]/ && puppet ==~ /^5\..*/)'
+      combination-filter: '!(ruby ==~ /^2\.[0-3]/ && puppet ==~ /^[56]\..*/)'
       touchstone:
-        expr: 'ruby=="2.1" && puppet=="4.10"'
+        expr: 'ruby=="2.4" && puppet=="6.0"'
```
mmoll: I think this should be either (Ruby 2.4 and Puppet 5.x) or (Ruby 2.5 and Puppet 6.x) to match the AIO Rubies.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
```diff
@@ -25,16 +25,18 @@
           - '2.2'
           - '2.3'
           - '2.4'
+          - '2.5'
       - axis:
           type: user-defined
           name: puppet
           values:
           - '4.10'
```
mmoll: this should get deleted now?

ekohl: We still support Puppet 4.10 in Kafo 3. I'm also not sure how well the Puppet 5 gem works on Ruby 2.0 and I'm not sure how easily we can test cross Ruby versions.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_pull_request.yaml
ekohl: Yes, I wrote this before I added Ruby 2.5.

mmoll:
merged, bedankt @ekohl!

PR #961, opened 2019-03-12
Seperate nightly arm builds and make them non-blocking

tbrisker:
Following the discussion in https://community.theforeman.org/t/dropping-arm-packages/12941, I'm opening this as a first stage. If we agree to drop arm from nightlies completely we can get rid of the second builder.


mmoll:
closing in favour of GH-962

PR #962, opened 2019-03-12
Drop arm builds from nightlies

tbrisker:
This is an alternative to GH-961, dropping arm completely from nightlies.


mmoll:
meged, thanks @tbrisker!

PR #965, opened 2019-03-17
runs UI related tests in parallel

ohadlevy:


ohadlevy:
@ekohl thanks - updated.

ohadlevy:
I've added another commit to address test pr's as well, thanks.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -23,6 +23,13 @@ bundle install --without=development --jobs=5 --retry=5
 # Rubocop
 bundle exec rake rubocop
 
+# setup UI testing
+
+if [ $database -eq 'postgres-integerations']; then
```
ekohl:
```suggestion
if [ $database -eq 'postgres-integrations']; then
```

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop_pr_core.yaml
```diff
@@ -26,6 +26,7 @@
           type: user-defined
           name: database
           values:
+            - postgresql-integerations
```
ekohl:
```suggestion
            - postgresql-integrations
```

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop_pr_core.yaml
```diff
@@ -36,7 +37,7 @@
             - fast
     execution-strategy:
       combination-filter: >
-        ((database == 'postgresql') && ((ruby == '2.3') || (ruby == '2.4') || (ruby == '2.5'))) || ((database == 'mysql') && (ruby == '2.5')) || ((database == 'sqlite3') && (ruby == '2.5'))
+        ((database == 'postgresql') && ((ruby == '2.3') || (ruby == '2.4') || (ruby == '2.5'))) || ((database == 'mysql') && (ruby == '2.5')) || ((database == 'sqlite3') && (ruby == '2.5')) || ((database == 'postgres-integerations') && (ruby == '2.5'))
```
ekohl:
```suggestion
        ((database == 'postgresql') && ((ruby == '2.3') || (ruby == '2.4') || (ruby == '2.5'))) || ((database == 'mysql') && (ruby == '2.5')) || ((database == 'sqlite3') && (ruby == '2.5')) || ((database == 'postgres-integrations') && (ruby == '2.5'))
```

ohadlevy:
@ekohl thanks, typo fixed.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -29,14 +29,40 @@ pipeline {
...
                         }
+                    }
+                    post {
+                        always {
+                            cleanup(env.RUBY, env.GEMSET)
+                            deleteDir()
+                        }
+                    }
+                }
+                stage('ruby-2.5-postgres-integerations') {
```
ekohl:
```suggestion
                stage('ruby-2.5-postgres-integrations') {
```

ohadlevy:
@ekohl rebased again.

ohadlevy:
@ekohl any idea how to test this?

ekohl:
AFAIK there's no way of testing other than to deploy.

ohadlevy:
@ekohl ok anything else that needs changing? /cc @evgeni ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -29,14 +29,40 @@ pipeline {
...
+                    agent { label 'fast' }
+                    environment {
+                        RUBY_VER = '2.5'
+                        GEMSET = 'ruby-2.5-postgres'
+                    }
+                    stages {
+                        stage("setup-2.5-postgres") {
+                            steps {
+                                git url: 'https://github.com/theforeman/foreman', branch: foreman_branch
+                                script {
```
evgeni: I don't think this step is required, we already get the `commit_hash` in the regular PostgreSQL+Ruby2.5 run.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
```diff
@@ -29,14 +29,40 @@ pipeline {
...
+                            cleanup(env.RUBY_VER, env.GEMSET)
+                            deleteDir()
+                        }
+                    }
+                }
+                stage('ruby-2.5-postgres-integrations') {
+                    agent { label 'fast' }
+                    environment {
+                        RUBY_VER = '2.5'
+                        GEMSET = 'ruby-2.5-postgres'
```
evgeni: Should this be a different `GEMSET` name than in the regular PostgreSQL+Ruby2.5 run?

ohadlevy:
But it might run on another slave..?

ohadlevy:
I don't think so, the aim is simply to run the integration tests in
parallel vs serial today.

evgeni:
Two comments, but looks good to me.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ohadlevy: But it might run on another slave..?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ohadlevy: I don't think so, the aim is simply to run the integration tests in
parallel vs serial today.

ekohl: I'm wondering about concurrency issues. What happens if one finishes and cleans up the gemset?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: Doesn't matter because the commit_hash is an input parameter to the whole job.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ehelms: I think it should to prevent clashing, or one parallel deleting the gemset before the other is done with it.

ohadlevy: oh, I was not sure this was actually shared and deleted each time,  I thought it will save some time by not downloading it again.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ohadlevy: should i remove the setup stage then?

ohadlevy:
updated the gemset and squashed the commits.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/test_foreman.groovy
ohadlevy: removed the script section

ohadlevy:
anything else needed for this PR? thanks.

ohadlevy:
I would agree to merge it now :-) if it doesn't work we can always revert.

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_develop.sh
```diff
@@ -33,7 +40,7 @@ bundle exec rake rubocop
 ) > $APP_ROOT/config/database.yml
 
 # we need to install node modules for integration tests (which only run on postgresql)
-if [ ${database} = postgresql -a -e "$APP_ROOT/package.json" ]; then
+if [ ${UI} -eq true ]; then
```
ekohl: I suspect this will break when it's unset:
```
$ if [ ${UNSET_VAR} -eq true ] ; then echo test; fi
bash: [: -eq: unary operator expected
```

ohadlevy: @ekohl thanks, updated all conditions

ohadlevy:
not sure why the tests fail, but it seems not related to the PR? anything that I can do to push this forward?

ohadlevy:
thanks! is it just me or the tests are not triggered?

tbrisker:
Looks like foreman-develop-test now does work properly and finishes in about 34-37 minutes instead of 45-70 before, but test_develop_pr_core doesn't start the postgres-integration tests for some reason.

ekohl:
It's probably a good idea to refactor the PR testing to a pipeline so we can share code rather than trying to invest more time into the old way.

tbrisker:
found it - theforeman/foreman-infra#999

PR #971, opened 2019-03-20
Add a timeout to JJB cronjobs

ekohl:
Sometimes this job hangs and needs to be killed or it'll run forever. Because it first takes a lock, this then blocks other jobs from updating.


komidore64:
I'm unfamiliar with `timeout`. Does it function as a simple timer or does it check for process activity?

ekohl:
It's a simple timer. https://www.gnu.org/software/coreutils/manual/html_node/timeout-invocation.html#timeout-invocation.

I'm applying this because we very regularly saw a JJB job running for a few days. That kept the lock and prevented other updates. Puppet has a built in timeout, but the cronjob doesn't.

mmoll:
merged, bedankt @ekohl!

PR #978, opened 2019-03-21
run debian nightly tests on ci.centos.org

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-nightly-deb-pipeline.yaml
```diff
@@ -0,0 +1,11 @@
+- job:
+    name: foreman-nightly-deb-pipeline
+    project-type: pipeline
+    triggers:
+      - timed: 'H 07 * * 1-5'
+    dsl:
+      !include-raw:
+        - pipelines/release/foreman-deb.groovy
```
ekohl: Doesn't this also need `release.groovy` for the deb pushing?

evgeni: I could have bet it has. Fixing.

PR #981, opened 2019-03-22
Use a SSH secret for Katello RPM pushing

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
```diff
@@ -12,6 +12,11 @@ void push_rpms_direct(repo_source, repo_dest, overwrite = true, merge = false) {
     }
 }
 
+void push_rpms_katello(version) {
+    sshagent(['katello-fedorapeople']) {
+        sh "ssh katelloproject@fedorapeople.org \"cd /project/katello/bin && sh rsync-repos-from-koji ${version}\""
+    }
+
```
evgeni:
```suggestion
}
```

ekohl: Just saw it too.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
```diff
@@ -12,6 +12,12 @@ void push_rpms_direct(repo_source, repo_dest, overwrite = true, merge = false) {
     }
 }
 
+void push_rpms_katello(version) {
+    sshagent(['katello-fedorapeople']) {
+        sh "ssh katelloproject@fedorapeople.org \"cd /project/katello/bin && sh rsync-repos-from-koji ${version}\""
```
evgeni:
```suggestion
        sh "ssh katelloproject@fedorapeople.org 'cd /project/katello/bin && sh rsync-repos-from-koji ${version}'"
```

ekohl: Why is this needed? I'm not quite sure where the variable interpolation happens.

evgeni: `version` is a groovy var here, groovy replaces vars in double quotes, so what the shell sees doesn't have any vars and does not need to be in double quotes.

evgeni: also, it's not *needed*, I just thought not having to escape the " is nice

ekohl: Updated

PR #987, opened 2019-03-23
Fix pipelines

ekohl:
* 3b7f4b57 Correct tfmROR51 repo location
* 2887651e Correct repo destination for foreman nightly
* 63d93301 Fix client pipelines. In the current version of Jenkins and plugins a stage doesn't guarantee it's ran on the same agent. That means we were cloning and preparing one agent and then running it on the next.


mmoll:
test failures related?

ekohl:
Yes, looking into it.

ekohl:
:green_heart:

PR #989, opened 2019-03-25
Add a runCicoJob function

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.10.groovy
```diff
@@ -39,42 +39,14 @@ pipeline {
...
-                        }
+                        runCicoJob("foreman-katello-3.10-test")
                     }
                 }
                 stage('Upgrade test') {
-                    agent { label 'el' }
                     steps {
-
-                        git_clone_foreman_infra()
                         sleep(5) //See https://bugs.centos.org/view.php?id=14920
```
ekohl: This may break because there's no agent to run on. Because of the duplication I'm going to move it inside the function.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -25,3 +25,32 @@ def set_job_build_description(job_name) {
 
     currentBuild.description += build_description
 }
+
+def runCicoJob(job_name, wait = false) {
```
evgeni: I think `wait` should be an integer, defaulting to 0

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
```diff
@@ -10,100 +10,12 @@ pipeline {
...
-                                set_job_build_description("${env.cico_job_name}")
-                            }
-                            deleteDir()
-                        }
-                    }
-                }
+            steps {
+                parallel(
+                    'debian9': { runCicoJob("foreman-nightly-debian9-test") },
+                    'ubuntu1604': { runCicoJob("foreman-nightly-ubuntu1604-test") },
```
evgeni: this needs to wait (was buggy in my code)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
```diff
@@ -10,100 +10,12 @@ pipeline {
...
+                    'ubuntu1804': { runCicoJob("foreman-nightly-ubuntu1804-test") }
```
evgeni: this needs double wait and the reason why I asked for an `int` above

ekohl:
Updated. Managed to trim a few more lines.

ekohl:
The solution I chose uses job counter (0, 1, 2, 3) and multiplies with a fixed number. This should also make it easy to iterate over an array of jobs and use the array index. I didn't go this far here, but it makes it easy to later increase or decrease the wait time.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
```diff
@@ -10,100 +10,12 @@ pipeline {
...
-                            script {
-                                set_job_build_description("${env.cico_job_name}")
-                            }
-                            deleteDir()
-                        }
-                    }
-                }
+            steps {
+                parallel(
+                    'debian9': { runCicoJob("foreman-nightly-debian9-test", 0) },
```
ehelms: This call signature is a little ugly and I think it could be made cleaner in one of two ways:

 1) make a `runCicoJobsInParallel` function that handles the numbering and creation based on a hash
 2) Make this a for loop (#1 includes #2)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
```diff
@@ -10,100 +10,12 @@ pipeline {
...
-                    post {
-                        always {
-                            script {
-                                set_job_build_description("${env.cico_job_name}")
-                            }
-                            deleteDir()
-                        }
-                    }
-                }
+            steps {
```
ehelms: No steps in this case, parallel acts like steps and this should error if you try to run it with both at the top level. However see https://github.com/theforeman/foreman-infra/pull/989/files#r268856449

ekohl: I'm using paralllel() and there it failed on me so I added this again.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
ekohl: I experimented with this. Thought in this PR it was a bit overkill and premature, but https://github.com/theforeman/foreman-infra/pull/992/files#diff-ef3be3111b17e7d2f2296bc04cd28ec8R45

ehelms: That logic wrapped in a function is exactly what I am thinking. Take a look as well at https://github.com/SatelliteQE/robottelo-ci/blob/master/workflows/lib/runPlaybook.groovy#L51

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -25,3 +25,30 @@ def set_job_build_description(job_name) {
...
+                            "jenkins_job_name": "${job_name}",
+                            "jenkins_username": "foreman",
+                            "jenkins_job_link_file": "${env.WORKSPACE}/jobs/${job_name}"
+                        ],
+                        sensitiveExtraVars: ["jenkins_password": "${env.PASSWORD}"]
+                    )
+                }
+            } finally {
+                script {
+                    set_job_build_description("${job_name}")
```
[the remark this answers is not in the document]
evgeni: nope

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
ekohl: I don't see a way to clean it up so it's more readable. Would you mind if we merge it as it is (after some testing) and iterate on it in a later stage?

ehelms: Sure. Here's what I envision just in case:

```
def runCicoJobInParallel(jobs) {

    for (int i = 0; i < jobs.size(); i++) {
        def index = i // fresh variable per iteration; i will be mutated
        branches[jobs[i]['name'] = {
            runCicoJob(jobs[i]['job'], i)
        }
    }

    parallel branches
}
```

```
steps {
    script {
        runCicoJobInParallel([
            ['name': 'ubuntu1804', 'job': 'foreman-nightly-ubuntu1804-test'],
            ['name': 'ubuntu1604', 'job': 'foreman-nightly-ubuntu1604-test']
        ])
    }
}
```

ekohl: The problem is that if you start to add the additional vars to also set the OS it becomes ugly. Maybe that could be solved by defining multiple cico jobs via templates.

Another thought: can we pass in a string template? So the call could be:

```groovy
steps {
    script {
        runCicoJobInParallel('foreman-nightly-{}-test', ['ubuntu1804', 'ubuntu1604'])
    }
}
```

https://stackoverflow.com/questions/45830294/can-groovy-string-interpolation-be-nested suggests this may be possible but not sure if we're overengineering things then.

I think I'll implement your solution since it at least hides the counter from the actual job definition.

ekohl:
Updated. I've introduced 3 functions: runIndividualCicoJob as an internal helper. Then runCicoJob as a wrapper that executes it in the correct environment for a single job. The runCicoJobsInParallel runs multiple jobs in parallel from a single node. This is an improvement in builder slot usage and more efficient since you don't need to clone a git repo multiple times.

ekohl:
https://ci.theforeman.org/job/foreman-nightly-deb-pipeline/21/ runs this new version.

ekohl:
Something I also thought about, but I'll leave that for the future. I didn't know how it would affect display of pipelines for example.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -25,3 +25,58 @@ def set_job_build_description(job_name) {
...
+            } finally {
+                deleteDir()
+            }
+        }
+    }
+}
+
+def runCicoJobsInParallel(jobs) {
+    def branches = [:]
+    for (int i = 0; i < jobs.size(); i++) {
```
evgeni: I think this can be written as

```groovy
    jobs.eachWithIndex { job ->
        branches[job['name']] = {
            runIndividualCicoJob(job['job'], index)
        }
    }
```

PR #991, opened 2019-03-25
Add a test for web::htpasswd

ekohl:
An old branch I had locally but never pushed.


mmoll:
:green_heart: locally

PR #992, opened 2019-03-25
Refactor release pipeline

ekohl:
Includes https://github.com/theforeman/foreman-infra/pull/989 and is untested. Likely has some mistakes.


ehelms:
This implies a special pipeline just for GA releases and not RCs? What is the difference between the two?

ekohl:
Naming is hard. I wanted to separate it from nightly but thought release was too ambiguous.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
```diff
@@ -0,0 +1,91 @@
...
+                always {
+                    deleteDir()
+                }
+            }
+        }
+        stage('Install Test') {
+            steps {
+                def pipelines = [:]
+                def operatingsystems = ['centos7', 'debian9', 'ubuntu1604', 'ubuntu1804']
+                for(int i = 0; i < operatingsystems.size(); i++) {
```
evgeni: can't this be written as:

```groovy
operatingsystems.each { os ->
  pipelines[os] = {}
}
```

ekohl: Maybe, still rather new to groovy.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
```diff
@@ -0,0 +1,93 @@
...
+                }
+            }
+        }
+        stage('Install Test') {
+            steps {
+                def operatingsystems = ['centos7', 'debian9', 'ubuntu1604', 'ubuntu1804']
+                def pipelines = operatingsystems.collect { os ->
+                    [
+                        name: os,
+                        job: 'foreman-release-test',
```
evgeni: this will generate 4 jobs with the same name, which will probably fail

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
```diff
@@ -0,0 +1,93 @@
...
+                        always {
+                            deleteDir()
+                        }
+                    }
+                }
+                stage('Push DEBs') {
+                    agent { label 'debian' }
+                    steps {
+                        parallel(
+                            "stretch": { push_debs_direct('stretch', 'nightly') },
```
evgeni:
```suggestion
                            "stretch": { push_debs_direct('stretch', major_version) },
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
evgeni: also, there is no such job at the moment. should it be in this PR too?

evgeni: for Katello, we have a job per major version (technically, minor, 3.x, same lingo error as in the rest of our jobs :D)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
```diff
@@ -0,0 +1,93 @@
...
+                    "-l el7-base",
+                    "-l el7-updates",
+                    "-l el7-epel",
+                    "-l el7-extras",
+                    "-l el7-scl",
+                    "-l el7-puppet-5"
+                ]
+
+                dir('repoclosure') {
+                    sh command.join(" ")
```
ehelms: Could this instead run `obal repoclosure  foreman-repoclosure-el7` ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman/ga.groovy
```diff
@@ -0,0 +1,93 @@
...
+            }
+            post {
+                always {
+                    deleteDir()
+                }
+            }
+        }
+        stage('Install Test') {
+            steps {
+                def operatingsystems = ['centos7', 'debian9', 'ubuntu1604', 'ubuntu1804']
```
ehelms: Does this work? My gut says this will error unless wrapped in a `script {}` block

ekohl:
@evgeni could you have a look? it's not ready and partly untested. The idea is that we create jobs in a matrix on centos ci, but we do pass the expected version via a parameter so we can verify it's installing the correct version.

evgeni:
```
IOError: Error reading file 'output/foreman-{foreman_version}-release-pipeline/config.xml': failed to load external entity "output/foreman-{foreman_version}-release-pipeline/config.xml"
```

heh, seems you outsmarted my stupid linter script :D

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-release-test.yml
```diff
@@ -0,0 +1,31 @@
+---
+- job-template:
+    name: foreman-{foreman_version}-{distro}-release-test
+    project-type: 'pipeline'
+    sandbox: true
+    parameters:
+      - string:
+          name: expected_version
+          default: 'UNKNOWN'
```
evgeni: you could make the default saner by passing it as `{foreman_version}` or even better define the expected version in the project definition.

see https://github.com/SatelliteQE/robottelo-ci/blob/8d758d2a12b17da0fc23e344fcc1d4747dfe388b/jobs/release/sat-release.yaml#L48-L60 for an example

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release_packages.yaml
```diff
@@ -34,6 +34,7 @@
                   block: true
     publishers:
       - trigger-parameterized-builds:
-        - project: release_mash
+        # TODO This probably doesn't work
```
evgeni: You're probably right, but I am not sure how to make it better.

evgeni: well, actually, we set `current-parameters: true`, so `$major_version` is passed as a param already, so you'd just need to have a generic `foreman-release-pipeline(-trigger)` job here, that triggers the real `foreman-VERSION-release-pipeline` based on a param. when triggering via a Groovy pipeline, you can do fancy vars just fine :)

ekohl: I'm debating if it should trigger at all. Since you need to sign RPMs in between, there's a good chance you need to trigger it anyway. On the other hand, you can often sign RPMs while the (ARM) debs are building and it can continue.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-release-test.yml
ekohl: But we don't know the exact version. Remember that this is in the form of 1.23.0-RC1. That's the whole point of this parameter.

ekohl:
I've tested this on a local jenkins by commenting out the mash + push steps and it correctly kicked off the jobs on centos ci. I'd like to continue testing this on the 1.23 release.

ekohl:
Note that in the end I chose not to trigger the release and leave that up to the release engineer for now. In the future we may end up with automatic package release jobs that would replace `release_packages`.

PR #993, opened 2019-03-25
push nightly debs in parallel

evgeni:


ekohl:
I think these are fairly heavy operations given they take about 15 minutes together. Any idea if we'll overload the server or is it a single CPU thread that's the limit?

evgeni:
Only one way to find out ;)
This is btw copied from your release job. So just syncing bugs ;)

PR #995, opened 2019-03-26
use the right credentials when submitting koji data to github

evgeni:
Closes: #994

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/koji.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/koji.groovy
@@ -5,7 +5,7 @@ def status_koji_links(repo) {
         taskinfo_yaml = readYaml text: taskinfo
         build_status = (taskinfo_yaml["State"] == 'failed') ? 'FAILURE' : 'SUCCESS'
         build_package = taskinfo_yaml["Request Parameters"]["Source"].split('/')[-1]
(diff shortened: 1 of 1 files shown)


evgeni:
@ekohl this still gives `The supplied credentials are invalid to login` when running it, is `theforeman-bot` not having the correct API token?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/koji.groovy
```diff
@@ -5,7 +5,7 @@ def status_koji_links(repo) {
         taskinfo_yaml = readYaml text: taskinfo
         build_status = (taskinfo_yaml["State"] == 'failed') ? 'FAILURE' : 'SUCCESS'
         build_package = taskinfo_yaml["Request Parameters"]["Source"].split('/')[-1]
-        githubNotify credentialsId: 'github-token', account: 'theforeman', repo: repo, sha: "${ghprbActualCommit}", context: "koji/${build_package}", description: "koji task #${task}" , status: build_status, targetUrl: "http://koji.katello.org/koji/taskinfo?taskID=${task}"
+        githubNotify credentialsId: 'theforeman-bot', account: 'theforeman', repo: repo, sha: "${ghprbActualCommit}", context: "koji/${build_package}", description: "koji task #${task}" , status: build_status, targetUrl: "http://koji.katello.org/koji/taskinfo?taskID=${task}"
```
ekohl: That's a username/password. I think it only works with a token. Currently we have an auto generated one but I can generate a manual with the github-token ID.

PR #996, opened 2019-03-26
Ensure rubygems is installed for rbot

tbrisker:


puppet/modules/rbot/manifests/package.pp
```diff
@@ -18,6 +18,8 @@
 
   $extra_gems = [ 'mechanize', 'tzinfo', 'tokyocabinet' ]
 
+  ensure_packages { [ 'rubygems' ] }
+  ->
```
ekohl: I don't think this is valid on a function. It'd be better to add a `require => Package['rubygems']` to `package { $extra_gems:`

tbrisker: we also have a gem on debian in $packages, so i think this needs to be before that? or maybe add that in that step?

ekohl: I guess this would work:
```puppet
$gem_packages = $facts['osfamily'] ? {
  'Debian' => ['rubygems-integration'],
  'RedHat' => ['rubygems'],
  default  => [],
} 
ensure_packages($gem_packages)

package { $extra_gems:
  require => Package[$gem_packages],
}
```

PR #1000, opened 2019-03-27
Update release jobs

ekohl:
Individual commits have relevant commit messages:

d741fbd6 (Ewoud Kohl van Wijngaarden, 3 minutes ago)
   Introduce a hammer-cli-katello-master-test job

   This runs the tests before releasing.

d6b1d286 (Ewoud Kohl van Wijngaarden, 20 minutes ago)
   Introduce a foreman-selinux-develop-test job

   This follows the pattern of having a -test job to verify functionality and a -release job to build packages.

   It avoids the rebuild for every foreman-packaging change issue.

   This change drops the timed builds which aren't needed for foreman-selinux since it doesn't bundle anything.

673b7205 (Ewoud Kohl van Wijngaarden, 35 minutes ago)
   Add a foreman-installer-develop-test job

   This works towards the pattern of having a -test job to run tests and then a -release job to actually release it.

   An additional benefit is that we will stop building an installer for every merged PR to foreman-packaging.

   It is not complete since the tests with the actual modules, as we do in Travis, are not ran. Those wouldn't be correct because there is a timing issue. The -test job runs, then the -release job builds a tarbal from potentially different sources.

0b22c711 (Ewoud Kohl van Wijngaarden, 15 minutes ago)
   Do not use isAllWhitespace

   This method is not allowed in a sandbox and breaks our release pipelines.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-installer-develop-test.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman-installer-develop-test.groovy
@@ -1,0 +1,68 @@
+def commit_hash = ''
+
+pipeline {
+    agent any
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage("Collect Git Hash") {
+            steps {
+                git(url: 'https://github.com/theforeman/foreman-installer', branch: 'develop')
+                script {
+                    commit_hash = archive_git_hash()
+                }
+            }
+        }
+        stage("test ruby 2.0 & puppet 5") {
+            steps {
+                run_test(ruby: '2.0.0', puppet: '5.5')
+            }
+        }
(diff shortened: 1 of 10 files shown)


ekohl:
Note that katello is still the only one left after this to not have a -test job.

ehelms:
Should we check with hammer-cli-katello team if they want a test job? Katello does not have a test job because it was removed by choice of katello developers to remove the complexity of having a long running test job that blocked PR testing. If the test jobs are instead meant to 'only generate a source' that is passing tests and not block PR merges, then they can make more sense (although I might still argue they can be a burden). 

I am getting a bit into furture design talk but perhaps we need to go a route of `source-release` jobs that run tests, generate source, and archive it.

ekohl:
> having a long running test job that blocked PR testing

The test job shouldn't block PR testing, unless you want to test it in nightlies. I'd argue that currently we don't really support that well anyway



> I am getting a bit into furture design talk but perhaps we need to go a route of `source-release` jobs that run tests, generate source, and archive it.

I do like that. It is much more like the old test/packaging_trigger jobs we had: generate validated sources that packaging could then use.

PR #1001, opened 2019-03-27
Do not use isAllWhitespace

ekohl:
This method is not allowed in a sandbox and breaks our release pipelines.


tbrisker:
Looks like installer and selinux are now failing because they don't have a git_ref parameter

tbrisker:
opened https://github.com/theforeman/foreman-infra/pull/1007 to add them to those jobs, though i'm not sure that's the right approach

PR #1002, opened 2019-03-27
Sort jenkins plugin list

tbrisker:
This doesn't change any of the versions, just reorders them so its
easier to find in the list.


tbrisker:
updated the command as well

PR #1008, opened 2019-03-31
set BUILD_TIMESTAMP in pbuilder scripts

mmoll:


mmoll:
This was coming from a jenkins plugin, I presume (https://plugins.jenkins.io/build-timestamp)

PR #1009, opened 2019-04-01
Ensure selinux-policy-devel is installed

ekohl:
5c1ba8ac124b120492c4d91ba2455f6da195ac90 introduced a foreman-selinux test job, but this requires selinux-policy-devel to be installed.


evgeni:
I do wonder if this starts to warrant a "redhat" subclass?

ekohl:
Probably, but didn't want to make this PR bigger than needed.

PR #1012, opened 2019-04-04
Handle nil values

ekohl:
I'm not 100% sure this improves things, but sometimes we get error mails on on nil values and this should improve the debug options.

### DIFF
--- a/puppet/modules/redmine/files/git_repos.sh
+++ b/puppet/modules/redmine/files/git_repos.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
+
+set -e -o pipefail
 
 CODE_DIR=$1
 DATA_DIR=$2
@@ -34,7 +36,7 @@ if [[ ! -n $CODE_DIR ]] || [[ ! -n $DATA_DIR ]] ; then
 fi
 
 # Sync repositories for all known git repos
-curl -s $REDMINE_REPOS | ruby -rjson -e '
+curl --fail --silent $REDMINE_REPOS | ruby -rjson -e '
 JSON.load(STDIN).each do |project_name,repos|
   repos.each do |repo,branches|
     org_name, repo_name = repo.split("/", 2)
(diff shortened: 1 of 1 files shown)


puppet/modules/redmine/files/git_repos.sh
```diff
@@ -36,6 +36,10 @@ fi
 # Sync repositories for all known git repos
 curl -s $REDMINE_REPOS | ruby -rjson -e '
 JSON.load(STDIN).each do |project_name,repos|
+  unless repos
```
evgeni: I think you're chasing the wrong `each`. The error I see in my INBOX is 
```
-e:2:in `<main>': undefined method `each' for nil:NilClass (NoMethodError)
```
*line 2*, but `repos.each` is on line 3 (the first line being the empty line at `ruby -rjson -e '`).

I think the failing `each` is `JSON.load(STDIN).each`. Test:

```
% cat /tmp/test.sh 
#!/bin/sh

echo "null" | ruby -rjson -e '
JSON.load(STDIN).each do |p|
	puts p
end'

% bash /tmp/test.sh
Traceback (most recent call last):
-e:2:in `<main>': undefined method `each' for nil:NilClass (NoMethodError)
```

ekohl:
So how about this?

evgeni:
Aren't we fetching just one url? I thought fail early helps only if you have multiple?

ekohl:
You're right. It doesn't imply `--fail`. Updated.
>        -f, --fail
>              (HTTP)  Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to bet‐
>              ter deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an  HTML
>              document stating so (which often also describes why and more). This flag will prevent curl from outputting that and
>              return error 22.
>
>              This method is not fail-safe and there are occasions where non-successful response codes will slip  through,  espe‐
>              cially when authentication is involved (response codes 401 and 407).

evgeni:
@ekohl I think this script needs a `set -o pipefail`, as otherwise `--fail` will fail the `curl` call, but the pipe will still be executed.

```
% curl --fail --silent https://die-welt.net/nope.json | ruby -rjson -e 'JSON.load(STDIN).each {|p| puts p}'
Traceback (most recent call last):
-e:1:in `<main>': undefined method `each' for nil:NilClass (NoMethodError)
```

ekohl:
Updated to set pipefail.

PR #1013, opened 2019-04-04
Modernize the JJB module

ekohl:
The timeout was redundant since it matched the default config value. Run is unused since 3826ee4d844d46e8800d638117b5e193463d2ba2.


puppet/modules/jenkins_job_builder/manifests/init.pp
```diff
@@ -12,15 +12,16 @@
 # $run is our addition and is so named becuase noop is a bit too magic in puppet syntax...
```
evgeni: should this comment be dropped to then?

ekohl: Done

mmoll:
merged, bedankt @ekohl!

PR #1017, opened 2019-04-05
ci: run rubocop version of plugin

timogoebel:


timogoebel:
> If this is needed

It is to avoid commits [like this](https://github.com/theforeman/foreman_monitoring/commit/1006f703af498890c8a7bc0604f9b69df7e7eccb) as every new rubocop version adds new cops which make the plugin's master branch fail.

> perhaps we should do this on line 64.

Not if we want the tests to fail early.

> Also, I don't see any current rubocop execution.

Jenkins executes the `test:jenkins` rake task. Plugins then [hook into it](https://github.com/theforeman/foreman_expire_hosts/blob/da5cf8bfa82be0751614071f6288f580359bec7d/lib/tasks/expired_hosts.rake#L46) and enhance that task to run rubocop.

> what if a newer rubocop disagrees with the rubocop in core

Core can use a different version than the plugin. I don't see a problem there.

> Does that always fail the build?

Why should it?

ekohl:
> Core can use a different version than the plugin. I don't see a problem there.

I'm worried that rubocop from core wants a certain style (never any trailing comma) and the plugin wants another style (always trailing comma). That's the conflict I'm worried about and I suspect that'd result in always failing builds.

timogoebel:
> That's the conflict I'm worried about and I suspect that'd result in always failing builds.

It won't. A plugin can use a different ruleset. That's actually something that is currently already possible.

timogoebel:
@ekohl: How can we continue here?

timogoebel:
As this will probably never go in, I'm closing this and move the plugins to github actions.

PR #1020, opened 2019-04-10
Move to a source and package release model

ehelms:
This changes the idea of a 'test' job that is run on PR merge to
a source release job. The source release job aims to test the code
and then create a source release if that tested code is green.

The previous release job is renamed to be a package-release job to
better specify that its focused only on releasing a package.

The overall idea is to split source generation and release from
package creation given that source can be an input to multiple places.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
@@ -24,7 +24,7 @@ def setup_nightly_build_environment(args) {
 def generate_sourcefiles(args) {
     def sourcefile_paths = []
     def project_name = args.project_name
-    def ruby_version = args.ruby_version ?: env.ruby_version
+    def ruby_version = args.ruby_version ?: '2.5'
     def source_type = args.source_type
 
     dir(project_name) {
@@ -36,11 +36,16 @@ def generate_sourcefiles(args) {
             archiveArtifacts(artifacts: sourcefiles.join(','))
             sourcefile_paths = sourcefiles.collect { "${pwd()}/${it}" }
         } else {
-            withRVM(["bundle install --jobs 5 --retry 5"], ruby_version)
-            withRVM(["bundle exec rake pkg:generate_source"], ruby_version)
-            archiveArtifacts(artifacts: 'pkg/*')
-            sourcefile_paths = list_files('pkg/').collect {
-                "${pwd()}/pkg/${it}"
+            try {
+                configureRVM(ruby_version)
+                withRVM(["bundle install --jobs 5 --retry 5"], ruby_version)
+                withRVM(["bundle exec rake pkg:generate_source"], ruby_version)
+                archiveArtifacts(artifacts: 'pkg/*')
+                sourcefile_paths = list_files('pkg/').collect {
+                    "${pwd()}/pkg/${it}"
+                }
+            } finally {
+                cleanupRVM(ruby_version)
             }
         }
     }
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman.groovy
@@ -1,5 +1,7 @@
 def commit_hash = ''
 foreman_branch = 'develop'
+project_name = 'foreman'
+source_type = 'rake'
 
 pipeline {
     agent { label 'fast' }
@@ -198,10 +200,20 @@ pipeline {
                 }
             }
         }
+        stage('Build and Archive Source') {
+            steps {
+                dir(project_name) {
+                    git url: 'https://github.com/theforeman/foreman', branch: foreman_branch
+                }
+                script {
+                    sourcefile_paths = generate_sourcefiles(project_name: project_name, source_type: source_type)
+                }
+            }
+        }
         stage('packaging') {
             steps {
                 build(
-                    job: 'foreman-develop-release',
+                    job: 'foreman-develop-package-release',
                     propagate: false,
                     wait: false,
                     parameters: [
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/source/foreman-develop-source-release.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/source/foreman-develop-source-release.yaml
@@ -1,5 +1,5 @@
 - job:
-    name: foreman-develop-test
+    name: foreman-develop-source-release
     project-type: pipeline
     sandbox: true
     quiet-period: 2700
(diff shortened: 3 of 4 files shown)


ehelms:
This is obviously not the cleanest re-usable solution but I wanted to get your thoughts @ekohl from our discussions.

ehelms:
This would require https://github.com/theforeman/foreman-packaging/pull/3635

ehelms:
@ekohl Tested and working now as well, I had missed a few things

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -29,22 +29,6 @@ pipeline {
                 )
             }
         }
-        stage('Archive Git Commit') {
```
ekohl: This file is included in various jobs:
```
$ rg foreman-x-develop -l
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-develop-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/smart-proxy-develop-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-installer-develop-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/hammer-cli-foreman-master-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/hammer-cli-master-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/katello-master-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-selinux-develop-release.yaml
puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/hammer-cli-katello-master-release.yaml
```

Which is why I suggested these all need to be updated.

ehelms: Duh. That makes complete sense. Since there is no harm in this existing till all are converted I put these lines back in this fileto migrate paradigms one by one.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -36,11 +36,16 @@ def generate_sourcefiles(args) {
             archiveArtifacts(artifacts: sourcefiles.join(','))
             sourcefile_paths = sourcefiles.collect { "${pwd()}/${it}" }
         } else {
-            withRVM(["bundle install --jobs 5 --retry 5"], ruby_version)
-            withRVM(["bundle exec rake pkg:generate_source"], ruby_version)
-            archiveArtifacts(artifacts: 'pkg/*')
-            sourcefile_paths = list_files('pkg/').collect {
-                "${pwd()}/pkg/${it}"
+            try {
+                configureRVM(ruby_version)
```
ekohl: Feels like we need a `rvm(ruby_version) { .. }` helper at some point.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
```diff
@@ -24,7 +24,7 @@ def setup_nightly_build_environment(args) {
 def generate_sourcefiles(args) {
     def sourcefile_paths = []
     def project_name = args.project_name
-    def ruby_version = args.ruby_version ?: env.ruby_version
+    def ruby_version = args.ruby_version ?: '2.5'
```
ekohl: Could you remind me: this is the groovy equivalent of `args.ruby_version || '2.5'`, right?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/source/foreman-develop-source-release.yaml
```diff
@@ -1,5 +1,5 @@
 - job:
-    name: foreman-develop-test
+    name: foreman-develop-source-release
     project-type: pipeline
     sandbox: true
     quiet-period: 2700
```
ekohl: We've also noticed this might be a bit long. I wouldn't mind if we lowered this but doesn't have to be now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -198,6 +200,16 @@ pipeline {
...
+                    git url: 'https://github.com/theforeman/foreman', branch: foreman_branch
+                }
+                script {
+                    sourcefile_paths = generate_sourcefiles(project_name: project_name, source_type: source_type)
+                }
+            }
+        }
         stage('packaging') {
             steps {
                 build(
```
ekohl: The line below needs to be `foreman-develop-package-release` now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/nightly_packaging.groovy
ehelms: Yep. Called the Elvis operator.

ehelms:
Updated. I didn't touch the quiet-period. We can work to tune that after. This is a bit of paradigm shift overall. @tbrisker @komidore64 @evgeni y'all should consider if this makes sense to y'all as well

tbrisker:
I might be missing something here, but if i understand correctly this pr only adds package generation to the tasks-formerly-know-as-test but doesn't remove them from the tasks-formerly-known-as-release?

ehelms:
That statement confused me @tbrisker . What this does is:

 * drop the idea of a 'test' job in favor of a job that releases source, the source release job runs tests to ensure a certified source
 * moves source release from the job that does packaging so packaging can focus on packaging

tbrisker:
>     * moves source release from the job that does packaging so packaging can focus on packaging
I think this is what I seem to be missing, i see the first point, but i don't see where you removed the source release from the packaging job.

ekohl:
@tbrisker some background:

In the past we had a `test_develop` job. That was responsible for testing any new commits in foreman develop. It also created a tarball and archived this. We used this tarball to release nightly packages from the latest successful build. It was also used in scratch builds (like PR testing).

A while back the job renamed to `foreman-develop-test` to better clarify what it was, but that also removed the tarball building because it was considered to be more of a packaging thing (foreman-develop-release). I disagreed, but didn't want to hold up the refactor.

We're now running into the issues that caused my disagreement: both RPMs and debs need to pass for the packaging build to succeed. That means PR testing doesn't have the latest source tarball which makes testing PRs that fix the build very hard.

To address this, we're reverting to the old model: one job to produce a verified source tarball. To better indicate its purpose this is now named `<project>-<branch>-source-release`. The other release job is only responsible for packages so it's renamed to `<project>-<branch>-package-release`.

Because this only renamed the foreman job, we still build the tarball in the package-release. This is not the final goal but rather to copy over the tarball from the source-release job. For that we need to convert all jobs first. I don't mind doing it in phases so we can see if this works rather than doing all jobs and having to fix all jobs.

ehelms:
@ekohl answered the question, but I did want to mention I see now what you were asking @tbrisker . The idea is if we like this paradigm to do one-by-one. The corresponding update is:

https://github.com/theforeman/foreman-packaging/pull/3635

This will officially point nightlies to the new source generated location.

PR #1025, opened 2019-04-15
Always install node modules for plugin prs

tbrisker:
Salt depends on foreman-tasks which needs npm install to properly
compile webpack, but because salt doesn't have a package.json npm
install is run too early, when only foreman core is set up.


tbrisker:
I know we should be moving this to pipelines eventually, but this is currently blocking foreman_salt PR tests from passing.

PR #1026, opened 2019-04-17
Fix hammer-cli-foreman pull request tests

jturel:
Should fix the problem here: https://github.com/theforeman/hammer-cli-foreman/pull/419#issuecomment-483842232


mbacovsky:
Yes that was the goal. Some past versions of hammer-cli-foreman are not compatible with master hammer -cli so fall-back to installing hammer-cli gem matching the deps restrictions sounds like the right thing to do.

mbacovsky:
@ekohl, thanks for merging. Any idea when the change is going to be effective on hammer-cli-foreman PRs? I triggered tests on  https://github.com/theforeman/hammer-cli-foreman/pull/419 and it fail with the same error.

ekohl:
Looks like https://ci.theforeman.org/job/puppet-deploy/55/ failed

PR #1029, opened 2019-04-22
Move main Foreman nightly projects to source and package release jobs

ehelms:


ehelms:
This converts foreman-selinux, smart-proxy and foreman-installer. Hammer will be their own set of changes and then Katello on it's own.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
```diff
@@ -33,10 +36,20 @@ pipeline {
                 run_test(ruby: '2.5', puppet: '6.3')
             }
         }
-        stage("Release Nightly Package") {
+        stage('Build and Archive Source') {
+            steps {
+                dir(project_name) {
+                    git url: "https://github.com/theforeman/${project_name}", branch: foreman_branch
```
ekohl: We already have a git clone so this may not be needed? I don't mind having a clean directory though.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
```diff
@@ -12,7 +15,7 @@ pipeline {
     stages {
         stage("Collect Git Hash") {
             steps {
-                git(url: 'https://github.com/theforeman/foreman-installer', branch: 'develop')
+                git(url: 'https://github.com/theforeman/foreman-installer', branch: foreman_branch)
```
ekohl:
```suggestion
                git(url: "https://github.com/theforeman/${project_name}", branch: foreman_branch)
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
ehelms: The current code requires it be in a directory named for the project. I could change that code before, but I wanted to instead adapt jobs to this pattern then do re-factoring.

ekohl: I noticed a similar thing and had the same thought: good improvement, not now.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
```diff
@@ -1,4 +1,7 @@
 def commit_hash = ''
+source_type = 'rake'
```
ekohl: For my understanding: what's the difference between using `def` and not using def?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-selinux.groovy
```diff
@@ -0,0 +1,61 @@
...
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage("Collect Git Hash") {
+            steps {
+                git(url: 'https://github.com/theforeman/foreman-selinux', branch: 'develop')
```
ekohl:
```suggestion
                git(url: "https://github.com/theforeman/${project_name}", branch: 'develop')
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
ehelms: `def` scopes the variable and `no def` keeps it global. I've seen issues where script variables with def are not seen by all parts of a script.

ehelms:
Updated - thanks for catching the missing name changes.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-installer.groovy
ekohl: In the future I would like to look at defining these in the job yaml file so we can template it anyway.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/source/foreman-installer-source-release.yaml
```diff
@@ -12,6 +12,6 @@
       - github
     dsl:
       !include-raw:
-        - pipelines/test/foreman-installer-develop-test.groovy
+        - pipelines/release/source/foreman-installer.groovy
```
ekohl: This needs to include the groovy file which defines `generate_sourcefiles` as well. Applies to `*-source-release.yaml`

ehelms: Good catch. Updated.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman-selinux.groovy
```diff
@@ -0,0 +1,61 @@
...
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage("Collect Git Hash") {
+            steps {
+                git(url: "https://github.com/theforeman/${project_name}", branch: 'develop')
```
ekohl:
```suggestion
                git(url: "https://github.com/theforeman/${project_name}", branch: foreman_branch)
```

ekohl:
Could you prepare a packaging PR that updates the manifest to use the new jobs? After that I think we can merge this.

ehelms:
Here ya go -- https://github.com/theforeman/foreman-packaging/pull/3679

PR #1030, opened 2019-04-24
fastly CDN configuration

evgeni:


evgeni:
(yes, the api key is revoked *whistle*)

.gitmodules
```diff
@@ -140,3 +140,7 @@
 [submodule "puppet/modules/transition"]
 	path = puppet/modules/transition
 	url = https://github.com/puppetlabs/puppetlabs-transition
+[submodule "ansible/roles/Jimdo.fastly"]
```
ehelms: Why submodule vs. using galaxy to install via requirements?

evgeni: For one because I didn't upload my fork to galaxy, and we use submodules for Puppet too, so it felt in sync.

ehelms: Okie doke. I can live with it, just asking in case there was an opportunity to start clean.

PR #1032, opened 2019-04-25
disable concurrent builds for foreman-packaging-release

evgeni:
not only set concurrent:false but also add disableConcurrentBuilds to
the pipeline definition

the way we currently track "hat has been built" does not allow for
concurrent building


ekohl:
I thought we set this in the YAML file and that was sufficient. I haven't seen this run concurrently since. Did you see it?

ehelms:
I'm with @ekohl -- shouldn't the yaml job definition set this?

evgeni:
Should...

https://ci.theforeman.org/job/foreman-packaging-release/75/ and https://ci.theforeman.org/job/foreman-packaging-release/76/ ran in parallel though 🤷

ehelms:
So what are we missing? This screenshot shoes it waiting.

Screenshot from 2019-04-25 13-01-45

evgeni:
Now that you say it. I think 76 did wait. But then still started while 75 was still running.

PR #1034, opened 2019-04-25
Add 1.22 pipelines

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/client/1.22.groovy
```diff
@@ -0,0 +1,75 @@
...
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Mash Koji Repositories') {
+            agent { label 'sshkey' }
+
+            steps {
+                mash("foreman-client-mash-split-${version}.py")
```
ehelms: env.version ?

ekohl: Probably. It's currently a copy of what's in 1.21. Longer term I'd like to migrate to `mash("foreman-client-mash-split.py, env.version)` but I thought to leave that in a separate PR.

PR #1038, opened 2019-04-29
Allow -package-release jobs as sources

ekohl:
Currently these jobs still build tarballs and are package sources so they should be allowed.


mmoll:
merged, bedankt @ekohl!

PR #1041, opened 2019-05-06
add yum.tfm.o to Fastly config

evgeni:


evgeni:
When this is approved, merged and executed, I will switch the DNS with Ohad.
The corresponding log buckets exist in Rackspace already (there is no good Ansible module for that).

ehelms:
@evgeni Make it so!

mmoll:
merged, спасибо Женя!

evgeni:
executed, works. mailed ohad to update the dns

PR #1042, opened 2019-05-06
Handle HTTPS in unmanaged_jobs.rb

ekohl:
Previously it stripped the scheme. That meant it fell back to HTTP which does a redirect to HTTPS. This raised an error instead of the intended operation. By including the scheme, net/http can perform the correct request.


mmoll:
merged, bedankt @ekohl!

PR #1043, opened 2019-05-07
shift time of unmanaged_jobs.rb

mmoll:


mmoll:
this should prevent the conflict of the lock with the delete-old job

PR #1046, opened 2019-05-09
update rvm

mmoll:


mmoll:
This is needed for Ruby 2.6 and hopefully fixes the problem we were seeing on RH based Distros with 1.29.7.

PR #1049, opened 2019-05-15
Deploy yum content to web02

ekohl:
Now that yum.theforeman.org is behind a CDN, we need to push to the actual host which is currently web02.


mmoll:
merged, bedankt @ekohl!

PR #1052, opened 2019-05-16
drop Ruby 2.4 in develop tests

mmoll:


mmoll:
As the SCLs of all supported versions use 2.5 now, we should drop this in order to add 2.6 soon.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/hammer-cli-x.groovy
```diff
@@ -24,11 +24,6 @@ pipeline {
                 run_test(ruby: '2.3')
             }
         }
-        stage("test-ruby-2.4") {
```
ekohl: Why do we drop 2.4 but not 2.3? Is it because we have no distro with 2.4, but still have some with 2.3?

mmoll: Exactly, Debian 9 and Ubuntu 16.04 use 2.3.

PR #1055, opened 2019-05-17
Remove Ruby 2.2 tests

mmoll:


mmoll:
As Puppet 4 AIO uses Ruby 2.1, these could still stay for a while...

PR #1056, opened 2019-05-17
Use Rubygems 3.0.3 with Ruby 2.6

mmoll:


mmoll:
Ruby 2.6 is bringing Rubygems 3.0.x, which is not compatible with some older versions and downgrading to an older version requires `--force`.

ekohl:
Do we really want to divert from the standard Ruby 2.6? We could easily mask issues that we'd hit in practice.

mmoll:
True, but then I'd suggest to just remove the version check and take what's coming which each Ruby at install time...

mmoll:
As it turns out, `slave::rvm_config` is already capable of doing all that... how about this? :)

PR #1057, opened 2019-05-19
Add Ruby 2.6 tests

mmoll:


mmoll:
A few proxy-jobs are missing, but I'd rather like to add these after this gets in when adding Puppet 6.x to these.

PR #1058, opened 2019-05-19
Only replace puppet in puppet.rb if it's present

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #1061, opened 2019-05-21
Trigger katello-master-package-release with correct parameter

tbrisker:
Otherwise, k-m-source-release triggers it with an empty gitref causing
the job to fail


evgeni:
NACK, see https://github.com/theforeman/foreman-infra/pull/1062

tbrisker:
closing in favor of #1062

PR #1051, opened 2019-05-16
Copy sources in foreman-x-develop-release

ekohl:
Since all jobs now follow $project-$branch-source-release, we can rely on those to generate the tarball.

Also cleans up an unused publisher and updates the conventions in README.md.

PR #1062, opened 2019-05-21
collect the commit_hash when releasing katello

evgeni:


ekohl:
Alternative is to finish https://github.com/theforeman/foreman-infra/pull/1051

evgeni:
@ekohl yeah, but k-master releases are broken now, so I went for the quickest route :)

PR #1066, opened 2019-05-28
Adding foreman-documentation job

lzap:
What I want to achieve in https://github.com/theforeman/foreman-documentation

* Generate documentation for master branch
* Generate documentation per PR
* Keep workspace (generated HTML and PDF files)
* Wipe old jobs after few days

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-documentation.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-documentation.yaml
@@ -1,0 +1,8 @@
+- job:
+    name: foreman-documentation-test
+    node: 'admin'
+    scm:
+      - git:
+          url: "https://github.com/theforeman/foreman-documentation"
+    builders:
+      - shell: !include-raw: scripts/test/docs_foreman_test.sh
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/docs_foreman_test.sh
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/docs_foreman_test.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+pushd guides/
+make clean html pdf
+popd


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-documentation.yaml
```diff
@@ -0,0 +1,9 @@
+- job:
+    name: foreman-documentation-test
+    node: 'admin && sshkey'
```
ekohl: You don't need this since you're not using any SSH keys

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-documentation.yaml
```diff
@@ -0,0 +1,9 @@
...
+    scm:
+      - git:
+          url: "https://github.com/theforeman/foreman-documentation"
+          wipe-workspace: false
```
ekohl: Just keeping the workspace doesn't mean you archive any result. It just means that you *may* get a non-clean workspace from a previous job when a new one starts.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-documentation.yaml
lzap: Thanks I was not sure.

lzap:
Rebased I need to see first what is the default retention policy and how artifacts are kept.

ehelms:
We've been migrating jobs to pipelines as a standardization point. I won't hold this up on account of it but I would *strongly* prefer we go that route.

I see you are wanting PR and master to be handled. That will require two separate jobs, one for each case. Have you considered using Travis since this is so simple? You could even have travis build the docs on push to master *and* push those built docs to a gh-pages branch and get automatic hosting on Github as a starting point.

ekohl:
From offline conversations: I know that @lzap wants the result to be online browsable and with Travis you need some sort of storage. That does imply that this is still missing archiving the artifacts.

ehelms:
Right -- here's your browsable storage -- https://pages.github.com/

lzap:
I don't understand - you recommend me to push resulting HTML into some github repository? Can I share credentials with Travis? I don't think so.

PR #1072, opened 2019-06-03
Isolate packaging checkout during repoclosure

ehelms:


ehelms:
This would override any existing checkouts in the primary directory such as foreman-infra and break things like pushing RPMs.

Test run -- https://ci.theforeman.org/job/foreman-plugins-rpm-pipeline/102/console

PR #1073, opened 2019-06-03
Refactor plugins pipeline

ekohl:
This refactors the pipeline to be driven by configuration and allows changing the versions by just changing the versions hash at the top. The end result is still readable for end users in the same way. It may be even more readable by printing the exception.

It also allows defining multiple distributions which will be useful when we add EL8 support.


ekohl:
Result on my local Jenkins:

image

PR #1074, opened 2019-06-06
Use an isolated workspace to perform repoclosure

ehelms:


ehelms:
I considered that. However, *every* parallel step needs to checkout a special branch therefore they still need isolated workspaces. Further, it makes the repoclosure step dependent on something having cloned and configured the foreman-infra environment. We could take a tactic similar to obal where we call a setup step that checks for an existing directory and otherwise clones. We will still need the isolated workspace.

ehelms:
@ekohl Can you take a look to fix client pipelines?

PR #1075, opened 2019-06-07
Improve discourse notifications

ekohl:
The goal of this PR is to make it easier to see why a build failed.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
@@ -10,13 +10,13 @@ def list_files(glob = '') {
     sh(script: "ls -1 ${glob}", returnStdout: true).trim().split()
 }
 
-def set_job_build_description(job_name) {
+def set_job_build_description(job_name, status) {
     def build_description = ""
     def file_name = "jobs/${job_name}"
 
     if (fileExists(file_name)) {
        link = readFile(file_name)
-       build_description += "<a href=\"${link}\">${job_name}</a><br/>"
+       build_description += "<a href=\"${link}\">${job_name}</a> (${status})<br/>"
     }
 
     if (currentBuild.description == null) {
@@ -27,6 +27,7 @@ def set_job_build_description(job_name) {
 }
 
 def runIndividualCicoJob(job_name, number = 0) {
+    def status = 'unknown'
     sleep(number * 5) //See https://bugs.centos.org/view.php?id=14920
     try {
         withCredentials([string(credentialsId: 'centos-jenkins', variable: 'PASSWORD')]) {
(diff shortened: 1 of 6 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -10,13 +10,13 @@ def list_files(glob = '') {
...
 
-def set_job_build_description(job_name) {
+def set_job_build_description(job_name, status) {
     def build_description = ""
     def file_name = "jobs/${job_name}"
 
     if (fileExists(file_name)) {
        link = readFile(file_name)
-       build_description += "<a href=\"${link}\">${job_name}</a><br/>"
+       build_description += "<a href=\"${link}\">${job_name}</a> (${status})<br/>"
```
ehelms: This would put it on the same line right at the end of the link. I think it might look better to do a `<br/>Status: ${status}<br/>` type of formatting.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -80,3 +85,16 @@ def runCicoJobsInParallel(jobs) {
         }
     }
 }
+
+def notifyDiscourse(heading=null) {
+    body = [heading, env.BUILD_URL]
+    if (currentBuild.description) {
```
ehelms: What is currentBuild? I think that should be passed in. Keep this as a simple function.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -80,3 +85,16 @@ def runCicoJobsInParallel(jobs) {
         }
     }
 }
+
+def notifyDiscourse(heading=null) {
+    body = [heading, env.BUILD_URL]
```
ehelms: Same here, let's pass this in instead of relying on magic `env` inside a function.

ehelms: I don't understand what `heading` is here. Makes me think of subject by its naming but that is already being set?

ehelms:
A few code organization comments, change looks like a good one.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
ekohl: Would `bodyHeading` better cover what it's intended to do? `introText`?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
ekohl: The intent was to keep this short (`(unknown)`, `(passed)` and `(failed)`) so you knew what it was relating to. If it's a new line and you have multiple builds then it can become unclear if it belongs to the one above or below. Given we have 3 builds in the Deb builds, that's a real risk with the middle one.

ekohl:
How about now?

Issue #1068, opened 2019-05-30
Move away from Jenkins RVM plugin


mmoll:
The plugin saw its last release in 2016 and is known not to work on Java 11.

ekohl:
I'm not sure we even use this RVM plugin. I think we always use shell scripts.

mmoll:
Looks like it...

PR #1081, opened 2019-06-22
don't install outdated Jenkins RVM plugin

mmoll:
closes GH-1068


tbrisker:
do we have any pipelines or jobs still using it?

mmoll:
AFAICT no. Besides searching in the job definitions even I deinstalled the plugin for a few days and watched out for messages and errors... nothing came up.

PR #1082, opened 2019-06-25
add /api/latest redirect

evgeni:


evgeni:
> Should we also add the next like we do with manuals?

there are no api docs for nightly, so no

PR #1085, opened 2019-07-01
Use Java 11 for Jenkins master

mmoll:


mmoll:
I just tested Java 11 manually and the master starts up fine and while I didn't run any jobs, no messages indicating incompatibilites could be found, so I'm quite confident this works.

The Puppet code is untested.

puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -191,6 +192,9 @@
     'zentimestamp'                       => {},
   }
 
+  package { $old_packages:
+    ensure => 'absent',
+  } ->
```
ekohl: This doesn't work. I've added a test to verify this. I'd also expect that you first install the new packages and then remove the old one so there's still a virtual that provides the JRE for Jenkins.

mmoll:
:green_heart: now

ekohl:
Feel free to merge and roll out when you have time.

PR #1086, opened 2019-07-01
Remove kubevirt hammer job

ekohl:
This job is tested as it were a foreman plugin. That doesn't work.


mmoll:
merged, bedankt @ekohl!

PR #1087, opened 2019-07-01
remove Jenkins jobConfigHistory plugin

mmoll:


mmoll:
As all jobs are handled via jjb now, this shouldn't be needed anymore. It fills up `/var/lib/jenkins/config-history/jobs` on the Jenkins master with thousands of files which get backupped onto the Puppet master, which in turn exceeds the INodes there...

PR #1089, opened 2019-07-10
Show the hammer gemset before testing

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/hammer-cli-x.groovy
```diff
@@ -68,9 +68,11 @@ pipeline {
 }
 
 def add_hammer_cli_git_repos(repos = []) {
+    content = ''
```
evgeni: content *could* be an array and we *could* join it with `"\n"`, but doesn't have to

ekohl:
I'll leave that as an exercise for a future reader.

PR #1093, opened 2019-07-15
Remove Puppet from the Smart Proxy testing matrix

ekohl:
Now that the Puppet gem is never included anymore, it's no longer needed to test this. Also adds Ruby 2.6 to the matrix.


mmoll:
merged, bedankt @ekohl!

PR #1094, opened 2019-07-15
Remove null values from the discourse notification

ekohl:


ekohl:
https://stackoverflow.com/questions/3285241/remove-null-items-from-a-list-in-groovy suggested this was the easiest way.

mmoll:
merged, bedankt @ekohl!

tbrisker:
looks like this in combination of the upgrade of the email extension to a new version that includes https://jenkins.io/security/advisory/2019-03-06/#SECURITY-1340, this fails because sandbox doesn't allow .findAll. shall we whitelist the method or change the implementation?

PR #1095, opened 2019-07-16
Compile webpack when running integration tests

tbrisker:
Plugins currently enhance jenkins:unit task with all of their tests
including integration tests if present. This causes failures since we no
longer compile webpack by default when running tests. As a side effect,
plugins that do not include integration tests will no longer install
node modules as the previous condition was incorrect - it tested for the
presence of package.json in APP_ROOT (which is foreman core) instead of
PLUGIN_ROOT (which is the plugin's folder).


mmoll:
merged, thanks @tbrisker!

tbrisker:
GH-1096 is the same change for plugin PR tests

PR #1096, opened 2019-07-17
Run webpack compile also for plugin pr tests when needed

tbrisker:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/builders/foreman-plugins-pull-request.yaml
```diff
@@ -54,12 +54,16 @@
...
-          npm install
-
           # Now let's add the plugin migrations
           bundle exec rake db:migrate
 
+          # If the plugin contains integration tests or triggers core integration tests,
+          # we need to install node modules and compile webpack
+          if [ -d ${{PLUGIN_ROOT}}/test/integration ] || [ ${{database}} = postgresql ]; then
+            npm install
+            bundle exec rake webpack:compile
```
ekohl: Perhaps we should change this:
```suggestion
            tasks="jenkins:integration ${{tasks}}"
```

Of course also move `tasks="jenkins:unit"` to before the if condition. Then we can drop the `[ ${{database}} = postgresql ]` part since that obviously has no effect.

tbrisker: dropping the db condition would cause running core integration tests as well on plugins for mysql and sqlite.

ekohl: Hmm, that's also ugly.

ekohl: How about using `tasks="webpack:compile ${{tasks}}"` instead? That would prevent duplicate compilation on `jenkins:integration` (which your PR does). Then for postgres it becomes `bundle exec rake webpack:compile jenkins:unit jenkins:integration`.

tbrisker: good idea, updated accordingly and also changed it for the plugin tests

PR #1097, opened 2019-07-18
Remove unused files

ekohl:
In the past we deployed this straight to /etc/puppet but we now only use
the modules. The other files are testing infrastructure for modules.


ekohl:
@mmoll can you also double check this? I may be missing something :)

mmoll:
I'm not aware of any usages of those files, so this should be fine.

mmoll:
merged, bedankt @ekohl!

PR #1098, opened 2019-07-18
configure a folder and a group for a shared git storage

evgeni:


puppet/modules/secretsgit/manifests/init.pp
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  file { $path:
+    ensure => directory,
+    owner => 'root',
+    group => $group,
+    mode => '2770',
+  }
+
+  $users.each |String $user| {
```
ekohl: I think we can also use `members => $users` on `group`.

puppet/modules/secretsgit/manifests/init.pp
```diff
@@ -0,0 +1,22 @@
...
+
+  group { $group:
+    ensure => present
+  }
+
+  file { $path:
+    ensure => directory,
+    owner => 'root',
+    group => $group,
+    mode => '2770',
```
ekohl: Can you align the parameters?

puppet/modules/secretsgit/manifests/init.pp
```diff
@@ -0,0 +1,22 @@
+class secretsgit(
+  String $group = 'secretsgit',
+  Stdlib::Absolutepath $path = '/srv/secretsgit',
+  Array[String] $users = []
+) {
+
+  group { $group:
+    ensure => present
```
ekohl: I like trailing commas :)
```suggestion
    ensure => present,
```

puppet/modules/secretsgit/manifests/init.pp
evgeni: I thought that's not a thing anymore in Puppet land? But I sure can.

puppet/modules/secretsgit/manifests/init.pp
evgeni: the [docs](https://puppet.com/docs/puppet/6.6/types/group.html#group-attribute-members) say "Requires features manages_members", and [`groupadd`](https://puppet.com/docs/puppet/6.6/types/group.html#group-provider-groupadd), the default implementation Puppet uses on Linux [does not have that feature](https://puppet.com/docs/puppet/6.6/types/group.html#group-provider-features).

PR #1103, opened 2019-07-30
use the unversioned mash script for client repos

evgeni:


evgeni:
I didn't touch them as they are "stable", but we can :)

PR #1106, opened 2019-07-31
update jenkins plugins to match reality

evgeni:


mmoll:
While this is OK, I think we should in the long run weed out this list to only contain the "top level"-plugins we really use in the jobs and leave it to those plugins to pull in their depenencies.

mmoll:
merged, спасибо Женя!

PR #1109, opened 2019-08-02
Create release test jobs on CentOS CI

ekohl:
Split off from https://github.com/theforeman/foreman-infra/pull/992 so I can start testing that part.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/jenkins-lint.py
+++ b/puppet/modules/jenkins_job_builder/files/jenkins-lint.py
@@ -7,48 +7,83 @@ import sys
 import lxml.etree
 import requests
 
-JENKINS='https://ci.theforeman.org'
-
-parser = argparse.ArgumentParser(description='lint a Jenkinsfile using a remote Jenkins instance')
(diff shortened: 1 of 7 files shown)


puppet/modules/jenkins_job_builder/files/jenkins-lint.py
```diff
@@ -7,48 +7,74 @@
...
+        validation_result = session.post(url, data=payload)
+        validation_result.raise_for_status()
+
+        validation_result_json = validation_result.json()
+
+        if not (validation_result_json['status'] == 'ok' and validation_result_json['data']['result'] == 'success'):
+            for err in validation_result_json['data']['errors']:
+                print(err)
+            print("{}: NOT OK".format(path))
+            sys.exit(1)
```
evgeni: this dies after the first failed validation, should we exit after everything is done?

ekohl: Yes. I actually thought about that but then forgot it again.

ekohl:
Updated to yield failures and print a summary at the end to stderr

PR #1111, opened 2019-08-02
Refactor the client release pipelines

ekohl:
This uses job-templates to reduce duplication.

Currently this is untested.


ekohl:
I think this is ready for review.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-release.yaml
```diff
@@ -18,12 +18,38 @@
         - 'pipelines/lib/ansible.groovy{empty}'
         - 'pipelines/lib/foreman_infra.groovy{empty}'
 
+- job-template:
+    name: 'foreman-client-{version}-rpm-pipeline'
```
ehelms: Can this be its own file please? I find it not obvious that a file named `foreman-release` would have `foreman-client-release` inside of it. I get that they follow the same versioning pattern but they are independent repositories and release streams. If you prefer one file still, can you name it `foreman-and-client-release` to be explicit?

ekohl: The intention here is to grow all release jobs. Basically https://ci.theforeman.org/view/1.23/ (and yes, I also want to define the view here). This includes the foreman and smart-proxy stable test jobs. We're also taking about a foreman-packaging release job for rpm/1.23 and I'd also like to implement it here. That way this is the only file where you add 1.24 when we start branching that. We'd only remove 1.20 here when it's dropped.

ehelms: There is a bit of a disconnect here in naming convention and what that is intended to mean. We've got `release-pipeline` and `rpm-pipeline` where the latter is used throughout other jobs and the former is exclusive to foreman (non-nightly). I get the goal of providing more centralization of updates. My only issue is that even as a long time developer in this space I am not going to know to look at this file for core release, client RPM release and core nightly.

Is there no way to include a file that contains configuration data? Or to template the files somehow? Is the plan to also include Katello into this file? and any jobs that reference version numbers?

PR #1112, opened 2019-08-05
add an UUID to the extra vars file we write

evgeni:
otherwise we might end up writing the same file twice in a very parallel
pipeline

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
@@ -45,7 +45,8 @@ def runPlaybook(args) {
 
 def writeExtraVars(args) {
     def timestamp = System.currentTimeMillis()
-    def extra_vars_file = 'extra_vars-' + timestamp.toString() + '.yaml'
+    def uuid = UUID.randomUUID()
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -45,7 +45,8 @@ def runPlaybook(args) {
 
 def writeExtraVars(args) {
     def timestamp = System.currentTimeMillis()
-    def extra_vars_file = 'extra_vars-' + timestamp.toString() + '.yaml'
+    def uuid = UUID.randomUUID()
+    def extra_vars_file = 'extra_vars-' + timestamp.toString() + '-' + uuid.toString() + '.yaml'
```
ekohl: Can you use `File.createTempFile` similar to `System.currentTimeMillis` here?

evgeni: `Scripts not permitted to use staticMethod java.io.File createTempFile` (aka: no)

evgeni:
> Below we also have `buildExtraVars` which has a very similar implementation and also calls `writeFile`. Could that be an issue as well?

Technically, yes. But there the file is only used for archiving the command line, not for actually getting the vars, so it's not that bad. Still updated.

PR #1115, opened 2019-08-14
Katello 3.13 test and release jobs

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.13.groovy
```diff
@@ -0,0 +1,52 @@
+pipeline {
+    agent none
+
+    environment {
+        foreman_version = "1.22"
```
tbrisker:
```suggestion
        foreman_version = "1.23"
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.13.groovy
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
```
tbrisker: 2 hours is not always enough for katello, i would raise it to 3 (not only here, but i won't ask for the rest in this pr)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/3.13.groovy
```diff
@@ -0,0 +1,52 @@
+pipeline {
+    agent none
+
+    environment {
+        foreman_version = "1.23"
+    }
+
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
```
evgeni:
```suggestion
        timeout(time: 3, unit: 'HOURS')
```

evgeni:
> It'd be nice to template these as well but doesn't have to be in this PR.

Yeah, I thought the same, but then went full "I want it working in the next 10 minutes" mode :)

PR #1122, opened 2019-09-10
add stable tests for foreman-tasks

mmoll:


mmoll:
@iNecas ACK? :point_up:

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman-tasks.yaml
```diff
@@ -4,5 +4,9 @@
     branch:
       - master:
           foreman_branch: develop
+      - 0.15.x:
+          foreman_branch: 1.22-stable
+      - 0.16.x:
```
ekohl: Nit: I think it's more logical to sort newest -> oldest or oldest -> newest but this feels a bit inconsistent.

iNecas:
@adamruzicka ACK? :point_up:

adamruzicka:
Makes sense, ACK

PR #1125, opened 2019-09-23
point "latest" symlinks to 1.23

evgeni:


puppet/modules/web/manifests/init.pp
```diff
@@ -11,7 +11,7 @@
 #          certs have to exist, so keep SSL vhosts disabled until the certs are present via the HTTP
 #          vhost and only then enable the SSL vhosts.
 class web(
-  String $stable = '1.22',
+  String $stable = '1.23',
   String $latest = '1.23',
```
evgeni: I don't see `$latest` or `@latest` used anywhere?

evgeni: seems unused since 130099f1d4ac9638611b1ce8e785785361210a27

PR #1126, opened 2019-09-23
drop latest param from the web class

evgeni:
we don't differentiate between "latest" and "stable" anymore

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -1,5 +1,5 @@
 # $stable::   latest release that users expect
-# $next::     next release (current nightly)
+# $next::     Next release (current nightly). To be updated as part of branching.
 #
 # $htpasswds:: Which htpasswds to create.
 #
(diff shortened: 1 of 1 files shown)


puppet/modules/web/manifests/init.pp
```diff
@@ -1,6 +1,5 @@
 # $stable::   latest release that users expect
-# $latest::   latest release that we have a manual for, change after copying it
-# $next::     latest release that we don't have a manual for, before copying it
+# $next::     next release (current nightly)
```
ekohl:
```suggestion
# $next::     Next release (current nightly). To be updated as part of branching.
```

tbrisker:
What should be done as part of branching after this change? e.g. right now, 1.23 is stable and 1.24 is nightly, but once the manual is branched nightly will become 1.25 and 1.24 will have its own manual despite not being stable yet.

tbrisker:
or in other words: is stable = 1.23 and nightly = 1.25 a valid combination? will the 1.24 manual still be accessible?

evgeni:
This PR only removed an unused param. So even if you had set it before, it was not doing anything

evgeni:
You'll have to ask the author of https://github.com/theforeman/foreman-infra/commit/130099f1d4ac9638611b1ce8e785785361210a27 for details ;)

PR #1131, opened 2019-09-26
download remote artifacts from ci.centos.org

evgeni:


puppet/modules/jenkins_job_builder/files/centos.org/ansible/jenkins_job.yml
```diff
@@ -101,6 +103,20 @@
       retries: 360
       delay: 30
 
+    - name: "Create artifacts folder"
```
ekohl: Nit: directory is the unixy term

```suggestion
    - name: "Create artifacts directory"
```

puppet/modules/jenkins_job_builder/files/centos.org/ansible/jenkins_job.yml
```diff
@@ -101,6 +103,20 @@
...
+      file:
+        path: "{{ jenkins_artifacts_folder }"
+        state: directory
+      when: jenkins_download_artifacts | bool
+
+    - name: "Fetch remote artifacts"
+      get_url:
+        url: "{{ jenkins_host }}/job/{{ jenkins_job_name }}/{{ poll_result.json.executable.number }}/artifact/{{ item.fileName }}"
```
ekohl: There's no URL in the JSON? :( Should there be a protocol in the URL as well? The rest doesn't but when I read `jenkins_host` I assume a hostname rather than a base url.

evgeni: Nope, that's all we get:
```json
 {"displayPath":"fdi-bootable-3.5.6.iso","fileName":"fdi-bootable-3.5.6.iso","relativePath":"result/fdi-bootable-3.5.6.iso"}
```

and `jenkins_host` is defined *with* schema ;)

evgeni:
updated to fetch the right path. it needs to be the relative path from the json

PR #1132, opened 2019-09-26
allow runCicoJob to pass params to the job

evgeni:


evgeni:
@ekohl you used in in `runIndividualCicoJob` and `runCicoJobsInParallel`, but not not in `runCicoJob` ;)

ekohl:
Possible. I've also had some draft branches that I've never submitted so I honestly don't know anymore.

PR #1140, opened 2019-10-09
make yum.tfm.o cache rpm (and compressed repodata) for a month

evgeni:


puppet/modules/web/manifests/init.pp
```diff
@@ -84,10 +84,24 @@
...
 
   # YUM
+  $yum_directory = '/var/www/vhosts/yum/htdocs'
+  $yum_directory_config = [
+    {
+      path    => $yum_directory,
+      options => ['Indexes', 'FollowSymLinks', 'MultiViews'],
+    },
+    {
+      path     => '.+\.(bz2|gz|rpm)$',
```
ekohl: Do we have any `.xz` files?

puppet/modules/web/manifests/init.pp
```diff
@@ -84,10 +84,24 @@
...
+    },
+  ]
+
   $yum_attrs = {
     servername      => 'yum.theforeman.org',
-    docroot         => '/var/www/vhosts/yum/htdocs',
+    docroot         => $yum_directory,
     docroot_mode    => '2575',
+    directories     => $yum_directory_config,
     custom_fragment => template('web/yum.conf.erb'),
```
ekohl: In this file we use mod_expires. Can we unify this and add `repomd.xml` to `$yum_directory_config`?

puppet/modules/web/manifests/init.pp
evgeni: I didn't `find(1)` any, but can still be added for future-proofing if you want?

puppet/modules/web/manifests/init.pp
evgeni: We'd loose the `IfModule`, but otherwise, yeah

evgeni: After a second look, no, it's not that easy.

evgeni: (don't trust the docs)

PR #1141, opened 2019-10-10
drop the includes again, we use default mods

evgeni:
:eyes:

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -17,7 +17,6 @@ class web(
 ) {
   include web::base
   include rsync::server
-  include apache::mod::expires
   include apache::mod::headers
 
   letsencrypt::certonly { 'theforeman.org':
--- a/puppet/spec/classes/web_spec.rb
+++ b/puppet/spec/classes/web_spec.rb
@@ -1,0 +1,11 @@
+require 'spec_helper'
+
+describe 'web' do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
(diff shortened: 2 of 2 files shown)


puppet/modules/web/manifests/init.pp
```diff
@@ -17,8 +17,6 @@
 ) {
   include web::base
   include rsync::server
-  include apache::mod::expires
-  include apache::mod::headers
```
ekohl: Headers is not included by default (only on freebsd). The expires one caught me off guard, but I think without this line it would break.

ekohl:
Could you add a test in `puppet/spec/classes/web_spec.rb`:

```ruby
require 'spec_helper'

describe 'web' do
  on_supported_os.each do |os, facts|
    context "on #{os}" do
      let(:facts) { facts }

      it { is_expected.to compile.with_all_deps }
    end
  end
end
```

evgeni:
nice, test failures, I'll look into it in a breeze

ekohl:
Oh, those are because the cron resource type(s) are bundled with puppet-agent but not in the puppet gem we're installing. Maybe you need to set up a `.fixtures.yml` for this. See https://github.com/theforeman/puppet-foreman/blob/master/.fixtures.yml but in this repo we're sort of doing a bastardized setup. https://github.com/puppetlabs/puppetlabs_spec_helper#using-fixtures has more docs but not sure we can combine the symlink to all checked out modules and adding additional modules.

evgeni:
After discussing with @ekohl: we'll fix the tests later.

PR #1144, opened 2019-10-10
cache normal content of yum.theforeman.org for 2 minutes

evgeni:
this should be short enough for release-time checks but also allow the
cache to serve a bit of content before expiring

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -88,8 +88,10 @@ class web(
   $yum_directory = '/var/www/vhosts/yum/htdocs'
   $yum_directory_config = [
     {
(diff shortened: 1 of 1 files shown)


evgeni:
this is partially influenced by https://community.theforeman.org/t/foreman-1-20-3-release-process/14433/8

evgeni:
@tbrisker @ekohl mind having a look? before, things were cached for 60 minutes (fastly default), now it's 2. should only affect directory indexes etc, as the packages are handled later.

evgeni:
> Any reason to use headers on the binary content and mod_expires on the others?

Yes, I didn't start working on the "consolidate cache settings" card I've droped on the backlog earlier today :)

PR #1145, opened 2019-10-10
add fixtures

evgeni:


evgeni:
@ekohl this seems to do the right thing, at least when the tests pass
otherwise you end up with the fixtures in the normal modules folder :see_no_evil:

evgeni:
@ekohl I fail to see how I can mock the `foreman()` call nicely.

ekohl:
I'd do it ugly and introduce a parameter to turn that part off. It has some more complications like generating SSH keys. A lot of magic involved.
https://github.com/theforeman/foreman-infra/commit/aa938a315ee8b9b83bb17efd66aa003bae9195e4

PR #1148, opened 2019-10-14
fixes for the foreman-discovery-image release job

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -61,14 +61,12 @@ def runIndividualCicoJob(job_name, number = 0, job_parameters = null, job_extra_
 }
 
 def runCicoJob(job_name, job_parameters = null, job_extra_vars = null) {
-    node('el') {
```
evgeni: @ekohl was there a reason to select a new node each time?

ekohl: I think this was to achieve concurrency and to make sure it ran on a correct node, but not sure this is actually needed.

evgeni: it should run just fine on the same node as it was called (imho) and the concurrency is done via `parallel`, so I'd expect it's not needed.

for me it's breaking my usecase, because the cico job trigger runs on a different node than the rest of the pipeline, where I need the downloaded artifacts

ekohl: I don't know if we can only run this on el nodes or if it would run on any. It would also be fine to modify the jobs

evgeni: oh, you actually had no agent defined in the job, right. I've updated it.

PR #1152, opened 2019-10-18
Include Debian packaging when relevant

ekohl:
Currently we've created a hostgroup for Debian just to include this class.  This creates a parity between deb and rpm packaging

Also marks private classes as such


mmoll:
tests fail :broken_heart:

ekohl:
Yes, the SSH setup is not very friendly to unit testing. In another class we've disabled it. I'll have a look at that here as well.

ekohl:
:green_heart: now. When deployed, the classes within Foreman also need to be modified.

mmoll:
merged, bedankt @ekohl!

PR #1156, opened 2019-10-21
run rsync with -rlt not -a

evgeni:
-a preserves owner, group and permissions, but we want to inherit what's
on the destination server

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-discovery-image.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-discovery-image.groovy
@@ -51,7 +51,7 @@ pipeline {
 
                     sshagent(['deploy-downloads']) {
                         // publish on web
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-discovery-image.groovy
```diff
@@ -51,7 +51,7 @@ pipeline {
 
                     sshagent(['deploy-downloads']) {
                         // publish on web
-                        sh "rsync -av --delete result/ ${destination_user}@${destination_server}:${destination_dir}/"
+                        sh "rsync -rltv --delete result/ ${destination_user}@${destination_server}:${destination_dir}/"
```
ekohl: Can we use long options?
```suggestion
                        sh "rsync --recursive --links --times --verbose --delete result/ ${destination_user}@${destination_server}:${destination_dir}/"
```

PR #1160, opened 2019-10-23
Add Debian/buster pipelines

mmoll:


mmoll:
good to review :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
```diff
@@ -67,6 +67,7 @@ pipeline {
                 // TODO: from variables
                 parallel(
                     "stretch": { push_debs_direct('stretch', foreman_version) },
+                    "buster": { push_debs_direct('buster', foreman_version) },
```
ekohl: This means it'll also attempt to push for buster on existing releases. Hence the TODO to get it from variables :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
```diff
@@ -16,6 +16,7 @@ pipeline {
                 script {
                     runCicoJobsInParallel([
                         ['name': 'debian9', 'job': 'foreman-nightly-debian9-test'],
+                        ['name': 'debian10', 'job': 'foreman-nightly-debian10-test'],
```
ekohl: Would this pass? We're very close to branching and adding it here probably means committing to getting it supported for 1.24. Do you think that's doable? I'd also be OK with experimental support.

cc @tbrisker?

tbrisker: would it be possible to have it as experimental without blocking releases if it fails? if not i'd rather wait for 1.25 with it

mmoll: The built packages worked fine locally in a vagrant VM, so I think if something fails, it would be something small, but I'm also fine with with just adding it first and run it manually to see, where it's going.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
mmoll: That would mean, it would get pushed for e.g. 1.23?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
tbrisker: makes sense to me, let's try building it manually for for now and if we feel it's stable enough we can still add it to the release pipelines in time for 1.24 GA - just so we don't get blocked in releasing RCs because something is still missing.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
mmoll: deleted.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
mmoll: deleted.

ekohl: To be clear, I think this particular line is fine for now and if it worked in a local vagrant machine, then we should keep it in here.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
ekohl: Correct. We'll need to do some groovy magic to make that configurable.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-deb.groovy
mmoll: OK, re-added... we'd need to see if it works in that particular setup anyway....

ekohl: You should also re-add the `Push DEBs` step.

ekohl:
After travis is finished of course :)

tbrisker:
travis is green. merging :crossed_fingers:

PR #1166, opened 2019-10-29
purge fastly cache for "latest" discovery image

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-discovery-image.groovy
```diff
@@ -56,6 +56,8 @@ pipeline {
                         // create symlinks
                         sh "ssh ${destination_user}@${destination_server} 'pushd ${base_dir}/releases/ && rm -f latest; ln -snf \$(ls -t | head -n 1) latest; popd' || true"
                     }
+
+                    sh "curl -X PURGE -H 'Fastly-Soft-Purge:1' https://downloads.theforeman.org/discovery/${output_dir}/fdi-image-latest.tar"
```
ekohl: Does this mean everyone can purge caches? Isn't that an easy DoS mechanism? I'm sure they limit it somehow.

evgeni: They rate limit the api requests, yeah.

For our backend the DoS is not more than if people would just request the files directly.

PR #1167, opened 2019-10-29
use mod_expires everywhere instead of setting headers ourself

evgeni:


puppet/modules/web/manifests/init.pp
```diff
@@ -100,15 +99,16 @@
...
+      provider        => 'filesmatch',
+      expires_active  => 'on',
+      expires_default => '"access plus 30 days"',
     },
     {
       path            => 'repomd.xml',
       provider        => 'files',
       expires_active  => 'on',
-      expires_default => '"access plus 0 seconds"',
+      expires_default => '"access plus 2 minutes"',
```
evgeni: IMHO it's OK to cache this for 2 minutes too :)

mmoll:
merged, спасибо Женя!

PR #1168, opened 2019-10-29
Add collection of junit XML files to ci.centos jobs

ehelms:


puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -50,8 +50,10 @@ pipeline {
 
             archiveArtifacts artifacts: 'debug/**/*.tap', allowEmptyArchive: true
             archiveArtifacts artifacts: 'debug/**/*.tar.xz', allowEmptyArchive: true
+            archiveArtifacts artifacts: 'debug/**/*.xml', allowEmptyArchive: true
 
             step([$class: "TapPublisher", testResults: "debug/**/*.tap"])
+            junit "debug/**/*.xml"
```
ekohl: I wonder if there's ever xml files other than junit. Perhaps it's just premature though.

ehelms: No idea, like you said, I'd cross that bridge when we get there.

evgeni: we did, just in a different way ;)

https://github.com/theforeman/foreman-infra/pull/1198

evgeni: https://ci.centos.org/view/Foreman/job/foreman-nightly-centos7-test/280/console

PR #1169, opened 2019-10-29
Add Katello 3.14 to Foreman 1.24 mapping

jturel:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -19,10 +23,6 @@ def katello_versions = [
         'foreman': '1.20-stable',
         'ruby': '2.5'
     ],
-    'KATELLO-3.9': [
-        'foreman': '1.20-stable',
-        'ruby': '2.5'
-    ],
```
jlsherrill: is there a reason we keep around exactly 5 releases?

jturel: We may not even need 3.10 at this point, but I don't think there's harm in keeping it around. ie I don't think 5 was explicitly chosen

mmoll:
merged, thnks @jturel!

PR #1172, opened 2019-11-06
always use web02 as the host to rsync-over-ssh things to

evgeni:
this allows deb.tfm.o to point at a CDN and ssh doesn't care for the
hostname anyways


mmoll:
merged, спасибо Женя!

PR #1176, opened 2019-11-06
mash rails repo when doing a foreman release

evgeni:
also explicitly mash nightly in the nightly pipe

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
@@ -13,6 +13,7 @@ pipeline {
             agent { label 'sshkey' }
             steps {
                 mash("foreman-mash-split.py", foreman_version)
+                mash("foreman-rails-mash-split.py", foreman_version)
             }
         }
         stage('Repoclosure') {
@@ -53,6 +54,7 @@ pipeline {
                     withRVM(["bundle install --jobs=5 --retry=5"])
                     // TODO: from variables
                     push_rpms_direct("foreman-${foreman_version}/RHEL/7", "releases/${foreman_version}/el7", false, true)
+                    push_rpms_direct("foreman-rails-${foreman_version}/el7", "rails/foreman-${foreman_version}/el7", false, true)
                 }
             }
             post {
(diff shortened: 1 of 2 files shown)


evgeni:
@ehelms when @zjhuntin and I did 1.24, we had to call the rails mashing by hand, as it seems nothing calls it, so there was no rails-1.24 repo.

is this the right approach to fix it?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
```diff
@@ -13,6 +13,7 @@ pipeline {
             agent { label 'sshkey' }
             steps {
                 mash("foreman-mash-split.py", foreman_version)
+                mash("foreman-rails-mash-split.py", foreman_version)
```
ehelms: If we are mashing, we would need to also push them?

evgeni: well, yes, of course!

evgeni: it's now being pushed.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-rpm.groovy
```diff
@@ -17,7 +17,7 @@ pipeline {
             agent { label 'sshkey' }
 
             steps {
-                mash('foreman-mash-split.py')
+                mash('foreman-mash-split.py', 'nightly')
```
ehelms: Why is this hard coded to nightly?

evgeni: because that's the nightly pipeline file and the script defaults to "nightly" if no params are given, I just wanted it to be explicit.

ekohl: Below we do call `repoclosure` with `env.foreman_version` which I think is `nightly` as well. Interestingly enough in a lot of other places we do hardcode nightly.

Issue #1162, opened 2019-10-24
per release plugin pipeline


evgeni:
That could be driven from `yaml/jobs/pipeline/foreman-release.yaml` so you automatically get a correct release pipeline. This will be especially useful when we add EL8 support since then it'll depend on the vars file.

PR #1177, opened 2019-11-06
split plugins pipeline into sepeate pipes

evgeni:
Fixes: #1162


evgeni:
WIP because it needs a new mash script that I did't create yet

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/plugins.groovy
```diff
@@ -0,0 +1,48 @@
...
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Mash Koji Repositories') {
+            agent { label 'sshkey' }
+
+            steps {
+                mash('foreman-plugins-mash-split.py', foreman_version)
```
evgeni: requires https://github.com/theforeman/foreman-infra/pull/1178

evgeni: which is now merged and I deployed the new script to koji

PR #1179, opened 2019-11-06
template katello pipelines and add 3.14

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello.groovy
```diff
@@ -18,7 +14,7 @@ pipeline {
 
             steps {
 
-                mash("katello-mash-split-3.13.py")
+                mash("katello-mash-split.py", katello_version)
```
evgeni: yes, I made the script accept a version

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/katello-rpm-pipeline.yaml
```diff
@@ -0,0 +1,25 @@
...
+        - 'pipelines/lib/obal.groovy{empty}'
+
+- project:
+    name: katello
+    jobs:
+      - 'katello-{version}-rpm-pipeline'
+    empty: ''
+    version:
+      - '3.10'
+      - '3.11'
```
tbrisker: pretty sure we can drop these two (also from the vars)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello.groovy
tbrisker: doesn't this script also need to be modified?

evgeni: define modified, it has a branch map, so it needs a modification at branching time (which I did for 3.14), but I made it not need to copy the script around as we had before.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/katello-rpm-pipeline.yaml
evgeni: yeah, probably true.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello.groovy
tbrisker: ah, i missed the previous comment about making it accept a param.

evgeni:
@ehelms @jlsherrill @chris1984 can you please ack that 3.10 and 3.11 are out of support now? :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello.groovy
evgeni: it can be seen in https://github.com/theforeman/foreman-infra/pull/1180

evgeni:
Eric said 10 and 11 can go. Mergin.

chris1984:
ACK

PR #1186, opened 2019-11-12
ignore delete errors in freight receive script

evgeni:
if we run multiple publishes in parallel, it might be that one already
finished the cleanup, so the other would error out trying to delete the
files. let's just ignore errors from the find -delete call.


mmoll:
merged, спасибо Женя!

PR #1187, opened 2019-11-13
Drop MySQL from PR & nightly testing

tbrisker:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,9 +1,7 @@
 def databaseFile(id, database = 'postgresql') {
```
ekohl: I think this function is still used by multiple versions so I'd prefer to keep this in until we don't test any mysql version anymore.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-plugins-pull-request.yaml
```diff
@@ -31,15 +31,14 @@
...
             - sqlite3
-            - mysql
       - axis:
           type: slave
           name: label
           values:
             - fast
     execution-strategy:
       combination-filter: >
-        ((database == 'postgresql') && ((ruby == '2.3') || (ruby == '2.5') || (ruby == '2.6'))) || ((database == 'mysql') && (ruby == '2.5')) || ((database == 'sqlite3') && (ruby == '2.5'))
```
ekohl: Don't we use this job also for PRs to stable branches?

tbrisker: I'm not sure, do we? tbh I wonder if we could just drop mysql testing completely, i don't expect any incompatible changes will get cherry picked to stable branches

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman.groovy
tbrisker: this is used by `release/source/foreman.groovy`, where i removed the mysql stage, and `release/source/katello.groovy` and `test/testKatello.groovy` which don't support mysql already so i think it's pretty safe

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-plugins-pull-request.yaml
tbrisker: Looks like we can use this job for testing plugins against stable branches, but I believe that it is fairly rare occurrence where a plugin is tested against a stable core branch, and even rarer when a change in a plugin would cause a failure on mysql when tested against the stable branch (since we don't want to continue testing on mysql against develop).

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-plugins-pull-request.yaml
```diff
@@ -31,15 +31,14 @@
           values:
             - postgresql
             - sqlite3
-            - mysql
```
ekohl: I think this needs to stay in if we want to keep testing it for stable jobs.

tbrisker: ok, i reverted the change to this file

tbrisker:
@ekohl removed the change from the plugin test, can you please take another look?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/foreman-plugins-pull-request.yaml
ekohl: Looking back at it, it's easier to just drop mysql altogether. The combination filters will just be very complex.

ekohl:
Sorry for wanting to test it on PRs and then going back on it. I now think that just relying on the test_1_x_stable jobs is probably good enough and they are separate. You'll just have to be careful with cherry picks. The risk is minimal enough since the PRs that might run into it are usually too complex to cherry pick.

tbrisker:
@ekohl no worries, pushed again with dropping mysql testing completely from plugins

PR #1188, opened 2019-11-15
Remove old ensure => absent resources

ekohl:
These were added a while ago and now we can assume they're all cleaned up.


ekohl:
Updated so the tests should be happier.

PR #1189, opened 2019-11-15
Introduce a unified mashing script

ekohl:
This new script combines the functionality of all previous scripts. To do so, it introduces a parameter collection which allows calling the appropriate mashing code.

It highlights the differences between the various mashing scripts but further unification is for a later stage.


evgeni:
This needs a rebase after https://github.com/theforeman/foreman-infra/pull/1192 and https://github.com/theforeman/foreman-infra/pull/1194

koji/collection-mash-split.py
```diff
@@ -328,34 +332,134 @@ def mash_split(self, whole_path, tmp_path, split_path, extras_path, mash_config,
...
+        if collection == 'foreman':
+            CONFIG["extras_baseloc"] = "%(tag)s:extras"
+            CONFIG["extras_path"] = "/mnt/koji/mash/extras/foreman"
+            extras = ["extras-foreman-rhel7"]
+        else:
+            extras = []
+
+        mashes = [
+            (
+                "{}-{}-rhel7-dist".format(collection, version),
```
evgeni: I wonder if this section would be more readable if we had smth like this:

```python
def mash_config_name(collection, version, suffix):
    return "{}-{}-{}".format(collection, version, suffix)

def comps_name(collection, suffix):
    return "comps-{}-{}.xml".format(collection, suffix)

(etc)
```

and then use it like:

```python
mashes = [
    (
        mash_config_name(collection, version, "rhel7-dist"),
        {comps_name(collection, "rhel7"), comps_path(collection, version, 'RHEL/7')},
        extras,
    ),
]
```

evgeni:
overall I like this a lot! my knowledge around the mashing process is quite limited, so I'd rather not just ACK it, but from reading over it I think it should do the same as before, while keeping everything in one place.

ekohl:
Rebased and addressed the conflict.

evgeni:
@ekohl seems still to conflict in katello-mash-split :(

ehelms:
Is this something that is still being pursued?

evgeni:
So @pcreech wanted to move the whole process to a modern pungi setup instead of mash, but this is a long(er) term plan with no exact deadline.
I think that the unified mashing script can be beneficial before that and also lead to a clearer pungi setup, so I'd be willing to take this over if @ekohl doesn't want/have time for it.

pcreech:
@evgeni is correct.  This seems to be a good bridge to the pungi solution.  Adopting pungi across all koji generated repos will take time.

I'd also prefer this initiative is driven forward for the short term

ekohl:
Rebased to fix conflicts. https://github.com/theforeman/foreman-infra/pull/1250 will introduce another trivial one, but I'd rather merge this after 2.0.0-RC1 has gone out since we're already behind schedule on that.

ekohl:
Updated to resolve conflicts

ekohl:
I have added some commits that try to extract common functionality out. They're separate commits so you can see the difference. Let me know what you think is the best version. Ideally we'd change tags and files so they consistently use the same codes and it can be simplified. `rhel7` vs `el7` vs `RHEL/7` is confusing.

ehelms:
I was wondering if maybe the mash script shouldn't care about dist but rather that be an input to allow control by the tools calling the script. For example:

```
collection-mash-split.py katello nightly el7
collection-mash-split.py foreman nightly el7
collection-mash-split.py client nightly el7
collection-mash-split.py client nightly el6
```

This could do a bit of verification as well to ensure that a mash config exists for that combination and error nicely if it does not.

ekohl:
I've been thinking about that and would like that. Then we shift the knowledge of distros to Jenkins rather than the mash script. We need a translation layer and IMHO the script should now accept normalized versions (`el7`) so Jenkins doesn't care about individual tags. Do you agree?

ehelms:
If the script does the translation, the script has to know the nuances of each version which is part of what I wanted to avoid so that the Jenkins jobs that do focus on version can handle the differences.

ekohl:
If you look at how MashConfig is called, there's still a lot of logic there. I would like to smooth out these differences by changing koji tags and staging repository changes (at least for nightly). By tackling the problem at the source, I think we'll reach the best result.

ehelms:
I agree with smoothing this out via tag changes looking ahead. If we wait for that, it may be a while before this could then get merged. I can live with this script knowing the logic, I just tend to find it cleaner for Jenkins to be able to drive calling the mash script to specify project, version and dist tag identifier.

zjhuntin:
@ekohl sorry for coming in late to this discussion, but what are the ways you're proposing changing the tags to account for how this script is called? (If i'm interpreting what you're saying correctly).

evgeni:
@zjhuntin there is a long standing plan to rename the "rhel" tags to "el", so that it becomes uniform with the dist tag in the RPM and the folder on yum.tfm. I guess that's what Ewoud meant.

Ad the discussion: I agree that this script becomes cleaner if we had clean tags. We don't. So we could fix the tags in nightly (and 2.1+) and have this script only do these (and the old scripts do the old releases). Or we keep some more logic here, until we cleaned the tags.

If the "more logic" makes us merge this sooner, I'm all for more logic. It can be removed later.

ekohl:
Rebased.

ekohl:
So is there agreement this can be merged now? Anything I need to do?

pcreech:
I'm good with going forward with merging.  To summarise the situation from the comments across the PR:

- We are trying to preserve branch/tag matrices with this change, to facilitate quicker adoption of this script
- We will be working to adopt koji tag structures and jenkins job configs that will help facilitate removing complex logic that currently will live with this script.
- Eventually this will all be hopefully replaced with Pungi when we complete the koji rearchitecture (no ETA yet)

I think that sums everything up.  LGTM, lets deploy it and get it working!

ekohl:
Deployed now

PR #1190, opened 2019-11-17
Drop xenial from nightly pipeline

tbrisker:


ekohl:
I'd prefer a more complete PR that actually removes it from all package building.

tbrisker:
@ekohl wouldn't that also prevent us from building it on the stable branches?

tbrisker:
closed for #1193

PR #1192, opened 2019-11-20
Mash pulpcore as well

pcreech:


koji/katello-mash-split.py
```diff
@@ -317,14 +317,18 @@ def main():
     mash_config = "katello-{}-rhel7".format(version)
     if version == 'nightly':
         mash_config_candlepin = "katello-thirdparty-candlepin-rhel7"
+        mash_config_pulp = "katello-nightly-pulpcore-el7"
```
ekohl: Isn't it sufficient to do this outside of the if? This is equivalent to `"katello-{}-pulpcore-el7".format(version)`

koji/katello-mash-split.py
```diff
@@ -317,14 +317,18 @@ def main():
...
     else:
         mash_config_candlepin = "katello-{}-thirdparty-candlepin-rhel7".format(version)
+        mash_config_pulp - "katello-{}-pulpcore-el7".format(version)
```
ekohl:
```suggestion
        mash_config_pulp = "katello-{}-pulpcore-el7".format(version)
```

koji/katello-mash-split.py
```diff
@@ -317,14 +317,18 @@ def main():
...
 
     output_paths = ["katello-{}/katello/el7".format(version)]
     output_paths_candlepin = ["katello-{}/candlepin/el7".format(version)]
+    output_paths_pulp = ["katello-{}/candlepin/el7".format(version)]
```
ekohl:
```suggestion
    output_paths_pulp = ["katello-{}/pulpcore/el7".format(version)]
```

koji/katello-mash-split.py
pcreech: Yup, at the time, it felt more consistent for some reason.  But definitely can move it up.

Also, something I just thought about, we aren't preventing this on any versions outside of nightly at the moment... so it may crash older katello mashes

koji/katello-mash-split.py
```diff
@@ -315,16 +315,20 @@ def main():
...
     mash_config = "katello-{}-rhel7".format(version)
+    mash_config_pulp = "katello-{}-pulpcore-el7".format(version)
     if version == 'nightly':
         mash_config_candlepin = "katello-thirdparty-candlepin-rhel7"
     else:
         mash_config_candlepin = "katello-{}-thirdparty-candlepin-rhel7".format(version)
 
     output_paths = ["katello-{}/katello/el7".format(version)]
     output_paths_candlepin = ["katello-{}/candlepin/el7".format(version)]
+    output_paths_pulp = ["katello-{}/pulp/el7".format(version)]
```
evgeni: shouldn't this be `pulpcore`?

we used to have `pulp` in 2.x times: https://fedorapeople.org/groups/katello/releases/yum/2.4/pulp/, so maybe something new would be better? :)

pcreech: Yup, good catch

PR #1193, opened 2019-11-21
Drop Xenial and Stretch from nightly

ekohl:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_plugin.yaml
```diff
@@ -12,7 +12,7 @@
           type: user-defined
           name: os
           values:
-          - xenial
+          - buster
```
ekohl: Do we want to stay on Ubuntu based machines for plugin building?

mmoll: Yes, because bionic is a tad older than bionic and we need to stay compatible.

ekohl:
Updated to use bionic in plugin building.

mmoll:
merged, bedankt @ekohl!

PR #1195, opened 2019-12-02
Add timing trigger to ci.centos job updater

zjhuntin:


ekohl:
I'm wondering if we want git triggers as well.

zjhuntin:
@ekohl If there were a way we could trigger it every time a merge request was merged into foreman-infra I'd prefer that. This however kind of works and at least will do the job in the mean time.

ekohl:
We already have one job that works on foreman-infra with a trigger:
https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/puppet-deploy.yaml

zjhuntin:
@ekohl Updated with that method! Thanks for pointing that out.

PR #1197, opened 2019-12-04
purge yum repos from the CDN after publishing them

evgeni:
this will purge a bit more than we want, but it's safer this way

### DIFF
--- a/deploy/config/deploy/yum.rb
+++ b/deploy/config/deploy/yum.rb
@@ -17,6 +17,8 @@ set :repo_path, "#{deploy_to}/#{repo_dest}"
 # hidden directory
 set :repo_instance_path, "#{deploy_to}/#{File.dirname(repo_dest)}/.#{File.basename(repo_dest)}-#{Time.now.utc.strftime("%Y%m%d%H%M%S")}"
 
+set :repo_rsync_log, "/tmp/rsync-#{File.dirname(repo_dest)}-#{File.basename(repo_dest)}-#{Time.now.utc.strftime("%Y%m%d%H%M%S")}.log"
+
 role :web, host
 
 ssh_options[:forward_agent] = true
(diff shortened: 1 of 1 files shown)


deploy/config/deploy/yum.rb
```diff
@@ -63,4 +64,8 @@
       run "if [ -e #{repo_instance_path}-previous ]; then rm -rf #{repo_instance_path}-previous; fi" if overwrite || merge
     end
   end
+
+  task :purgecdn do
+    run "find #{repo_path} | sed 's|#{deploy_to}|https://yum.theforeman.org/|' | xargs curl -X PURGE -H 'Fastly-Soft-Purge:1'"
```
evgeni: it would be nice if we could only purge things that changed, not *everything*

evgeni: one possibility would be to use `find … -mtime -3` to find any files that have been modified in the last three days.
as the modification time is pulled from Koji (via rsync), it might be older than "just now".

deploy/config/deploy/yum.rb
```diff
@@ -63,4 +64,8 @@
       run "if [ -e #{repo_instance_path}-previous ]; then rm -rf #{repo_instance_path}-previous; fi" if overwrite || merge
     end
   end
+
+  task :purgecdn do
+    run "find #{repo_path} -mtime -3 | sed 's|#{deploy_to}|https://yum.theforeman.org/|' | xargs --no-run-if-empty curl -X PURGE -H 'Fastly-Soft-Purge:1'"
```
evgeni: one could also use something like `rsync … --log-file rsync.log --log-file-format "CHANGED %o %f"` and then `grep` out the `CHANGED` lines from the log and purge these names

ekohl: Looks like `%i` can tell whether it was actually changed. Is it safe enough to use `%i %f` and grep for c in the first column? Untested, but  `awk '$1 ~ /c/ { print $2 }' rsync.log`

evgeni: you line looks like this:

```
2019/12/04 10:13:22 [13933] >f+++++++++ sles12/x86_64/subscription-manager-1.23.3-4.1.suse1315.x86_64.rpm
```

and for comparison, mine:

```
2019/12/04 09:50:19 [13139] CHANGED recv sles12/x86_64/subscription-manager-1.23.3-4.1.suse1315.x86_64.rpm
```

ekohl: After some manual testing it looks like yours is good enough. It does look like `--log-file` appends so we should make sure it's clean before we start. Other than that I think the rsync approach is nice. Given the foreman-release handling is manual, you probably need to purge those explicitly.

evgeni: That's exactly what I thought too, so updated this PR to match that.

deploy/config/deploy/yum.rb
```diff
@@ -63,4 +68,10 @@
       run "if [ -e #{repo_instance_path}-previous ]; then rm -rf #{repo_instance_path}-previous; fi" if overwrite || merge
     end
   end
+
+  task :purgecdn do
+    run %Q{awk '/ CHANGED /{print "https://yum.theforeman.org/"#{repo_dest}$5}' #{repo_rsync_log} | xargs --no-run-if-empty curl -X PURGE -H 'Fastly-Soft-Purge:1'}
```
ekohl: I think the script can be called with `replace` instead of `rsync` in which case the file will not exist. Probably good to check if it exists.

evgeni: not sure what you mean. there is a `replace` task, but if that is called, the `purgecdn` task is not called either. and if one calls the `sync` task, the `rsync` task is always executed.

ekohl: So I was mixing things up and confused. Tasks are always executed but we do have parameters to tune:

https://github.com/theforeman/foreman-infra/blob/2cc59353f8d74943e687fac2724967e2b40176e8/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy#L12

PR #1200, opened 2019-12-06
Added docs vhost to web02

lzap:
As a puppet hater, I have to say this is where it pays off.

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -33,6 +33,7 @@ class web(
       'stagingdeb.theforeman.org',
       'www.theforeman.org',
       'yum.theforeman.org',
+      'docs.theforeman.org',
     ],
     webroot_paths => [
       '/var/www/vhosts/web/htdocs',
@@ -42,6 +43,7 @@ class web(
       '/var/www/vhosts/stagingdeb/htdocs',
       '/var/www/vhosts/web/htdocs',
       '/var/www/vhosts/yum/htdocs',
+      '/var/www/vhosts/docs/htdocs',
     ],
   }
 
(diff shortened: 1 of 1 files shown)


lzap:
Oh right, rebased. I don't see any tests?!

ohadlevy:
I've created docs.tf.org dns entry, so once merged this should be live.

mmoll:
let's try this...

mmoll:
merged, díky @lzap!

puppet/modules/web/manifests/init.pp
```diff
@@ -129,6 +131,24 @@
...
   }
 
+  # DOCS
+  $docs_directory = '/var/www/vhosts/docs/htdocs'
+  $docs_directory_config = [
+    {
+      path            => $docs_directory,
+      options         => ['Indexes', 'FollowSymLinks', 'MultiViews'],
+      expires_active  => 'on',
+      expires_default => 'access plus 4 hours',
```
ekohl: Is it wise to set caching headers? This depends on the build process and probably safer to start without.

puppet/modules/web/manifests/init.pp
```diff
@@ -129,6 +131,24 @@
...
+    },
+  ]
+
+  $docs_attrs = {
+    servername   => 'docs.theforeman.org',
+    docroot      => $docs_directory,
+    docroot_mode => '2575',
```
ekohl: That's a weird mode. Why is it writable for the group but not the user? We have a special case with yum, but docs should be more like `website`. I'd also prefer a specific user that owns the docroot so we don't have to abuse root for it.

puppet/modules/web/manifests/init.pp
lzap: I copy-pasted from the yum example, do you want me to remove it?

puppet/modules/web/manifests/init.pp
lzap: Yeah I picked a really special example. Let me fix it in a new PR.

lzap:
Filed https://github.com/theforeman/foreman-infra/pull/1202

PR #1201, opened 2019-12-06
Use nodesource repositories version 12 on slaves

ekohl:


puppet/modules/slave/manifests/init.pp
```diff
@@ -150,13 +150,13 @@
...
   if $::osfamily == 'RedHat' {
-    package { 'npm':
-      ensure  => present,
-      require => Package['bzip2'],
-    } -> Package <| provider == 'npm' |>
+    class { 'nodejs':
+      repo_url_suffix       => '12.x',
+      nodejs_package_ensure => latest,
+      npm_package_ensure    => latest,
+    }
```
mmoll: `-> Package <| provider == 'npm' |>` is IMHO still needed, or does the nodejs class handle this on its own?

ekohl: Updated

puppet/modules/slave/manifests/init.pp
```diff
@@ -150,12 +150,12 @@
...
-  # packages only really available on EL6+
+  # nodejs/npm for JavaScript tests
   if $::osfamily == 'RedHat' {
-    package { 'npm':
-      ensure  => present,
-      require => Package['bzip2'],
+    class { 'nodejs':
+      repo_url_suffix       => '12.x',
+      nodejs_package_ensure => latest,
+      npm_package_ensure    => latest,
```
mmoll: needs to be `absent`, as the `nodejs` package is already containing it.

ekohl: Updated. Also explicitly set `manage_repo` to true.

mmoll: `Class[Nodejs]: has no parameter named 'manage_repo'` :upside_down_face:

mmoll:
merged, bedankt @ekohl!

PR #1202, opened 2019-12-12
Fixed docs vhost for web02

lzap:
Some overlooks: https://github.com/theforeman/foreman-infra/pull/1200

### DIFF
--- a/puppet/modules/web/manifests/init.pp
+++ b/puppet/modules/web/manifests/init.pp
@@ -135,18 +135,18 @@ class web(
   $docs_directory = '/var/www/vhosts/docs/htdocs'
   $docs_directory_config = [
     {
-      path            => $docs_directory,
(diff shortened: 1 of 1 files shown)


puppet/modules/web/manifests/init.pp
```diff
@@ -137,16 +137,16 @@
     {
       path            => $docs_directory,
       options         => ['Indexes', 'FollowSymLinks', 'MultiViews'],
```
ekohl: Please correct the alignment here (and in the attrs) to avoid the empty whitespace.

lzap: Done.

lzap:
Looks like puppet executed already :-)

https://docs.theforeman.org/

We are online.

mmoll:
merged, díky @lzap!

PR #1203, opened 2019-12-13
Drop libvirt-slave and promoted-builds plugins

ekohl:
These plugins are actually unused and libvirt-slave contains a security issue.


mmoll:
merged, bedankt @ekohl!

PR #1207, opened 2019-12-16
Move katello nightly RPM pipeline to 3 hour timeout

ehelms:


evgeni:
You might want to adjust https://github.com/theforeman/foreman-infra/blob/82ee23b15863e18d538a6ee7e12bcc3293a84eed/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/katello/nightly.groovy#L10?

ehelms:
Updated

PR #1208, opened 2019-12-16
properly quote the string inside the awk call

evgeni:
also call curl with --silent

### DIFF
--- a/deploy/config/deploy/yum.rb
+++ b/deploy/config/deploy/yum.rb
@@ -70,8 +70,8 @@ namespace :repo do
   end
 
   task :purgecdn do
-    run %Q{awk '/ CHANGED /{print "https://yum.theforeman.org/"#{repo_dest}$5}' #{repo_rsync_log} | xargs --no-run-if-empty curl -X PURGE -H 'Fastly-Soft-Purge:1'}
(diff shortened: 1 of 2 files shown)


deploy/config/deploy/yum.rb
```diff
@@ -70,8 +70,8 @@
   end
 
   task :purgecdn do
-    run %Q{awk '/ CHANGED /{print "https://yum.theforeman.org/"#{repo_dest}$5}' #{repo_rsync_log} | xargs --no-run-if-empty curl -X PURGE -H 'Fastly-Soft-Purge:1'}
-    run %Q{for d in #{repo_instance_path}/*; do purge_base="https://yum.theforeman.org/#{repo_dest}/$(basename $d)"; curl -X PURGE -H 'Fastly-Soft-Purge:1' $purge_base/foreman-release.rpm $purge_base/foreman-client-release.rpm; done}
+    run %Q{awk '/ CHANGED /{print "https://yum.theforeman.org/#{repo_dest}"$5}' #{repo_rsync_log} | xargs --no-run-if-empty curl --silent -X PURGE -H 'Fastly-Soft-Purge:1'}
```
evgeni: a quote makes a nice `fatal: division by zero attempted` error :/

PR #1210, opened 2019-12-17
Drop build-metrics and global-build-stats

ekohl:
These plugins are unused and contain XSS vulnerabilities. The global-build-stats is also unused since it was only installed as a dependency for build-metrics.


mmoll:
merged, bedankt @ekohl!

PR #1213, opened 2019-12-19
Add packaging release jobs for other branches

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-packaging-rpm-release.yaml
```diff
@@ -0,0 +1,29 @@
+---
```
ekohl: Does it make sense to include this in https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-release.yaml since that already controls all the version specific pipelines? It'd mean there's just one place to drop a release and add a new one.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-packaging-rpm-release.yaml
```diff
@@ -0,0 +1,29 @@
...
+      !include-raw:
+        - 'pipelines/vars/foreman/{version}.groovy'
+        - 'pipelines/release/foreman_packaging_release.groovy{empty}'
+        - 'pipelines/lib/git.groovy{empty}'
+        - 'pipelines/lib/ansible.groovy{empty}'
+        - 'pipelines/lib/obal.groovy{empty}'
+        - 'pipelines/lib/rpm_packaging.groovy{empty}'
+
+- project:
+    name: foreman
```
evgeni:
```
jenkins_jobs.errors.JenkinsJobsException: Duplicate entry found in '/home/travis/build/theforeman/foreman-infra/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-packaging-rpm-release.yaml: 'foreman' already defined
```

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/release/foreman-packaging-rpm-release.yaml
evgeni: agreed.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-release.yaml
```diff
@@ -60,10 +60,7 @@
       - 'foreman-plugins-{version}-rpm-pipeline'
     empty: ''
     version:
-      - '1.21'
-      - '1.22'
-      - '1.23'
-      - '1.24'
+      !include: ../../includes/foreman_versions.yaml.inc
```
ehelms: Thoughts on this approach to keep the jobs split out but achieve the single version location? See https://github.com/theforeman/foreman-infra/pull/1213/files#diff-ac0a597bb541cb1ebc55f7e2c6ab538dR27 as well

evgeni: I like. I'd like it even more if we could get rid of the odd nighly jobs, but that's not possible at the moment as far as I see it.

evgeni:
two comments (one I already mentioned on IRC, but wanted to leave it here for others to see):
* for foreman/proxy/installer/selinux this is already handled by release_packages and is now kinda duplicated? it doesn't hurt, release_packages will just skip those that are already built, but one should be aware of this fact.
* this will fail on packages that don't have `foremandist` and thus build identical NEVRs in different branches (and we usually koji tag-build those). again, this is fine, as long as the person merging and seeing the failed build knows what to do instead.

ehelms:
For the first point, I would opt for monitoring how that plays out for the releases going forward (probably first z-stream of 1.24).

For the second point, I did change this to notify discourse on failures to increase visibility. I also feel like you raise a good situation that our obal release code should probably handle. Detecting if a build exists and performing a tag.

evgeni:
Agreed.

Also, we really should update https://github.com/theforeman/foreman-packaging/tree/rpm/develop#releasing-a-package ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-release.yaml
ekohl: I dislike it. There's already a lot of files and I liked to have a single file that defines all the release jobs.

evgeni: but the jobs are still in foreman-release.yml, just the versions are external so that they can be reused?

ekohl: Right now there's 2 places that use it, but I'd argue that everything in `foreman-packaging-rpm-release.yaml` should also be in `foreman-release.yaml`. Then there's just one place and you can clearly see what it impacts.

evgeni: Right, so I somehow thought this would move the jobs to one file and the version definition to another file. I would *prefer* the jobs in one file too, but as strongly as Ewoud ;)

evgeni:
Needs at least a rebase too :)

ehelms:
To add a little, I look at folders as achieving this bringing things together so that individual jobs can be broken out into files for ease of understanding and finding.

The one issue I have is we are mixing our nomenclatures which has in the past led us to end up with confusion and mixing of terminology. We decided to name single release jobs with the `release` terminology and full, multi-faceted, end user facing jobs `pipelines`. The building of RPMs into Koji does one thing and does not fully carry the RPMs out to the end user. So if I put this into the `foreman-release.yml` file (which is, by these rules, incorrectly named) we now have releases with pipelines. Organizationally we now are throwing our conventions out the window. I get y'alls thinking somewhat, but I don't find it worth breaking our conventions since we established them with good reason. The broken out version file solves the problem of one place to create all the jobs we need for a Foreman release.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman_packaging_release.groovy
```diff
@@ -13,10 +13,13 @@ pipeline {
     stages {
         stage('Clone Packaging') {
             steps {
+                script {
+                    $foreman_branch = foreman_version == 'nightly' ? "rpm/develop" : "rpm/${foreman_version}"
```
evgeni: You could argue we could just set this in `pipelines/vars/foreman/{version}.groovy`, but I'm fine either way.

ehelms: I did debate that. I suppose I could call it foreman_packaging_branch but given it was localized to this pipeline I figured I'd keep it close to home.

Issue #1069, opened 2019-05-30
Change Puppet module directory structure


mmoll:
I want to collect comments from @ehelms @evgeni @ekohl for the following separation:

In the puppet `modules` folder, only modules that are available on puppet forge are kept.

In the new puppet `external_modules` folder, all modules from external git repos (e.g. dirvish) or are kept.

In the new puppet `local_modules` folder, all modules that are files in this repo, are kept.

This way, r10k can be used later for `modules` and `external_modules`, which would unclutter this repo here.

ekohl:
In general I'd be in favor of such a solution. We'd also drop the configs then.

ehelms:
I would generally also be in favor of this. When I first started reading this I thought maybe it would be an organization our concerns (e.g. jenkins related modules together). Any way to also achieve that?

mmoll:
@ehelms I'm a bit unsure how this structure would look like and what problem it would solve...

When talking about jenkins, that would be a `jenkins_modules` folder, containing at least the modules that start with "jenkins", correct?

ehelms:
I think our repository layout tries to solve two problems:

 1) providing a puppet structure that can be imported and used by a Puppet server
 2) provide a place to make changes and store them in git

For newcomers and the casual developer, #2 can be hard to understand how all the pieces fit together and where certain parts live because its a big bucket of modules. Your example of Jenkins is a good one. There are a few obvious ones: 

 * jenkins_master
 * jenkins
 * jenkins_job_builder

But also `slave` is part of this sub-set of infrastructure. Similarly, I imagine there are modules related to the the web hosts for yum/deb, our application infrastructure (prprocessor, rbot). Something like `rvm` I know is used by Jenkins, but maybe its used by more than that?

We don't have to solve this problem with your proposal. I did think it worthy of at least discussing in case there was any obvious quick wins here to help with #2.

mmoll:
@ehelms Ah, I see... Unfortunately I don't see any quick wins, but as I would want all that to happen in small steps, maybe we can throw something in later.

At the end of the day, the easiest would be to merge all modules together into one dir in the deployment job anyway, so there are a lot of possibilities...

ekohl:
[...]

PR #1215, opened 2019-12-23
Move puppet modules to specific dirs

mmoll:


mmoll:
First part of GH-1069

ekohl:
Should we add an [environment.conf](https://puppet.com/docs/puppet/latest/config_file_environment.html) to set the `modulepath` instead of solving it via rsync?

mmoll:
Unfortunately I have no idea how to handle multiple module dirs in puppetlabs_spec_helper...

ekohl:
In 3ba9d9869357cca35a614602818813bfab0b5461 I used bare rspec-puppet and puppet-syntax. Also removed the now unused `.fixtures.yml` file. Still not entirely sure about the `_core` modules since Puppet Agent bundles them. They're just needed for testing. Perhaps a `test_modules` directory would make sense for them?

mmoll:
should be ready now :rocket:

puppet/modules/deploy/files/script.sh
```diff
@@ -2,7 +2,9 @@
...
 echo "Deploy started at `date`"
 dir=`mktemp -d`
-trap "rm -rf $dir" EXIT
-git clone --recurse-submodules https://github.com/theforeman/foreman-infra $dir/
-rsync -aqx --delete-after --exclude=.git $dir/puppet/modules/ /etc/puppetlabs/code/environments/production/modules/
+trap "rm -rf ${dir}" EXIT
+git clone --recurse-submodules https://github.com/theforeman/foreman-infra ${dir}/
+prod_dir="/etc/puppetlabs/code/environments/production"
+rsync -aqx --delete-after --exclude=.git ${dir}/puppet/*modules ${prod_dir}/
+echo 'modulepath = forge_modules:git_modules:modules:$basemodulepath' >${prod_dir}/environment.conf
```
ekohl: Should we just include a static `environment.conf` in the repo?

puppet/spec/spec_helper.rb
```diff
@@ -1,11 +1,35 @@
...
+      'operatingsystem'        => 'CentOS',
+      'operatingsystemrelease' => ['7'],
+    },
+    {
+      'operatingsystem'        => 'Debian',
+      'operatingsystemrelease' => ['10', '9'],
+    },
+    {
+      'operatingsystem'        => 'Ubuntu',
+      'operatingsystemrelease' => ['16.04'],
```
ekohl: Should we update this to 18.04 since we upgraded?

puppet/modules/deploy/files/script.sh
```diff
@@ -2,7 +2,9 @@
 set -e # dont rsync if clone fails
 echo "Deploy started at `date`"
 dir=`mktemp -d`
-trap "rm -rf $dir" EXIT
-git clone --recurse-submodules https://github.com/theforeman/foreman-infra $dir/
-rsync -aqx --delete-after --exclude=.git $dir/puppet/modules/ /etc/puppetlabs/code/environments/production/modules/
+trap "rm -rf ${dir}" EXIT
+git clone --recurse-submodules https://github.com/theforeman/foreman-infra ${dir}/
+prod_dir="/etc/puppetlabs/code/environments/production"
+rsync -aqx --delete-after --exclude=.git ${dir}/puppet/*modules ${prod_dir}/
```
mmoll: This might deleta a bit too much, but I don't know if recent puppetserver releases are tripping over non-existing `data` and `manifests` folders...

TBH, after all this wrangling I see the idea of just syncing all modules into one dir at deployment much more feasible now. ;)

ekohl: IMHO this is mostly a temporary solution and we should use r10k to deploy after this since it has caching built in as well.

mmoll:
rebased.

mmoll:
how's this looking now? :)

PR #1218, opened 2019-12-31
Fixes #28589 - Test proxy with Ruby 2.7

mmoll:


mmoll:
Locally this is coming back green, despite a few new warnings (all in dependencies, AFAICT). For plugins, I'd just extend the matrix and see, if something new pops up...

PR #1219, opened 2019-12-31
Generate a gemset per executor

ekohl:
When using parallel pipelines in Jenkins, a single job can run multiple times on the same node. They can be identified by a combination of `BUILD_TAG` (`JOB_NAME`-`BUILD_ID`) and `EXECUTOR_NUMBER`. It can stay simplified if the number is 0. Note they are strings and not integers.


mmoll:
merged, bedankt @ekohl!

PR #1146, opened 2019-10-13
adjust workspace directory

mmoll:

PR #1220, opened 2019-12-31
Use variables for directories

ekohl:
Also ensures the home directory exists.

I know https://github.com/theforeman/foreman-infra/pull/1146 also exists and I'm not entirely sure how we should approach this. This is probably a good start.


mmoll:
merged, bedankt @ekohl!

PR #1223, opened 2020-01-01
Drop bastion testing

ekohl:
The bastion project has been archived so there's no more need for CI


mmoll:
merged, bedankt @ekohl!

PR #1224, opened 2020-01-01
Drop phantomjs on slaves

ekohl:
It is no longer needed in tests and doesn't install on recent nodejs versions.

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -188,10 +188,6 @@ class slave (
       ensure   => '1.7.9',
       provider => npm,
     }
-    package { 'phantomjs':
-      ensure   => latest,
-      provider => npm,
-    }
     package { 'grunt-cli':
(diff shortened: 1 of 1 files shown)


mmoll:
AFAICT, the `angular-ui` tests of the Katello pipeline are still using it.

ekohl:
@waldenraines might be able to answer that.

mmoll:
Waiting on https://github.com/Katello/katello/pull/8488

mmoll:
merged, bedankt @ekohl!

puppet/modules/slave/manifests/init.pp
```diff
@@ -188,10 +188,6 @@
       ensure   => '1.7.9',
       provider => npm,
     }
-    package { 'phantomjs':
-      ensure   => latest,
```
ekohl: Should we ensure absent perhaps to make sure things don't accidentally count on it?

PR #1226, opened 2020-01-02
Fixes #28601 - Test Hammer with Ruby 2.7

mmoll:


mmoll:
@zjhuntin Could this also go in?

PR #1227, opened 2020-01-03
Refactor Foreman plugins to pipeline plugins

ekohl:
This is an effort to remove the workflow jobs in favor of pipelines. Currently still a bit of WIP. It doesn't convert virt-who yet since that needs katello and does some odd things. Also includes some other commits I needed to provision a fresh Jenkins master + slave.


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/foreman_plugin.groovy
```diff
@@ -0,0 +1,121 @@
...
+                                addGem()
+                                databaseFile(gemset())
+                                configureDatabase(ruby)
+                            }
+                        }
+                    }
+                    stage('Unit Tests') {
+                        steps {
+                            dir('foreman') {
+                                withRVM(['bundle exec rake jenkins:unit TESTOPTS="-v" --trace'], ruby)
```
ekohl, line 51: Is this really the way to test a plugin?

ShimShtein: I think this will run all tests (including foreman core ones).
For my plugin, I test with `rake test:foreman_rh_cloud`.

ekohl: That's exactly my point, but I think currently this is how we test in Jenkins since I copied the builder.

ShimShtein: So maybe it makes sense to make core tests at least optional? So I will be able to decide for my plugin if I want to run core tests too. For smaller plugins that do not modify core objects it should be OK to run only plugin related tests.

ekohl: This is still a draft for now, so this line was partly a TODO so I wouldn't forget.

ehelms:

ekohl:
Replaced by https://github.com/theforeman/jenkins-jobs/pull/11

PR #1228, opened 2020-01-03
Drop fpm scripts for freight

ekohl:
There are proper packages available and FPM is no longer needed.


mmoll:
merged, bedankt @ekohl!

PR #1229, opened 2020-01-03
Install wget to fix RVM gpg key import

mmoll:


mmoll:
Regardless of this specific problem, there seems to be no current Puppet module handling RVM, so it's probably time for theforeman or voxpupuli to take over, or are there other alternatives (docker?) for our infra?

PR #1230, opened 2020-01-03
Adjust Rubygems versions

mmoll:


mmoll:
At the moment, new installs are semi-broken anyway, as Rubygems 3.0.6 gets installed on all Ruby versions, despite being supported only on 2.3+.

We really should think what needs to be done to drop everything before 2.5 in all projects, as e.g. older Rubies can't get installed with the current RVM mechanics on at least CentOS 8.

PR #1231, opened 2020-01-03
Switch CentOS 7 to rh-postgresql10 SCL

mmoll:


puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
+        version              => '10',
+        client_package_name  => 'rh-postgresql10-postgresql-syspaths',
+        server_package_name  => 'rh-postgresql10-postgresql-server-syspaths',
+        contrib_package_name => 'rh-postgresql10-postgresql-contrib-syspaths',
+        service_name         => 'postgresql',
+        datadir              => '/var/opt/rh/rh-postgresql10/lib/pgsql/data',
+        confdir              => '/var/opt/rh/rh-postgresql10/lib/pgsql/data',
+        bindir               => '/usr/bin',
+      }
+      file { '/usr/bin/pg_config':
```
mmoll: It seems there's no `rh-postgresql10-postgresql-devel-syspaths`, which would bring this..

puppet/modules/slave/manifests/init.pp
```diff
@@ -101,8 +101,11 @@
     'postgresql-dev':
```
ekohl: Should be use `postgresql::lib::devel` from puppetlabs-postgresql?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
 # @api private
 class slave::postgresql {
-  Postgresql_psql {
-    cwd => '/',
-  }
+  # only CentOS slaves are used to run unit tests
+  if $::osfamily == 'RedHat' {
+    if ::operatingsystemmajrelease == '7' {
+      include scl
```
ekohl: Why is this needed? The whole point of the `-syspaths` packages is that you don't need to explicitly enable them.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
-  # Simple known user/pass that will allow Jenkins to create its
-  # own databases when required
-  postgresql::server::role { 'foreman':
-    password_hash => postgresql_password('foreman', 'foreman'),
-    superuser     => true,
-    login         => true,
-  }
+      $non_scl_rpms = ['postgresql', 'postgresql-server', 'postgresql-devel', 'postgresql-libs']
+      $non_scl_rpms.each |String $non_scl_rpm| {
+        package { "non_scl_${non_scl_rpm}":
```
ekohl: You can also specify it at once. Probably also good to ensure it happens before postgresql is ran.
```ruby
package { $non_scl_rpms:
  ensure => absent,
  before => Class['postgresql::globals'],
}
```

puppet/modules/slave/manifests/postgresql.pp
ekohl: I think that if you install `rh-postgresql10-syspaths` this shouldn't be needed.

puppet/modules/slave/manifests/postgresql.pp
mmoll: Nope...
```
Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Package[postgresql-devel] is already declared at (file: /root/foreman-infra/puppet/modules/postgresql/manifests/lib/devel.pp, line: 21); cannot redeclare (file: /root/foreman-infra/puppet/modules/slave/manifests/postgresql.pp, line: 19) (file: /root/foreman-infra/puppet/modules/slave/manifests/postgresql.pp, line: 19, column: 1) on node localhost.localdomain
```

puppet/modules/slave/manifests/init.pp
mmoll: If it work with SCLs, yes... setting the devel_package_name seems not to suffice..

puppet/modules/slave/manifests/postgresql.pp
mmoll: At least in my test VM SCLs are not enabled by default, so nothing can get installed.

puppet/modules/slave/manifests/postgresql.pp
mmoll: Nope...
```
[root@localhost ~]# rpm -q rh-postgresql10-syspaths
rh-postgresql10-syspaths-3.1-1.el7.x86_64
[root@localhost ~]# which pg_config
/usr/bin/which: no pg_config in (/sbin:/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin:/usr/local/rvm/bin)

```

puppet/modules/slave/manifests/postgresql.pp
ekohl: Perhaps it's easier to just install `centos-release-scl-rh` then to get the repos?

ekohl:
Btw, in the SCL config we still build with the old libs. It should be possible to use the non-SCL libs but SCL server since the wire protocol hasn't changed. Perhaps that's an easier goal and more realistic of the actual deployment?

puppet/modules/slave/manifests/postgresql.pp
mmoll: I think I need to test all that also on RHEL first, as we do have the AWS slaves on RHEL.

ekohl: On RHEL I think it's a separate subscription. Not entirely sure.

mmoll:
This should work now... on the AWS slaves I installed the `centos-release-scl-rh` RPM manually.

Locally I had postgresql not starting up when SELinux was enabled, but frankly I'm not willing to care about this. :broken_heart:

ehelms:
Might as well upgrade to 12 if thats easy enough, thats what we are moving nightly to now that they released the SCL for PG12.

ekohl:
@ehelms the AWS machines are RHEL, not CentOS - do you know if we can use the RH SCLs rather than CentOS on them?

mmoll:
I'm leaving this to the RH experts, because ENOTIME :crying_cat_face:

PR #1232, opened 2020-01-06
Rename foreman-release to foreman-release-pipelines

ehelms:


ekohl:
I figured this wasn't needed because it's in the pipeline directory.

ehelms:
Then we should drop the naming on all our other jobs I would think. Since we have a concept of `release` grouped jobs, this makes it clearer what this file actually holds.

ekohl:
Recently I've been looking at the files and I think it's very complicated. I'd even say that currently the directories don't make sense at all. Various styles and conventions are used and there's no consistency.

IMHO we should rewrite all jobs from workflow to pipeline. That removes the need for many top level directories since `builders`, `parameters`, `publishers`, `scm` and `wrappers` will be gone. During the refactor various jobs will disappear (`test_plugin_matrix` etc) because a job is just a single job instead of calling other jobs. Should we revisit the file naming after that?

ehelms:
That's a lot of work laid out and will take some time to achieve. As we convert, we established conventions to guide in the renaming since prior while there were some conventions things were also a bit wild wild west. I'd be happy to revisit the naming conventions later, but we have conventions now for things that we've built as pipelines, so if we could, to reduce confusion, let's say we follow those conventions?

ehelms:
@evgeni mind adding your thoughts to the conversation? I can live with whatever consensus is. I just get confused trying to find the things I am looking for.

evgeni:
I don't mind the current state (I just `git grep` stuff when searching), but I agree the file you touched stands out a bit. Whether you rename it to `-pipeline(s)`, or drop that suffix altogether is fine with me. As other folders also have their files suffixed with the folder name, I'd lean to add it here for now too.

evgeni:
@ekohl is there anything particularly blocking *this specific* PR? I do agree that we should get rid of many of those folders in the long run, but here we have a trivial improvement :)

ekohl:
No, I guess this doesn't make it worse than it already is and if it helps @ehelms then sure.

PR #1235, opened 2020-01-11
Handle UNSTABLE as a job result

ekohl:
When the build steps pass but the test results fail, it's marked as UNSTABLE. Without this, the code will continue polling until it times out.


mmoll:
merged, bedankt @ekohl!

PR #1236, opened 2020-01-13
add el8 to client release config

evgeni:


koji/foreman-client-mash-split.py
```diff
@@ -304,6 +304,7 @@ def main():
     split_path = "/mnt/koji/releases/split"
 
     dists = {
+        "el8": "el8",
```
ekohl: See how easy it is when you follow the same conventions everywhere? :)

PR #1237, opened 2020-01-13
Collect smoker logs on all pipelines + refactor

ekohl:
* a2fd8ac4 Add {install,upgrade}PipelineVars()

All install and upgrade pipelines are the same and this reduces the duplication. It also makes sure that smoker logs are collected in every single pipeline rather than just nightly-centos7.

* 00fd1254 Drop nightly debian9 vars

a1750fa620044620a610ae346ffb3768ea4ad5f3 dropped Debian 9 nightly builds but left this file.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
@@ -1,0 +1,27 @@
+// action, type, version, os, extra_vars = nil
+def pipelineVars(Map args) {
+    if (args.action == 'install') {
+        boxes = ["pipeline-${args.type}-server-${args.version}-${args.os}", "pipeline-${args.type}-smoker-${args.version}-${args.os}"]
+    } else if (args.action == 'upgrade') {
+        boxes = ["pipeline-upgrade-${args.type}-${args.version}-${args.os}", "pipeline-upgrade-${args.type}-smoker-${args.version}-${args.os}"]
+    } else {
+        boxes = []
+    }
+
+    extra_vars = [
+        'pipeline_version': args.version,
+        'pipeline_os': args.os,
+        'pipeline_type': args.type
+    ]
+
+    if (args.extra_vars != null) {
+        extra_vars.update(args.extra_vars)
+    }
+
+    vars = [
+      'boxes': boxes,
+      'pipeline': "${args.action}_pipeline.yml",
+      'extraVars': extra_vars
+    ]
+    return vars
+}
--- a/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/nightly-centos7.groovy
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/nightly-centos7.groovy
@@ -1,12 +1,4 @@
 def playBookVars() {
-    playBook = [
-      'boxes': ['pipeline-foreman-server-nightly-centos7', 'pipeline-foreman-smoker-nightly-centos7'],
-      'pipeline': 'install_pipeline.yml',
-      'extraVars': [
-        'pipeline_version': 'nightly',
-        'pipeline_os': 'centos7',
-        'pipeline_type': 'foreman'
-      ]
-    ]
+    playBook = pipelineVars(action: 'install', type: 'foreman', version: 'nightly', os: 'centos7')
     return playBook
 }
--- a/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/release.groovy
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/release.groovy
@@ -1,13 +1,4 @@
 def playBookVars() {
-    playBook = [
-        'boxes': ["pipeline-foreman-server-${params.foreman_version}-${params.distro}"],
-        'pipeline': 'install_pipeline.yml',
-        'extraVars': [
-            'pipeline_version': params.foreman_version,
-            'pipeline_os': params.distro,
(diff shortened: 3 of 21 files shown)


ekohl:
I haven't tested this yet so review carefully.

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
```diff
@@ -0,0 +1,25 @@
...
+        'pipeline_type': type
+      ]
+    ]
+    return playBook
+}
+
+def upgradePipelineVars(version, os, type) {
+    playBook = [
+      'boxes': ["pipeline-upgrade-${type}-${version}-${os}", "pipeline-upgrade-${type}-smoker-${version}-${os}"],
+      'pipeline': 'upgradel_pipeline.yml',
```
evgeni:
```suggestion
      'pipeline': 'upgrade_pipeline.yml',
```

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
```diff
@@ -0,0 +1,25 @@
+def installPipelineVars(version, os, type) {
```
evgeni: those two methods only differ in the used playbook and the box-name prefix. I wonder if we can merge them into a `genericPipeline(action, version, os, type)` and have the others as aliases that set action to `install` or `upgrade`?

or will that become too complex if we ever want to differentiate more inside the methods?

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
```diff
@@ -0,0 +1,25 @@
...
+      'extraVars': [
+        'pipeline_version': version,
+        'pipeline_os': os,
+        'pipeline_type': type
+      ]
+    ]
+    return playBook
+}
+
+def upgradePipelineVars(version, os, type) {
```
evgeni: you use `version, os, type` here, but all callers use `type, version, os`, which seems better to me?

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/release.groovy
```diff
@@ -1,13 +1,5 @@
...
-        'pipeline': 'install_pipeline.yml',
-        'extraVars': [
-            'pipeline_version': params.foreman_version,
-            'pipeline_os': params.distro,
-            'pipeline_type': 'foreman',
-            'foreman_expected_version': params.expected_version
-        ]
-    ]
-    return playBook
+    playbook = installPipelineVars(type: 'foreman', version: params.foreman_version, os: params.distro)
```
evgeni: should `installPipeline` accept an (optional) `extra_vars`?

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/katello/nightly-upgrade.groovy
```diff
@@ -1,4 +1,3 @@
 def playBookVars() {
-    playBook = ['boxes': ['pipeline-upgrade-katello-nightly-centos7'], 'pipeline': 'upgrade_pipeline.yml', 'extraVars': ['pipeline_version': 'nightly', 'pipeline_os': 'centos7', 'pipeline_type': 'katello']]
-    return playBook
+    return upgradePipelineVars(type: 'katello', version: 'nightly', os: 'centos7')
```
evgeni: kinda feels like `os` should default to `centos7` :)

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/pipelines.groovy
```diff
@@ -0,0 +1,25 @@
+def installPipelineVars(version, os, type) {
+    playBook = [
```
ehelms: Nit: I think this could just be `playbook` as thats how the term looks in Ansible vs `play book`

ekohl: I'm pretty sure I copied this from something else. I was actually thinking about naming it vars or something

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/katello/nightly-upgrade.groovy
ekohl: I'd rather be explicit and add centos8 soon as well.

ekohl:
CI failure unrelated. I added another commit that I think addresses the feedback.

evgeni:
feel free to test in prod, the code *reads* correctly :)

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/test/foreman/nightly-centos7.groovy
```diff
@@ -1,12 +1,4 @@
...
-    playBook = [
-      'boxes': ['pipeline-foreman-server-nightly-centos7', 'pipeline-foreman-smoker-nightly-centos7'],
-      'pipeline': 'install_pipeline.yml',
-      'extraVars': [
-        'pipeline_version': 'nightly',
-        'pipeline_os': 'centos7',
-        'pipeline_type': 'foreman'
-      ]
-    ]
+    playBook = pipelineVars(action: 'install', type: 'foreman', version: 'nightly', os: 'centos7')
```
ekohl: If I just had return here with the additional variable, it all failed. Not sure what that's about.

PR #1239, opened 2020-01-16
don't include el8 for 1.22/1.23/1.24

evgeni:
1.20/1.21 is not supported anymore

### DIFF
--- a/koji/foreman-client-mash-split.py
+++ b/koji/foreman-client-mash-split.py
@@ -318,9 +318,8 @@ def main():
     else:
         git_tag = "rpm/{}".format(version)
 
-        if version in ('1.20', '1.21'):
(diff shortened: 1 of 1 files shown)


koji/foreman-client-mash-split.py
```diff
@@ -318,9 +318,8 @@ def main():
     else:
         git_tag = "rpm/{}".format(version)
 
-        if version in ('1.20', '1.21'):
-            del dists['fedora29']
-            dists['fedora27'] = 'fc27'
+        if version in ('1.22', '1.23', '1.24'):
```
evgeni: yes, this explicitly lists all current versions, that way we can just drop that when we only have el8-supported released

PR #1240, opened 2020-01-17
Update the Debian README header

ekohl:
This removes HTML body tags since Apache already adds those. This also
means the title can't be changed. Dropping these just removes invalid
HTML.

It also stops listing the stable version since that's not being updated
anymore. Users shouldn't be using it.

The versions are updated to list the versions that are actually present.

Lastly it adds a short example, including adding the GPG key.

### DIFF
--- a/puppet/modules/freight/files/deb-HEADER.html
+++ b/puppet/modules/freight/files/deb-HEADER.html
@@ -1,7 +1,4 @@
-<title>The Debian Foreman repo</title>
-<head><h1>The Debian Foreman repo</h1></head>
-<body>
-<p>The distros have changed!</p>
+<h1>The Debian Foreman repo</h1>
 
 <p>You will find them at:</p>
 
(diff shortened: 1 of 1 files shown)


puppet/modules/freight/files/deb-HEADER.html
```diff
@@ -1,32 +1,30 @@
-<title>The Debian Foreman repo</title>
-<head><h1>The Debian Foreman repo</h1></head>
-<body>
-<p>The distros have changed!</p>
+<h1>The Debian Foreman repo</h1>
 
 <p>You will find them at:</p>
 
 <pre>deb http://deb.theforeman.org/ &lt;codename&gt; &lt;component&gt;</pre>
```
evgeni: What's a `component` in this case? I mean yeah, that's the Debian speak for main/contrib/non-free, but we use more of a `version` here, right?

ekohl:
Didn't intend to merge this yet but I guess it's now up to the next person to look at component.

PR #1244, opened 2020-01-30
Increase foreman-release-pipeline timeout to four hours

zjhuntin:


ekohl:
I thought usually they passed in like 30 minutes so 2 hours should be sufficient. Why is this needed?

zjhuntin:
We were having issues with really long pipelines when porting over the rails packages. Which resulted in 2 hour + jobs which would cuase it to fail. @ehelms and I thought this might help in those situations.

I may have gone overboard with 2 -> 4 hours, would 3 be more agreeable?

PR #1245, opened 2020-01-31
Use g10k to fetch modules from Puppet forge

mmoll:


ekohl:
What's the reason to prefer g10k over r10k? No strong preference, but it'd be good to know the reasoning.

mmoll:
Mainly performance reasons.

Also, if we use the default Ruby it would 2.0 currently, which is even more problematic.

PR #1250, opened 2020-02-13
Update koji mash scripts for foreman 2.0

zjhuntin:


koji/foreman-rails-mash-split.py
```diff
@@ -1,314 +0,0 @@
-#!/usr/bin/python
```
ekohl: This file should be kept until we drop support for all 1.x versions.

koji/katello-mash-split.py
```diff
@@ -305,9 +305,9 @@ def main():
     
     branch_map = {
         'nightly': 'develop',
+        '3.15': '2.0',
         '3.14': '1.24',
-        '3.13': '1.23',
-        '3.12': '1.22',
+        '3.13': '1.23'
```
ekohl: I'd recommend to keep trailing comma's to keep diffs smaller

koji/foreman-client-mash-split.py
```diff
@@ -318,7 +318,7 @@ def main():
     else:
         git_tag = "rpm/{}".format(version)
 
-        if version in ('1.22', '1.23', '1.24'):
+        if version in ('1.23', '1.24', '2.0'):
```
ekohl: 2.0 does have el8 support. We also haven't dropped the 1.22 pipelines yet so until then I'd keep it in.
```suggestion
        if version in ('1.22', '1.23', '1.24'):
```

koji/katello-mash-split.py
```diff
@@ -305,9 +305,9 @@ def main():
     
     branch_map = {
         'nightly': 'develop',
+        '3.15': '2.0',
         '3.14': '1.24',
-        '3.13': '1.23',
-        '3.12': '1.22',
```
ekohl: While the 3.12 jobs haven't been dropped, I'd keep this in since it doesn't hurt.

PR #1251, opened 2020-02-13
Add Foreman 2.0 & Katello 3.15 to pipelines axes and remove 1.22 & 3.12 from them

zjhuntin:


zjhuntin:
@ekohl does that guard look OK to you?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
```diff
@@ -13,7 +13,9 @@ pipeline {
             agent { label 'sshkey' }
             steps {
                 mash("foreman-mash-split.py", foreman_version)
-                mash("foreman-rails-mash-split.py", foreman_version)
+                if(foreman_version != '2.0') {
```
ehelms: I think this wold be more robust if it checked for `1.23` and `1.24` explicitly. Otherwise, when 2.1 rolls around this will break.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-pipeline.groovy
```diff
@@ -13,7 +13,15 @@ pipeline {
...
             steps {
                 mash("foreman-mash-split.py", foreman_version)
-                mash("foreman-rails-mash-split.py", foreman_version)
+            }
+        }
+        stage('Mash Rails Koji Repositories') {
+            agent { label 'sshkey' }
+            when {
+		expression { foreman_version == '1.23' || foreman_version == '1.24' }
+	    }
```
ehelms: Bracket alignment is off

PR #1255, opened 2020-02-19
Use lowercase for release candidates

ekohl:
Starting Foreman 2.0 the convention changed to lowercase. This because Debian was already using lowercase and in RPMs it's also useful because of sorting: RC < develop < rc. That means with lowercase there's no need to bump the release where there is with uppercase.


mmoll:
merged, bedankt @ekohl!

PR #1256, opened 2020-02-20
Do not run Ruby 2.0.0 tests for Smart Proxy Plugins

ekohl:
The Proxy is now Ruby 2.5+ so these will fail.


mmoll:
merged, bedankt @ekohl!

PR #1258, opened 2020-02-20
drop xenial and stretch from the 2.0 build matrix

evgeni:


mmoll:
merged, спасибо Женя!

PR #1261, opened 2020-02-21
move our shielding from Fastly's IAD to BWI datacenter

evgeni:


evgeni:
@tbrisker for visibility

PR #1264, opened 2020-02-22
Upgrade to nodes to postgresql12

ehelms:
First attempt at something that works across all our various nodes that run Postgres: CentOS and AWS RHEL. The RHEL on AWS uses RHUI to get it's repositories which have different identifiers than standard RHEL repositories. This also handles removal of older postgresql.

I'm sure there are cleaner ways to handle this logic but I wanted something that I tested and worked to start from. @mmoll @ekohl please tell me how to make this cleaner :)

### DIFF
--- a/puppet/modules/slave/manifests/postgresql.pp
+++ b/puppet/modules/slave/manifests/postgresql.pp
@@ -1,5 +1,61 @@
 # @api private
 class slave::postgresql {
+  # only CentOS slaves are used to run unit tests
+  if $facts['os']['family'] == 'RedHat' {
+    if $facts['os']['release']['major'] == '7' {
+
+     ['postgresql-server', 'postgresql-client', 'postgresql'].each |$pkg| {
+        package { "${pkg}-nonscl":
+          ensure  => absent,
+          name    => $pkg,
+          before  => Class['postgresql::globals'],
+        }
+      }
+
+      if $facts['os']['name'] == 'CentOS' {
+        package { 'centos-release-scl-rh':
+          ensure => 'present',
+          before => Class['postgresql::globals'],
+        }
+      } elsif $facts['ec2_metadata'] {
+        yumrepo { 'rhel-server-rhui-rhscl-7-rpms':
+          enabled => true,
+          before => Class['postgresql::globals'],
+        }
+      }
+
+      class { 'postgresql::globals':
+        version              => '12',
+        client_package_name  => 'rh-postgresql12-postgresql-syspaths',
+        server_package_name  => 'rh-postgresql12-postgresql-server-syspaths',
+        contrib_package_name => 'rh-postgresql12-postgresql-contrib-syspaths',
+        service_name         => 'postgresql',
+        datadir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
+        confdir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
+        bindir               => '/usr/bin',
+      }
+    }
+  }
+
+  # Tune DB settings for Jenkins slaves, this is UNSAFE for production!
+  $settings = {
+    'fsync'                        => 'off',
+    'full_page_writes'             => 'off',
+    'synchronous_commit'           => 'off',
+    'autovacuum'                   => 'off',
+    'effective_cache_size'         => '512MB',
+    'shared_buffers'               => '256MB',
+    'checkpoint_completion_target' => '0.9',
+    'wal_level'                    => 'minimal',
+    'max_wal_senders'              => '0',
+  }
+
+  $settings.each |$setting, $value| {
+    postgresql::server::config_entry { $setting:
+      value => $value,
+    }
+  }
+
   Postgresql_psql {
     cwd => '/',
   }
@@ -7,33 +63,15 @@ class slave::postgresql {
   include ::postgresql::client
   include ::postgresql::server
 
+  Class['postgresql::server'] -> Package['postgresql-dev']
+
   # Simple known user/pass that will allow Jenkins to create its
   # own databases when required
   postgresql::server::role { 'foreman':
     password_hash => postgresql_password('foreman', 'foreman'),
     superuser     => true,
     login         => true,
+    require       => Service['postgresql'],
   }
 
-  # only CentOS slaves are used to run unit tests
-  if $::osfamily == 'RedHat' {
-    # Tune DB settings for Jenkins slaves, this is UNSAFE for production!
-    $settings = {
-      'fsync'                        => 'off',
-      'full_page_writes'             => 'off',
-      'synchronous_commit'           => 'off',
-      'autovacuum'                   => 'off',
-      'effective_cache_size'         => '512MB',
-      'shared_buffers'               => '256MB',
-      'checkpoint_segments'          => '20',
-      'checkpoint_completion_target' => '0.9',
-      'wal_level'                    => 'minimal',
-    }
-
-    $settings.each |$setting, $value| {
-      postgresql::server::config_entry { $setting:
-        value => $value,
-      }
-    }
-  }
 }


puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
-  Postgresql_psql {
-    cwd => '/',
-  }
+  # only CentOS slaves are used to run unit tests
+  if $::osfamily == 'RedHat' {
+    if $::operatingsystemmajrelease == '7' {
 
-  include ::postgresql::client
-  include ::postgresql::server
+			exec { 'remove postgresql':
```
ekohl: Mind fixing the tabs in this file?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
+				command => "/bin/yum -y remove postgresql-server postgresql-client postgresql",
```
ekohl: It should be possible to use the `Package` resource for these since the postgresql module now defines the SCL packages, right?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
+        server_package_name  => 'rh-postgresql12-postgresql-server-syspaths',
+        contrib_package_name => 'rh-postgresql12-postgresql-contrib-syspaths',
+        service_name         => 'postgresql',
+        datadir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
+        confdir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
+        bindir               => '/usr/bin',
+      }
+    }
 
-  # only CentOS slaves are used to run unit tests
```
ekohl: These tunings were made RH-only because Debian didn't like `checkpoint_segments` but now that's removed so they can be made generic and moved to the bottom.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
-    superuser     => true,
-    login         => true,
-  }
+      if $::operatingsystem == 'CentOS' {
+      	package { 'centos-release-scl-rh':
+          ensure => 'present',
+          before => Class['postgresql::globals'],
+        }
+      } else {
+        exec { 'enable SCL repo from RHUI':
```
ekohl: Can this be made idempotent? https://github.com/voxpupuli/puppet-rhsm uses `subscription-manager` instead of `yum-config-manager` but maybe that's a better way? They already did the hard work of making sure `subscription-manager` isn't called on every Puppet run.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
 # @api private
 class slave::postgresql {
-  Postgresql_psql {
-    cwd => '/',
-  }
+  # only CentOS slaves are used to run unit tests
+  if $::osfamily == 'RedHat' {
```
ekohl: The modern way of using facts is
```suggestion
  if $facts['osfamily'] == 'RedHat' {
```

When using non-legacy facts it looks like:
```suggestion
  if $facts['os']['family'] == 'RedHat' {
```

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
+    if $::operatingsystemmajrelease == '7' {
```
ekohl:
```suggestion
    if $facts['os']['release']['major'] == '7' {
```

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
+			}
 
-  # Simple known user/pass that will allow Jenkins to create its
-  # own databases when required
-  postgresql::server::role { 'foreman':
-    password_hash => postgresql_password('foreman', 'foreman'),
-    superuser     => true,
-    login         => true,
-  }
+      if $::operatingsystem == 'CentOS' {
```
ekohl:
```suggestion
      if $facts['os']['name'] == 'CentOS' {
```

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -36,4 +52,23 @@
...
+
+  Class['postgresql::server'] -> Package['postgresql-dev']
+
+  # Simple known user/pass that will allow Jenkins to create its
+  # own databases when required
+  postgresql::server::role { 'foreman':
+    password_hash => postgresql_password('foreman', 'foreman'),
+    superuser     => true,
+    login         => true,
+    require       => Service['postgresql'],
```
ekohl: I believe this one is redundant since the role definition already requires this. I know it was there before, but might as well fix it up.

puppet/modules/slave/manifests/postgresql.pp
ehelms: I don't see checkpoint segments anywhere.

puppet/modules/slave/manifests/postgresql.pp
ehelms: I tried that, got duplicate resource definition errors.

puppet/modules/slave/manifests/postgresql.pp
ehelms: I can look at their for an example. These repositories are not controlled by subscription manager hence the need for yum.

puppet/modules/slave/manifests/postgresql.pp
ehelms: Without this, I would always get an error with the role trying to create itself before Postgres was running.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -1,22 +1,38 @@
...
-  Postgresql_psql {
-    cwd => '/',
-  }
+  # only CentOS slaves are used to run unit tests
+  if $facts['os']['family'] == 'RedHat' {
+    if $facts['os']['release']['major'] == '7' {
 
-  include ::postgresql::client
-  include ::postgresql::server
+      exec { 'remove postgresql':
```
ekohl: I think you can work around it:
```puppet
      ['postgresql-server', 'postgresql-client', 'postgresql'].each |$pkg| {
        package { "${pkg}-nonscl":
          ensure  => absent,
          package => $pkg,
          before  => Class['postgresql::globals'],
        }
      }
```

puppet/modules/slave/manifests/postgresql.pp
ekohl: If they're already in a yum repo, you may get away with:
```puppet
yumrepo { 'rhel-server-rhui-rhscl-7-rpms':
  enabled => true,
}
```

ehelms:
I made a round of updates and tested that they worked for me on a RHEL system (non AWS).

puppet/modules/slave/manifests/postgresql.pp
ekohl: It was line 28. It didn't work on Debian since it was already PG >= 10 which dropped `checkpoint_segments`.

mmoll:
let's try it :bomb:

mmoll:
merged, thanks @ehelms!

ekohl:
When deploying we did hit the SELinux issue and it looks like puppetlabs-postgresql doesn't actually manage the config file so it doesn't restore the context on the second run. I've ran a REX job on all (`puppet agent -t ; restorecon -Rv /var/opt/rh/rh-postgresql12 ; puppet agent -t ; true` (last one because REX looks for exit code 0 and Puppet exits with 2)) to fix it up. That only failed on one machine where Puppet is ran in a different way. Fixing that up manually.

PR #1265, opened 2020-02-22
Handle postgresql-devel SCL package

mmoll:


ehelms:
We likely need something like this on the nodes:

https://github.com/theforeman/forklift/blob/master/roles/postgresql_scl/tasks/main.yml#L13

As bundle install will likely fail compiling pg gem as we've seen in devel
environments.

PR #1267, opened 2020-02-23
symlink pg_config

mmoll:


puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -35,11 +35,12 @@
       }
 
       file { '/etc/profile.d/enable_postgresql12_scl.sh':
-        ensure  => file,
-        owner   => 'root',
-        group   => 'root',
-        mode    => '0755',
-        content => 'source scl_source enable rh-postgresql12',
```
mmoll: It seems this doesn't work with Jenkins.

tbrisker:
Thanks @mmoll , let's hope this fixes the tests

PR #1271, opened 2020-02-25
Use vars files in source-release jobs

ekohl:
This also corrects the smart-proxy project_name which fixes its package release job. By making the source and package release use the same vars file, it's much more likely they'll do the same thing.


mmoll:
merged, bedankt @ekohl!

PR #1274, opened 2020-02-26
add previous_commit param to allow fast forward of release rpm builds

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman_packaging_release.groovy
```diff
@@ -34,7 +34,11 @@ pipeline {
 
                 script {
 
-                    if (fileExists('commit')) {
+                    if (params.previous_commit) {
```
evgeni: I am not entirely sure this condition is correct, tho.

ekohl: Maybe it needs to check if it's a non-empty string? I've learned to be defensive about that.

ehelms: I was thinking this could be simpler by dropping the `else` clause that builds all packages. And instead, if no commit file exists take the current commit of HEAD of the packaging repository. The only time this commit file should be lacking is when a new job is created and that would take the commit at the time of the initial run of the job.

ekohl: I like that idea. I was thinking that you'd want to do this on job creation but just taking HEAD is much more elegant. We should make sure to update the procedure to call the job right after branching so we don't miss a commit.

evgeni:
@ehelms @ekohl updated to do what I think you wanted :)

PR #1275, opened 2020-02-27
add pulp to 3.15 mashing

evgeni:


koji/katello-mash-split.py
```diff
@@ -326,7 +326,7 @@ def main():
 
     s.mash_split(whole_path, tmp_path, split_path, mash_config, options=arches, arches=arches, compses=["comps-katello-server-rhel7.xml"], git_tag=git_tag, output_paths=output_paths)
     s.mash_split(whole_path, tmp_path, split_path, mash_config_candlepin, options=options, arches=arches, compses=["comps-katello-candlepin-server-rhel7.xml"], git_tag=git_tag, output_paths=output_paths_candlepin)
-    if version == 'nightly':
+    if version in ['nightly', '3.15']:
```
ekohl: To avoid this in the future, how about: (versions taken from `branch_map`)
```suggestion
    if version in not ['3.12', '3.13', '3.14']:
```

PR #1279, opened 2020-03-03
Update Katello mash script to account for pulp el8

zjhuntin:


ekohl:
Rather than making even more changes, can we merge https://github.com/theforeman/foreman-infra/pull/1189 first and iterate on that? I feel it's much more prepared for adding additional configs.

zjhuntin:
@ekohl agreed

PR #1281, opened 2020-03-08
Remove old branch plugin tests

mmoll:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
```diff
@@ -4,9 +4,5 @@
     branch:
       - develop:
           foreman_branch: develop
-      - 15.0-stable:
-          foreman_branch: 1.22-stable
-      - 15.1-stable:
-          foreman_branch: 1.23-stable
```
mmoll: @lzap Should new versions get added here?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman-tasks.yaml
```diff
@@ -4,9 +4,5 @@
     branch:
       - master:
           foreman_branch: develop
-      - 0.16.x:
-          foreman_branch: 1.23-stable
-      - 0.15.x:
-          foreman_branch: 1.22-stable
```
mmoll: @adamruzicka Should new versions get added here?

ekohl:
Until 2.0 goes GA, 1.23 is supported. Should we keep those versions around for a bit longer?

mmoll:
I'm unsure if anybody cares about the results, tbh :)

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman-tasks.yaml
adamruzicka: That would make sense to me

PR #1285, opened 2020-03-10
Add postgresql-evr package and its repo to unit test slaves

ianballou:
PR from the discussion here: https://community.theforeman.org/t/adding-postgresql-evr-to-jenkins-slaves/17686/6

A couple notes:

The postgresql-evr package isn't available in nightly yet so I put the staging repo in its place.

`init.pp` was mentioned for configuring the repo but it seemed like just using `yumrepo` in `postgresql.pp` would work based on other examples.

### DIFF
--- a/puppet/modules/slave/manifests/postgresql.pp
+++ b/puppet/modules/slave/manifests/postgresql.pp
@@ -16,6 +16,15 @@ class slave::postgresql {
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl => 'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+        } ->
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          notify  => Class['postgresql::server::service'],
+          require => Class['postgresql::server::install'],
+        }
       } elsif $facts['ec2_metadata'] {
         yumrepo { 'rhel-server-rhui-rhscl-7-rpms':
           enabled => true,


ianballou:
I think having a [repo for foreman-infra](https://community.theforeman.org/t/adding-postgresql-evr-to-jenkins-slaves/17686/7?u=iballou) could be a good alternate strategy.  I'm not sure how much work that entails though.  If that's something I could help set up I'd be glad to.

And yeah I thought using the Koji repo wasn't great.  We're a bit stuck at the moment because nightly isn't publishing.  Maybe this hypothetical infra repo could get the postgresql-evr package on the test machines faster?

jlsherrill:
@ekohl do you see this being controlled by a koji tag and living at: http://yum.theforeman.org/infra/ ?

ekohl:
I wasn't thinking so complex and rather just dropping the file + call createrepo. Given its size and how little we expect it to change I thought that might be sufficient.

jlsherrill:
I'm not sure who has access to do that on yum.theforeman.org? I sure don't.

ehelms:
If we are looking for dead simple solutions, did we consider just installing this link?

http://koji.katello.org/kojifiles/packages/rh-postgresql12-postgresql-evr/0.0.1/2.el7/x86_64/rh-postgresql12-postgresql-evr-0.0.1-2.el7.x86_64.rpm

ianballou:
> If we are looking for dead simple solutions, did we consider just installing this link?
> 
> http://koji.katello.org/kojifiles/packages/rh-postgresql12-postgresql-evr/0.0.1/2.el7/x86_64/rh-postgresql12-postgresql-evr-0.0.1-2.el7.x86_64.rpm

I'm okay with this.  The only dependencies of the RPM are `rh-postgresql12-postgresql-server` and `rh-postgresql12-runtime`  Those should both be on the machine already I believe.

ekohl:
I wonder if this reaches out to koji every puppet run. I haven't really seen an argument against an infra repo on yum.theforeman.org.

jlsherrill:
The downsides i see:
* its not infra as code like everything else we're managing is
* fewer people can update it

ekohl:
> its not infra as code like everything else we're managing is

It's not really different than the general pattern: ensure yum repos are present and then that a package is installed.

> fewer people can update it

We also rely on packages from CentOS and we can't update those either. Also all people who can merge to foreman-infra should have access to yum.theforeman.org. If an update is broken and needs to be reverted, you probably also need infra access. I don't really see it as an issue.

ehelms:
I haven't seen the argument against using our own existing nightly
repository either. In this way, our code is being tested against that which
we deploy. A one off, unmanaged infra repo provides not the same guarantee
and requires a human to go touch this each time.

I vote if we are going this one off infra route, we do it right.

jlsherrill:
> It's not really different than the general pattern: ensure yum repos are present and then that a package is installed.
the configuration here isn't, but the contents of the repo are not.

We're now 6 days into this discussion, i say we either:

1) point to the pulpcore repo
2) create a simple yum repo on yum.theforeman.org

I get that creating a simple yum repo is 'as easy as making a directory, adding rpms, and running create repo', but we're now 6 days in and seem to be going nowhere.  So i would say @ekohl please take the 5 minutes to make the repository, or lets just go with 1)

pcreech:
So, to weigh in, I think it will be beneficial long term to have our own infrastructure repo that we utilize cradle to grave from koji -> published on yum.theforeman.org

My suggestion would be we create the repo manually (for now) on yum.theforeman.org, and we can start building out the initiative to create the build tags, mashing, jobs, publish tasks, etc... to make this a first class internal (to foreman/katello) repo.  i.e. not necessarily for community user consumption

ianballou:
It looks like we have a couple votes on creating a small repo on yum.theforeman.org at least in the short term.  How does that sound? Can we get that created?

ekohl:
I'd like to hold this off until we actually get green nightlies. We keep adding more changes while it's red. That's why I haven't been giving this priority.

ianballou:
With nightly passing again, are we okay to continue on here?

ianballou:
@ekohl @pcreech @ehelms we're still blocked on getting the EVR extension installed on the test machines.  Can the infra repo on yum.theforeman.org be created?

The RPMs we'd want are these:

https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/rh-postgresql12-postgresql-evr-0.0.1-2.el7.x86_64.rpm

and

https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el7/x86_64/rh-postgresql12-postgresql-debversion-1.1.1-1.el7.x86_64.rpm

@jlsherrill let me know if we don't want the debversion RPM in there as well.

ehelms:
Here ya go: https://yum.theforeman.org/infra/el7/

ianballou:
@ehelms thanks!

I've updated the PR.  Does the ordering look okay? I figured the yum repo and RPM should be added after `Class['postgresql::globals']` since I need PostgreSQL to be installed first.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,15 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl =>  'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+          require => Class['postgresql::globals'],
+        }
```
ehelms: This repository can be setup at any point.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,15 @@
...
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          require =>  yumrepo['foreman-infra-el7'],
+        }
```
ehelms: The requires there is good, but I think `postgresql::globals` won't ensure the package is installed. I am thinking this needs to happen after https://github.com/theforeman/foreman-infra/pull/1285/files#diff-8fa48bcc6aed22cd7a1aa8b436d72823R87

puppet/modules/slave/manifests/postgresql.pp
ianballou: Do you mean that the infra repo doesn't need a `require`? Or that it doesn't need to be in the `if $facts['os']['name'] == 'CentOS'`?

puppet/modules/slave/manifests/postgresql.pp
ianballou: That makes sense.  And you're referring to the package requiring `postgresql::server::role`?

puppet/modules/slave/manifests/postgresql.pp
ehelms: The repo doesn't need a require. As in, it can technically be laid down at any point. The package installation is what matters order of operations wise.

puppet/modules/slave/manifests/postgresql.pp
ehelms: Actually, you can probably just do:

```
requires => Package['postgresql::globals::server_package_name']
```

puppet/modules/slave/manifests/postgresql.pp
ianballou: There won't be a case where puppet tries to install the evr package before the infra repo is enabled?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,14 @@
...
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl =>  'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+        }
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          requires => Package['postgresql::globals::server_package_name'],
+        }
```
ehelms: @ekohl does this make sense to you?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,14 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl =>  'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+        }
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          requires => Package['postgresql::globals::server_package_name'],
```
ehelms, line 25: s/requires/require/g

puppet/modules/slave/manifests/postgresql.pp
ehelms: Oh yes, there could be, you can add a `} ->` here.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,14 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl =>  'https://yum.theforeman.org/infra/el7/',
```
ekohl, line 20:
```suggestion
          baseurl => 'https://yum.theforeman.org/infra/el7/',
```

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,14 @@
...
+          enabled => true,
+        }
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          require => Package['postgresql::globals::server_package_name'],
```
ekohl, line 25: I don't think this literal package name exists. You can also use the wrapping class instead:
```suggestion
          require => Class['postgresql::server::install'],
```

 Also know that right now it's not guaranteed to be installed before the service starts. You may want to notify `postgresql::server::service` as well.

ianballou: I don't think restarting postgresql is required after installing an extension, but I can add it anyway to be safe

ianballou:
Think I've addressed the issues.

puppet/modules/slave/manifests/postgresql.pp
ekohl: I also don't know for sure if you need to, but as you say: it doesn't hurt.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,15 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl => 'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+        } ->
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          notify  =>  Service['postgresql::server::service'],
```
ekohl, line 25:
```suggestion
          notify  => Service['postgresql::server::service'],
```

ianballou: oops, thanks.

ianballou:
Oh looks like I'm not restarting PostgreSQL correctly: `Could not find resource 'Service[postgresql::server::service]' in parameter 'notify'`

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,15 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
+          baseurl => 'https://yum.theforeman.org/infra/el7/',
+          enabled => true,
+        } ->
+        package { 'rh-postgresql12-postgresql-evr':
+          ensure  => 'present',
+          notify  => Service['postgresql::server::service'],
```
ekohl, line 25:
```suggestion
          notify  => class['postgresql::server::service'],
```

ianballou: Great, looks like that fixed it.

ianballou:
Looks like the build passed, not sure why Travis is still pending.

ehelms:
It's been doing this from time to time lately.

ehelms:
@ianballou Merging, please try to keep an eye out on PR tests across Foreman and Katello for any weirdness. This should land in 30 minutes or less across the nodes.

ianballou:
Thanks @ehelms, I'll keep an eye out.

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -16,6 +16,15 @@
           ensure => 'present',
           before => Class['postgresql::globals'],
         }
+        yumrepo { 'foreman-infra-el7':
```
ekohl, line 19: Looks like if you don't add a name, yum becomes verbose and now we get hourly cron mails with:
```
/etc/cron.hourly/0yum-hourly.cron:

Repository 'foreman-infra-el7' is missing name in configuration, using id
```

Mind adding one?

ianballou: Sure I'll make another PR

ianballou: https://github.com/theforeman/foreman-infra/pull/1305

PR #1286, opened 2020-03-11
pass playbook extra vars also to collect_debug playbook

evgeni:


evgeni:
needed for https://github.com/theforeman/forklift/pull/1125

PR #1289, opened 2020-03-12
Fix pulpcore mash config

ekohl:


koji/collection-mash-split.py
```diff
@@ -445,7 +447,9 @@ def main():
         ]
 
         if version not in ['3.12', '3.13', '3.14']:
-            pulpcore = MashConfig(collection, version, "el7", "pulpcore-el7", "pulpcore/el7")
+            pulpcore = MashConfig(collection, version, "pulpcore-el7", "pulpcore-el7", "pulpcore/el7")
+            # The default builds katello-{}-pulpcore-el7
+            pulpcore.name = 'katello-pulpcore-{}-el7'.format(version)
```
ekohl: I was wondering about introducing a `tag_suffix` parameter instead

pcreech: Taking it to a logical conclusion... It might be better to utilize something along the lines of this pattern:
```
tag_name = "katello-pulpcore"
dist = "el7"
"{name}-{version}-{dist}".format(name=tag_name, version=version, dist=dist)
```

Perhaps in a separate function or something.

pcreech:
All in all, LGTM.  The comment above would be a good candidate for a future refactor, depending on expected life of the script.

PR #1290, opened 2020-03-12
Add el8 nightly Foreman and Pulpcore mash options

zjhuntin:


koji/collection-mash-split.py
```diff
@@ -393,6 +393,10 @@ def main():
             extras = []
 
         mashes = [MashConfig(collection, version, "rhel7-dist", "rhel7", "RHEL/7", extras)]
+
+        if collection is 'foreman' and version is 'nightly':
```
ekohl: You shouldn't compare strings with `is`. I'd also make it future proof. This ensures that after branching, you don't need to go back and add 2.1 (same for pulpcore):
```suggestion
        if collection == 'foreman' and version not in ['1.22', '1.23', '1.24', '2.0']:
```

zjhuntin: Done!

PR #1291, opened 2020-03-16
Convert kafo_parsers PR testing to a pipeline

ehelms:


ehelms:
The `kafo.groovy` script already calls that. I posted some further updates that were needed to create isolated bundler environments for the matrix. Here is an example PR passing against this code:

https://github.com/theforeman/kafo_parsers/pull/37

PR #1295, opened 2020-03-19
remove augeas devel package

mmoll:


mmoll:
Augeas was used at least in the smart-proxy in the past, but AFAICY, nothing in our stack uses it outside Puppet. RHEL/CentOS 8 doesn't bring this package anymore and starting today RHEL 7 has problems with it, too.

ekohl:
I this today was just a sync issue. Looks like now they have matching versions again. I wonder if kafo still relies on this because Puppet might.

mmoll:
> I wonder if kafo still relies on this because Puppet might.

According to my tests, it doesn't. :)

PR #1296, opened 2020-03-19
Current state of process running to update koji's flattened el8 repo

pcreech:
Note:  grobisplitter role is taken almost verbatim from fedora-ansible folder.  Small modifications have been made to fit our environment

### DIFF
--- a/koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
+++ b/koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+HOMEDIR=/root/koji-sync/podman/
+BINDIR=/usr/local/bin
+
+ARCHES="x86_64"
+DATE=$(date -Ih | sed 's/+.*//')
+
+DATEDIR=${HOMEDIR}/koji/${DATE}
+
+if [ -d ${DATEDIR} ]; then
+    echo "Directory already exists. Please remove or fix"
+    exit
+else
+mkdir -p ${DATEDIR}
+fi
+
+for ARCH in ${ARCHES}; do
+    # The archdir is where we daily download updates for rhel8
+    ARCHDIR=${HOMEDIR}/repos/
+    if [ ! -d ${ARCHDIR} ]; then
+	echo "Unable to find ${ARCHDIR}"
+	exit
+    fi
+
+    # We consolidate all of the default repositories and remerge them
+    # in a daily tree. This allows us to point koji at a particular
+    # day if we have specific build concerns.
+    OUTDIR=${DATEDIR}/${ARCH}
+    mkdir -p ${OUTDIR}
+    if [ ! -d ${OUTDIR} ]; then
+	echo "Unable to find ${ARCHDIR}"
+	exit
+    else
+	cd ${OUTDIR}
+    fi
+
+    # Begin splitting the various packages into their subtrees
+    ${BINDIR}/splitter.py --action hardlink --target RHEL-8-001 ${ARCHDIR}/BaseOS/ &> /dev/null
+    if [ $? -ne 0 ]; then
+	echo "splitter ${ARCH} baseos failed"
+	exit
+    fi
+    ${BINDIR}/splitter.py --action hardlink --target RHEL-8-002 ${ARCHDIR}/AppStream/ &> /dev/null
+    if [ $? -ne 0 ]; then
+	echo "splitter ${ARCH} appstream failed"
+	exit
+    fi
+    ${BINDIR}/splitter.py --action hardlink --target RHEL-8-003 ${ARCHDIR}/PowerTools/ &> /dev/null
+    if [ $? -ne 0 ]; then
+	echo "splitter ${ARCH} codeready failed"
+	exit
+    fi
+
+    # Copy the various module trees into RHEL-8-001 where we want them
+    # to work.
+    echo "Moving data to ${ARCH}/RHEL-8-001"
+    cp -anlr RHEL-8-002/* RHEL-8-001
+    cp -anlr RHEL-8-003/* RHEL-8-001
+    # Go into the main tree
+    pushd RHEL-8-001
+    rm -rf ruby\:2.6*
+    # Mergerepo didn't work so lets just createrepo in the top directory.
+    createrepo_c .  &> /dev/null
+    popd
+
+    # Cleanup the trash 
+    rm -rf RHEL-8-002 RHEL-8-003
+#loop to the next
+done
+
+## Set up the builds so they are pointing to the last working version
+cd ${HOMEDIR}/koji/
+if [[ -e staged ]]; then
+    if [[ -h staged ]]; then
+	rm -f staged
+    else
+	echo "Unable to remove staged. it is not a symbolic link"
+	exit
+    fi
+else
+    echo "No staged link found"
+fi
+
+echo "Linking ${DATE} to staged"
+ln -s ${DATE} staged
+
+
+# for ARCH in ${ARCHES}; do
+#     pushd latest/
+#     mkdir -p ${ARCH}
+#     dnf --disablerepo=\* --enablerepo=RHEL-8-001 --repofrompath=RHEL-8-001,https://infrastructure.fedoraproject.org/repo/rhel/rhel8/koji/staged/${ARCH}/RHEL-8-001/ reposync -a ${ARCH} -a noarch -p ${ARCH} --newest --delete  &> /dev/null
+#     if [[ $? -eq 0 ]]; then
+# 	cd ${ARCH}/RHEL-8-001
+# 	createrepo_c .  &> /dev/null
+#     else
+# 	echo "Unable to run createrepo on latest/${ARCH}"
+#     fi
+#     popd
+# done
+
--- a/koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
+++ b/koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
@@ -1,0 +1,397 @@
+#!/bin/python3
+
+# Import libraries needed for application to work
+
+import argparse
+import shutil
+import gi
+import gzip
+import librepo
+import hawkey
+import tempfile
+import os
+import subprocess
+import sys
+
+# Look for a specific version of modulemd. The 1.x series does not
+# have the tools we need.
+try:
+    gi.require_version('Modulemd', '2.0')
+    from gi.repository import Modulemd
+except:
+    print("We require newer vesions of modulemd than installed..")
+    sys.exit(0)
+    
+mmd = Modulemd
+
+# This code is from Stephen Gallagher to make my other caveman code
+# less icky.
+def _get_latest_streams (mymod, stream):
+    """
+    Routine takes modulemd object and a stream name.
+    Finds the lates stream from that and returns that as a stream
+    object. 
+    """
+    all_streams = mymod.search_streams(stream, 0)
+    latest_streams = mymod.search_streams(stream,
+                                          all_streams[0].props.version) 
+    
+    return latest_streams
+    
+def _get_repoinfo(directory):
+    """
+    A function which goes into the given directory and sets up the
+    needed data for the repository using librepo.
+    Returns the LRR_YUM_REPO
+    """
+    with tempfile.TemporaryDirectory(prefix='elsplit_librepo_') as lrodir:
+        h = librepo.Handle()
+        h.setopt(librepo.LRO_URLS, ["file://%s" % directory])
+        h.setopt(librepo.LRO_REPOTYPE, librepo.LR_YUMREPO)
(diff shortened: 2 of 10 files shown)


koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
```diff
@@ -0,0 +1,100 @@
...
+    # to work.
+    echo "Moving data to ${ARCH}/RHEL-8-001"
+    cp -anlr RHEL-8-002/* RHEL-8-001
+    cp -anlr RHEL-8-003/* RHEL-8-001
+    # Go into the main tree
+    pushd RHEL-8-001
+    rm -rf ruby\:2.6*
+    # Mergerepo didn't work so lets just createrepo in the top directory.
+    createrepo_c .  &> /dev/null
+    popd
```
ekohl: `pushd` can fail in unexpected ways (previous directory no longer exists). Doesn't this work just as well?
```suggestion
    rm -rf RHEL-8-001/ruby\:2.6*
    # Mergerepo didn't work so lets just createrepo in the top directory.
    createrepo_c RHEL-8-001 &> /dev/null
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
```diff
@@ -0,0 +1,100 @@
...
+    mkdir -p ${OUTDIR}
+    if [ ! -d ${OUTDIR} ]; then
+	echo "Unable to find ${ARCHDIR}"
+	exit
+    else
+	cd ${OUTDIR}
+    fi
+
+    # Begin splitting the various packages into their subtrees
+    ${BINDIR}/splitter.py --action hardlink --target RHEL-8-001 ${ARCHDIR}/BaseOS/ &> /dev/null
```
ekohl: Since it's not really RHEL, should this use `EL` or `CentOS`?

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
```diff
@@ -0,0 +1,100 @@
...
+ARCHES="x86_64"
+DATE=$(date -Ih | sed 's/+.*//')
+
+DATEDIR=${HOMEDIR}/koji/${DATE}
+
+if [ -d ${DATEDIR} ]; then
+    echo "Directory already exists. Please remove or fix"
+    exit
+else
+mkdir -p ${DATEDIR}
```
ekohl: Nit: indenting

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
ekohl: I'm always a fan of `-e`
```suggestion
#!/bin/bash -e
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
```diff
@@ -0,0 +1,397 @@
+#!/bin/python3
```
ekohl: Not all distros symlink `/bin` to `/usr/bin`
```suggestion
#!/usr/bin/python3
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
```diff
@@ -0,0 +1,397 @@
...
+    Returns a set of file locations.
+    """
+    sack = package_sack
+    pkgs = set()
+
+    for pkg in hawkey.Query(sack):
+        if pkg.location in modpkgset:
+            continue
+        pkgs.add(pkg.location)
+    return pkgs
```
ekohl:
```suggestion
    return {pkg.location for pkg in hawkey.Query(sack) if pkg.location in modpkgset}
```

Or if you modify `_get_modular_pkgset` to return a set:
```suggestion
    return {pkg.location for pkg in hawkey.Query(sack)} & modpkgset
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
```diff
@@ -0,0 +1,397 @@
...
+    idx.upgrade_streams(2)
+    for modname in idx.get_module_names():
+        mod = idx.get_module(modname)
+        for stream in mod.get_all_streams():
+            templ = list()
+            for pkg in stream.get_rpm_artifacts():
+                if pkg in pkgs_list:
+                    templ.append(pkgs_list[pkg])
+                else:
+                    continue
```
ekohl:
```suggestion
            templ = [pkgs_list[pkg] for pkg in stream.get_rpm_artifacts() if pkg in pkgs_list]
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
```diff
@@ -0,0 +1,397 @@
...
+    packages are inside it. 
+    Returns a list of packages
+    """
+    pkgs = set()
+
+    for modcts in mod.values():
+        for pkg in modcts:
+            pkgs.add(pkg)
+
+    return list(pkgs)
```
ekohl: Why should this be a list? Later on you only use it to check if items are in it and that's much faster with a set. You can even do `set() & set()` then instead of doing it in the loop.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
```diff
@@ -0,0 +1,397 @@
...
+            ourname = stream.get_NSVCA()
+            tmp_name = "%s:%s" % (stream.props.module_name,
+                                 stream.props.stream_name)
+            # Get dependencies is a list of items. All of the modules
+            # seem to only have 1 item in them, but we should loop
+            # over the list anyway.
+            for deps in stream.get_dependencies():
+                isprovided = True # a variable to say this can be added.
+                for mod in deps.get_runtime_modules():
+                    tempstr=""
```
ekohl: I think this can be dropped
```suggestion
```

koji/flatten-el8/koji-ansible/roles/grobisplitter/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: install python packages
+  package: name={{ item }} state=present
+  with_items:
```
ekohl: IIRC you can pass a list to `name` and have a single yum/dnf transaction which is faster.

koji/flatten-el8/script.sh
```diff
@@ -0,0 +1,10 @@
+dnf install -y epel-release
+dnf install -y ansible "dnf-command(reposync)"
+
+pushd /root/koji-sync/podman/repos
+dnf reposync --repoid AppStream --repoid PowerTools --repoid BaseOS --download-metadata
+
+pushd /root/koji-sync/podman/koji-ansible/
+ansible-playbook -i inventory/inventory master.yaml --connection=local
```
ekohl: If you're doing to use podman, wouldn't it make sense to write a `Dockerfile` to install all the tools so you can cache the image?

koji/flatten-el8/koji-ansible/roles/grobisplitter/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+
+- name: copy local/bin files
+  copy: src={{item}} dest=/usr/local/bin/ mode=0755
+  with_items: 
+    - splitter.py
+    - rhel8-split.sh
+  tags:
+    - grobi
+
+- name: daily cron job to split and merge repos
```
ekohl: It's odd to see cron items in a container.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
pcreech: Unkown.  Did not write the script, and I'm inclined to leave it as is, at least for now.  II would also say, it's highly unlikely this directory will just "dissappear" in the time it's used.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
pcreech: *shrug*  My assumption is, they had access in fedora to the RHEL repo internally to their infra, so they named it this way.  I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
pcreech: I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
pcreech: . I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
pcreech: . I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
pcreech: No idea, it's what they wrote for epel. . I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
pcreech: I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/script.sh
pcreech: Yes! But that would require me to learn how to write a Dockerfile.  In the very short timeframe I had, i worked with what resources I knew to get it knocked out.

As I get time, i'll swing back to this and add one.

koji/flatten-el8/koji-ansible/roles/grobisplitter/tasks/main.yml
pcreech: This is the role as taken from fedora-ansible.  It can probably be removed.  It's existence or removal is inconsequential in the grand scheme of things.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
pcreech: I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/splitter.py
pcreech: I'd prefer not to deviate from their script more than necessary at this time.

koji/flatten-el8/script.sh
ekohl: It'd roughly be https://github.com/theforeman/foreman-infra/commit/105043aac95155f06733488a1d3fed2e8351f108

pcreech: That leaves two options in my mind.  a) merge current state "as is" then update it, then improve upon it, or b) Go ahead and do the revisions now for this pr.

I'm in favor of a, primarily because it captures what we are doing at this exact moment, in case of catastrophic failure, and then we can go back and improve upon it iteratively.

ekohl: Note that my commit is on top of this commit.

ehelms:
This appears to be between two routes:

 * Merge as is since its being used and fix it up over time
 * Hold this here, to be fixed up when time allows, but continue to use the way it exists today operationally

@evgeni @ekohl @zjhuntin can you please weigh in on your choice of option

evgeni:
Given it's already used, let's document the status quo and fix it later.

koji/flatten-el8/koji-ansible/roles/grobisplitter/files/rhel8-split.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+else
+    echo "No staged link found"
+fi
+
+echo "Linking ${DATE} to staged"
+ln -s ${DATE} staged
+
+
+# for ARCH in ${ARCHES}; do
```
zjhuntin: Any reason to keep this?

pcreech: For now, in-script documentation of how the repo merging worked for them, in case we need inspiration.  At least in my opinion.

PR #1297, opened 2020-03-19
use different timeout type for test_proxy_develop_pr_rubocop

mmoll:


mmoll:
Message in Jenkins: `MissingFieldException: No field 'timeoutMinutes' found in class 'hudson.plugins.build_timeout.impl.LikelyStuckTimeOutStrategy'`

PR #1303, opened 2020-03-31
Add dnf to nodes to enable use of dnf repoclosure for EL8 builds

ehelms:


mmoll:
merged, thanks @ehelms!

PR #1309, opened 2020-04-02
Fix some puppet-lint warnings

mmoll:


puppet/modules/motd/manifests/params.pp
```diff
@@ -1,6 +1,6 @@
 class motd::params {
   case $::osfamily {
-    redhat, debian, suse: {
+    'RedHat', 'Debian', 'Suse': {
```
ekohl, line 3: I took this a step further in https://github.com/theforeman/foreman-infra/pull/1310

PR #1310, opened 2020-04-02
Update motd module to use modern standards

ekohl:
Removes the fairly redundant params.pp file and uses puppet-strings documentation as well as data types.


mmoll:
merged, bedankt @ekohl!

PR #1311, opened 2020-04-02
Cleanup ssh module

mmoll:


puppet/modules/ssh/manifests/init.pp
```diff
@@ -1,15 +1,4 @@
 class ssh {
-  file { '/root/.ssh':
-    ensure => directory,
-    owner  => 'root',
-    group  => 'root',
-  }
-
-  file { '/root/.ssh/authorized_keys':
-    ensure  => file,
-    content => $::root_ssh_recovery_key,
```
mmoll, line 10: This was not set for at least some years.

puppet/modules/ssh/manifests/init.pp
```diff
@@ -1,15 +1,4 @@
 class ssh {
-  file { '/root/.ssh':
-    ensure => directory,
-    owner  => 'root',
-    group  => 'root',
-  }
-
-  file { '/root/.ssh/authorized_keys':
-    ensure  => file,
```
ekohl, line 9: Should we ensure it's absent?

mmoll: Shouldn't make a difference with `PermitRootLogin no` ;)

mmoll: changed anyway

PR #1312, opened 2020-04-02
Add a Jenkins node profile

ekohl:
The goal of this class is to replace the config group currently used in Foreman. Long term all machines should only be assigned classes from the profiles module and Foreman only imports those classes.

This also removes the hub and copr configs because I believe they are unused.

### DIFF
--- a/puppet/modules/profiles/manifests/jenkins/node.pp
+++ b/puppet/modules/profiles/manifests/jenkins/node.pp
@@ -1,0 +1,27 @@
+# @summary A Jenkins node
+#
+# @param koji_certificate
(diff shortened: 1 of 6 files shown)


ekohl:
A better alternative is to manage the koji certificate in Jenkins natively using secrets and only place it just before it's being used. That requires some refactoring that I don't have time for now.

ehelms:
:+1: Seems like a good change if I understand things correctly. This implies over time the `slave` module would never be directly assigned to a host?

ekohl:
Correct - my idea is that for every type of server we have (essentially all host groups) we'd have a `profiles::$group` class + `profiles::base`. In Foreman we'd change the class import filter to only import `profiles::*`.

PR #1313, opened 2020-04-03
Rewrite the webserver profile

ekohl:
This is a rather large PR with various commits to first clean up the environment. It has been tested with vagrant and that correctly spins up a machine.

This is a preparation for reinstalling the webserver on OSOUSL. Installing web03.osousl.theforeman.org and rsyncing content will be cleaner than a full rsync. Doing so allows us to move over a single vhost at a time, starting with stagingdeb.

It also prepares for a regular user for the yum instead of sending over files as root by splitting it to a separate class.


ekohl:
To add to this: when merging it I'd prefer to put the machine in noop and then deploy it to see if there's anything I missed.

mmoll:
Test fails, but I guess we can exlcude Debian/Ubuntu there.

ekohl:
Actually, I took the time to make it compile on a current version of puppetlabs-apt since that was trivial to do. Also moved things around a bit so the profile now includes the vhosts.

I'd like to move to a certificate per vhost rather than one shared certificate but I'll leave that till after the deployment.

ekohl:
Updated, I think it's ready for review now.

mmoll:
tests fail :broken_heart:

ehelms:
Approach generally looks good. If we create the new machines I'd love to
see the naming be more reflective of what it is. E.g. just
web.osuosl.theforeman.org if it's one machine or yum.osuosl or maybe
yum-web if we end up splitting per vhost.

ekohl:
The current name is web02.rackspace.theforeman.org because this allows us to replace it. web01 was centos 6 based, web02 was a reinstall. We did the same thing with master01 -> master02. Then just change the CNAME to point to the correct instance to migrate. Never reuse hostnames is a practice I've learned and still really like.

ekohl:
Added the selinux_core test fixture.

evgeni:
@ekohl do you want to handle the deployment, or shall someone else?

ekohl:
I was going to yesterday, but then theforeman.org transfered and everything broke. Will have a look.

PR #1318, opened 2020-04-06
also use homedir when removing the cert

evgeni:


evgeni:
@ekohl or did you intentionally not refactor that?

PR #1320, opened 2020-04-06
allow to configure the user of the yum repo

evgeni:


evgeni:
draft as I didn't test it yet at all

evgeni:
done, also now the user is created by secure_ssh::receiver, still need to test if capistrano will be able to deploy

puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -27,11 +28,26 @@
     },
   ]
 
+  secure_ssh::receiver_setup { 'yum':
```
ekohl, line 31: In other vhosts we've had to create a boolean for this since the type always creates the keys in `/etc/puppet` which CI doesn't have permission to. Ideally we should use Puppet's settings for the state dir (which are dynamic and can use the homedir), but that's also a migration.

puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -27,11 +28,26 @@
...
+    user           => $user,
+    foreman_search => 'host ~ node*.jenkins.osuosl.theforeman.org and (name = external_ip4 or name = external_ip6)',
+    script_content => file('web/yum.sh'),
+  }
+
+  file { "${yum_directory}":
```
ekohl, line 37:
```suggestion
  file { $yum_directory:
```

ekohl:
Should we close this in favor of https://github.com/theforeman/foreman-infra/pull/1333?

evgeni:
yes

PR #1321, opened 2020-04-06
use 'mysql::bindings' to install mysql-devel package

mmoll:


puppet/modules/slave/manifests/init.pp
```diff
@@ -75,15 +75,6 @@
...
-    'mysql-dev':
-      ensure => present,
-      name   => $::osfamily ? {
-        'RedHat' => $::operatingsystemmajrelease ? {
-          '6'     => 'mysql-devel',
-          default => 'mariadb-devel',
-        },
-        default  => 'libmysqlclient-dev'
-      };
     'postgresql-dev':
```
ekohl, line 78: There may be an equivalent here but let's consider that out of scope now :)

PR #1330, opened 2020-04-07
Drop Puppet 4.10 from Kafo testing

ehelms:


ehelms:
We moved the installer into the SCL so I believe we can drop this testing matrix.

PR #1332, opened 2020-04-08
drop puppet and web from capistrano config

evgeni:
we don't use that anymore


ekohl:
Actually, not sure which key it exactly uses. Let's leave that to https://github.com/theforeman/foreman-infra/pull/1320 to clean up.

PR #1333, opened 2020-04-08
Introduce a dedicated user for yumrepo deployments

ekohl:
This introduces a user yumrepo who owns the yum.theforeman.org vhost. A script is put in place to sync content which replaces the previous capistrano setup.


puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -27,11 +29,22 @@
     },
   ]
 
+  if $setup_receiver {
+    secure_ssh::receiver_setup { $user:
+      user           => $user,
+      foreman_search => 'host ~ node*.jenkins.osuosl.theforeman.org and (name = external_ip4 or name = external_ip6)',
+      allowed_ips    => ['192.168.122.1'],
```
evgeni, line 36: probably not ;)

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
...
+		latest=$(ls -t foreman-client-release-[0-9]*.rpm 2>/dev/null | head -n1)
+		if [[ -n "$latest" ]] ; then
+			ln -sf "$latest" foreman-client-release.rpm
+		fi
+
+		if [[ $MERGE == true ]] ; then
+			createrepo --skip-symlinks --update .
+		fi
+	)
+done
```
evgeni, line 48: those two look weirdly indented

puppet/modules/web/manifests/vhost/yum.pp
ekohl: Technically this will be unused in production when Foreman API credentials are present, but you're right; that was my test code :)

puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -27,11 +29,21 @@
     },
   ]
 
+  if $setup_receiver {
+    secure_ssh::receiver_setup { $user:
+      user           => $user,
+      foreman_search => 'host ~ node*.jenkins.osuosl.theforeman.org and (name = external_ip4 or name = external_ip6)',
```
ekohl, line 35: Btw, I think we should pull this up to the profile level and unify with the web user since that's the same setup. I'll take a stab at looking how we reuse those keys and credentials but IMHO that doesn't have to block deployment.

puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -27,11 +29,21 @@
...
+    }
+  }
+
   web::vhost { 'yum':
-    servername   => $servername,
-    docroot      => $yum_directory,
-    docroot_mode => '2575',
-    directories  => $yum_directory_config,
+    servername    => $servername,
+    docroot       => $yum_directory,
```
evgeni, line 42: don't we need to create that? or will web::host (or apache::vhost) handle that for us?

ekohl: apache::vhost will, which is why we pass the owner, group and mode.

evgeni: cool.

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
...
+				createrepo --skip-symlinks --update .
+			fi
+		)
+	done
+set -f
+}
+
+replace() {
+	if [[ -e $REPO_PATH ]]; then
+		mv "${REPO_PATH}" "${REPO_INSTANCE_PATH}-previous"
```
evgeni, line 54: should the formated version `"${REPO_INSTANCE_PATH}-previous"` be saved in a var, like `$REPO_INSTANCE_PATH_PREV`? So we can change it in one place if ever needed

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
...
+			if [[ -n "$latest" ]] ; then
+				ln -sf "$latest" foreman-client-release.rpm
+			fi
+
+			if [[ $MERGE == true ]] ; then
+				createrepo --skip-symlinks --update .
+			fi
+		)
+	done
+set -f
```
evgeni, line 49: should be aligned with the `set +f`

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
...
+	else
+		opts+=('--delete')
+	fi
+	rsync "${opts[@]}" --log-file-format 'CHANGED %f' "${REPO_SOURCE_RPM}/*" "${REPO_INSTANCE_PATH}/"
+	rsync "${opts[@]}" --log-file-format 'CHANGED source/%f' "${REPO_SOURCE_SRPM}/" "${REPO_INSTANCE_PATH}/source/"
+
+	set +f
+	for d in "${REPO_INSTANCE_PATH}"/*; do
+		(
+			cd "$d"
```
evgeni, line 32: pushd/popd?

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
...
+REPO_SOURCE_RPM="${REPO_SOURCE_BASE}/yum/${REPO_SOURCE}"
+REPO_SOURCE_SRPM="${REPO_SOURCE_BASE}/source/${REPO_SOURCE}"
+
+DEPLOY_TO="/var/www/vhosts/yum/htdocs"
+REPO_PATH="${DEPLOY_TO}/${REPO_DEST}"
+REPO_INSTANCE_PATH="${DEPLOY_TO}/$(dirname $REPO_DEST)/.$(basename $REPO_DEST)-$(date "+%Y%m%d%H%M%S")"
+
+REPO_RSYNC_LOG=$(mktemp)
+
+trap "rm -rf $REPO_RSYNC_LOG $REPO_INSTANCE_PATH" EXIT
```
evgeni, line 96: the old cap script has a rollback that moves back the -previous folder in case of an error, should we do too?

evgeni:
few comments on the actual capistrano2bash conversion, but nothing serious.

I think given 2.0 is out of the door it's the perfect time to go and break things with this :)

puppet/modules/web/files/deploy-yumrepo.sh
ekohl: I wasn't quite sure how to convert that and do it in bash. Suggestions are welcome.

puppet/modules/web/files/deploy-yumrepo.sh
ekohl: I'm not a fan of that. I've seen weird interactions when used with sudo where the user couldn't return to the previous directory. Changing directory in a subshell is what the previous script did and in my experience it works reliably.

ekohl:
> I think given 2.0 is out of the door it's the perfect time to go and break things with this :)

I was thinking that we should:
* spin up a new web node in OSUOSL
* deploy Puppet on it
* rsync all content
* test the new script
* prepare actual migration
  * Lower TTLs where needed / change all vhosts to be fronted by Fastly
  * prepare patches with all changes of hostnames
* Actual migration
  * rsync all content again
  * Apply prepared patches

puppet/modules/web/files/deploy-yumrepo.sh
evgeni: k

puppet/modules/web/files/deploy-yumrepo.sh
evgeni: my idea was also to use `trap`, but I didn't actually try it :)

evgeni:
rebased to resolve conflicts

ehelms:
@evgeni Needs another rebase. What is the next action needed to close this out?

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
+set -xe
+# This is a forced SSH command - uncomment to test locally
+set -f -- $SSH_ORIGINAL_COMMAND
+
+prepcache() {
+	if [[ -e $REPO_PATH ]]; then
+		if [[ $MERGE == false ]] && [[ $OVERWRITE == false ]] ; then
```
evgeni, line 9: this was `== true || == true` before, and I think that was wrong (the original code has `unless`, not `if`)

@ekohl would you mind double checking me?

ekohl: Yes, I think you're right and it makes more sense. If you read the error message, we should change `or` to `and` though.

puppet/modules/web/files/deploy-yumrepo.sh
ekohl: Yes, I think that's a good idea. Will you implement this or shall I?

puppet/modules/web/files/deploy-yumrepo.sh
ekohl: I wonder how much sense there is in a rollback.
* We sync everything to a full deep copy 
* We prepare everything in the copy
* We rename directories

If we fail somewhere along the way, we have an incomplete copy and just clean that up. The only thing that can fail is `mv "${REPO_INSTANCE_PATH}" "${REPO_PATH}"` and then we should place back the `previous`, but that's quite unlikely.

We may want to also catch `SIGINT` as well. Is that useful?

puppet/modules/web/files/deploy-yumrepo.sh
evgeni: Yepp, true

puppet/modules/web/files/deploy-yumrepo.sh
evgeni: I will.

evgeni:
given this has been running on web01 for a while, I will first merge this, fix the resulting permission chaos on web02, and then start re-syncing web01, as otherwise one ends up syncing wrong perms and rsync doesn't make this a pleasure.

PR #1335, opened 2020-04-14
Pin tito to 0.6.12 to workaround FetchBuilder issue

ehelms:
Working around this issue https://github.com/rpm-software-management/tito/pull/354#issuecomment-613523823

### DIFF
--- a/puppet/modules/slave/manifests/packaging/rpm.pp
+++ b/puppet/modules/slave/manifests/packaging/rpm.pp
@@ -63,8 +63,10 @@ class slave::packaging::rpm (
   }
 
   # tito
+  # Work around to fix https://github.com/rpm-software-management/tito/pull/354#issuecomment-613523823
+  # Pulled from the infra repository
   package { 'tito':
-    ensure => latest,
+    ensure => '0.6.12',
   }
 
   file { "${homedir}/.titorc":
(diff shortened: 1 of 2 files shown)


puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -63,8 +63,10 @@
   }
 
   # tito
+  # Work around to fix https://github.com/rpm-software-management/tito/pull/354#issuecomment-613523823
+  # Pulled from the infra repository
   package { 'tito':
-    ensure => latest,
```
mmoll, line 67: you also need to add it to the excludes in `puppet/modules/unattended/manifests/init.pp`

ehelms: Can you explain why that is if we are pinning it? Ohh I should probably explain more that this should pull this version from http://yum.theforeman.org/infra/el7/ since it no longer exists in EPEL.

ekohl: @ehelms we run yum-cron to update. If you don't do it there, yum-cron will update it and then Puppet will downgrade it. That is a bad experience for all. In the future we should replace all `latest` with `installed` so we only rely on yum-cron.

ehelms: Makes more sense now. Updated.

mmoll:
merged, thanks @ehelms!

PR #1338, opened 2020-04-19
Use nulldb for assets precompilation

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -95,7 +95,7 @@ pipeline {
                         }
                         stage("assets-precompile-2.5-postgres-ui") {
```
tbrisker, line 96: The stage name probably needs to be changed ;)

tbrisker: and perhaps it could be extracted to run in parallel to other steps

ehelms:
Added a parallel step, this is untested as of the moment. I will set aside some time to update Jenkins locally and run a few test builds.

ehelms:
Test run with this code change -- https://ci.theforeman.org/blue/organizations/jenkins/foreman-develop-source-release/detail/foreman-develop-source-release/594/pipeline/

PR #1340, opened 2020-04-20
Mash katello for EL8

ehelms:


koji/collection-mash-split.py
```diff
@@ -457,6 +457,10 @@ def main():
             mashes.append(pulpcore)
 
             if version != '3.15':
+                el8_katello = MashConfig(collection, version, "el8", "el8", "katello/el8")
+                el8_katello.name = 'katello-{}-el8'.format(version)
```
evgeni, line 461: this is the default and not required to be set, only pulpcore/candlepin needed that

PR #1341, opened 2020-04-22
Move motd module to stand alone

ehelms:


ehelms:
I created https://github.com/theforeman/puppet-motd with a copy of the module. The idea we had was extract this to a stand-alone module so we can use it for testing.

Adding to modulesync to follow up with fixing the repository up to our normal standards for release to the forge:

https://github.com/theforeman/foreman-installer-modulesync/pull/117

And then using for bats tests:

https://github.com/theforeman/forklift/pull/1151

.gitmodules
```diff
@@ -13,6 +13,9 @@
 [submodule "puppet/git_modules/htpasswd"]
 	path = puppet/git_modules/htpasswd
 	url = https://github.com/ghoneycutt/puppet-htpasswd
+[submodule "puppet/git_modules/motd"]
```
ekohl, line 16: This is now a git submodule, but in the forklift bats test we install from the forge. That means we should release it to the forge, which I'll prepare. Then we can add it to the Puppetfile instead since the goal is to get rid of the submodules.

puppet/Puppetfile_forge
```diff
@@ -40,6 +40,7 @@ mod 'saz/sudo', '6.0.0'
 mod 'theforeman/foreman', '13.1.0'
 mod 'theforeman/foreman_proxy', '12.1.0'
 mod 'theforeman/git', '6.0.1'
+mod 'theforeman/motd', '0.0.1'
```
ekohl, line 43:
```suggestion
mod 'theforeman/motd', '0.1.0'
```

mmoll:
merged, thanks @ehelms!

PR #1344, opened 2020-04-23
Drop sqlite testing from core PRs

tbrisker:
Rails 6 will not work with the sqlite version shipped in EL7. We are
dropping support for it. Nulldb testing is added in a seperate commit.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop_pr_core.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop_pr_core.yaml
@@ -34,7 +34,7 @@
             - fast
     execution-strategy:
       combination-filter: >
-        ((database == 'postgresql') && ((ruby == '2.5') || (ruby == '2.6'))) || ((database == 'postgresql-integrations') && (ruby == '2.5'))
+        (database == 'postgresql') || ((database == 'postgresql-integrations') && (ruby == '2.5'))
     builders:
       - shell: !include-raw: scripts/test/test_develop.sh
     publishers:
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_develop_pr_core.yaml
```diff
@@ -27,15 +27,14 @@
...
-            - sqlite3
       - axis:
           type: label-expression
           name: slave
           values:
             - fast
     execution-strategy:
       combination-filter: >
-        ((database == 'postgresql') && ((ruby == '2.5') || (ruby == '2.6'))) || ((database == 'sqlite3') && (ruby == '2.5')) || ((database == 'postgresql-integrations') && (ruby == '2.5'))
+        ((database == 'postgresql') && ((ruby == '2.5') || (ruby == '2.6'))) || ((database == 'postgresql-integrations') && (ruby == '2.5'))
```
ekohl, line 37: Untested, but how about:
```suggestion
        (database != 'postgresql-integrations') || (ruby == '2.5')
```

tbrisker: that would work, but I expect we'll be adding 2.7 soon, and at least for me it's easier to understand a list of the combinations we test instead of trying to figure the boolean logic of a negation

ekohl: I found the shorter version a lot easier to understand. However, if you want, there's also
```suggestion
        (database == 'postgresql') || ((database == 'postgresql-integrations') && (ruby == '2.5'))
```

tbrisker: works for me, updated, squash on merge please.

PR #1345, opened 2020-04-23
Allow bundler 2

mmoll:


mmoll:
I think this is needed as otherwise at least some Ruby 2.27 jobs will fail.

mmoll:
Thinking about this again, the problem where rather limits in the projects gemspecs, bundler 2 was used anyway, so let's check with Ruby 2.7 first...

PR #1349, opened 2020-04-28
Build core DEB PRs only on x86

mmoll:


ekohl:
Any reason to not do this for foreman-proxy? Should we just drop ARM altogether since it's being phased out by Scaleway?

mmoll:
All the other builds, except some dependencies are not built on ARM anyway.

I'd like to keep it as long as possible for 1.24 and 2.0.

ekohl:
Should we look at the branch then and only skip it in develop?

mmoll:
> Should we look at the branch then and only skip it in develop?

Like that? :)

PR #1350, opened 2020-04-28
Drop Ruby 2.0 from foreman-installer source build

ehelms:


mmoll:
merged, thanks @ehelms!

PR #1352, opened 2020-04-29
Enable plugins pushing to EL8

ehelms:


ehelms:
needs this build to pass -- https://github.com/theforeman/foreman-packaging/pull/5128

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/plugins.groovy
```diff
@@ -21,6 +21,12 @@ pipeline {
             steps {
                 // TODO: from variables
                 repoclosure('plugins', 'el7', foreman_version)
+
+                script {
+                    if (foreman_version != '2.0' || foreman_version != '1.24' || foreman_version != '1.23') {
```
ekohl, line 26: Rather than hardcoding it here, have you considered the TODO with variables?

ehelms: I have, except, I don't know what that TODO means :)

ekohl: In https://github.com/theforeman/foreman-infra/tree/master/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman we have variables for every release. The idea was to define the EL versions there. That prepares you for dropping EL7 as well. I didn't do that when I wrote this because we only had one EL version.

ehelms: Now I gotcha, updated

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman/1.23.groovy
```diff
@@ -14,5 +14,7 @@ def foreman_server_distros = [
     'ubuntu1604',
     'ubuntu1804'
 ]
-
+def foreman_supported_dists = [
```
ekohl, line 17: Perhaps this should be `foreman_redhat_releases` to match the debian counter part?

PR #1354, opened 2020-05-01
Re-factor top drop Katello nightly groovy and add parallel repoclosures

ehelms:


ehelms:
Client repoclosures are already running in parallel and we are generating new temp cache dir per run:

https://github.com/theforeman/obal/blob/master/obal/data/roles/repoclosure/tasks/repoclosure.yml#L22

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
```diff
@@ -33,8 +36,8 @@ pipeline {
             steps {
                 script {
                     runCicoJobsInParallel([
-                        ['name': 'Install test', 'job': "foreman-katello-${katello_version}-test"],
-                        ['name': 'Upgrade test', 'job': "foreman-katello-upgrade-${katello_version}-test"]
+                        ['name': 'centos7 install test', 'job': "foreman-katello-${katello_version}-test"],
+                        ['name': 'centos7 upgrade test', 'job': "foreman-katello-upgrade-${katello_version}-test"]
```
ekohl, line 40: Have you considered renaming the jobs to include the version in the name, like we do in Foreman? Or is that a future refactor?

ehelms:
I should note this is part of a larger set of re-factoring and cleanup I started doing when I wanted to add centos8 support for Katello. So I am trying to break the changes into smaller targeted PRs.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
ehelms: Hah, I was just writing a comment. There is a larger set of changes I started making to bring cohesion to our job naming conventions and reduce overhead of declaring pipeline jobs. So I'm trying to do as best I can smaller targeted changes to build up to that point.

ekohl: I suspected as much.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/foreman-pipeline.groovy
```diff
@@ -38,12 +42,13 @@ pipeline {
...
         stage('Install Test') {
             agent { label 'el' }
+
             steps {
                 script {
                     def pipelines = foreman_server_distros.collect { os ->
                         [
                             name: os,
-                            job: "foreman-${foreman_version}-${os}-release-test",
+                            job: "foreman-${foreman_version}-${os}-install-test",
```
ekohl, line 51: Will you rename all jobs in CentOS CI as well? I don't see that in this PR.

ehelms: Ahh yea, that was a leftover from breaking changes out. Will put tha tback for this PR.

PR #1355, opened 2020-05-03
Re-factor job-templates to drop individual version groovy files

ehelms:
Modifies the names of some existing jobs to align all of them to
a common convention:

namespace-type-version-distro-action-test

For example:

foreman-nightly-debian9-upgrade-test
foreman-katello-nightly-centos7-install-test


ehelms:
The new set of generated jobs:

```
foreman-1.23-centos7-install-test
foreman-1.23-debian9-install-test
foreman-1.23-debian10-install-test
foreman-1.23-ubuntu1604-install-test
foreman-1.23-ubuntu1804-install-test
foreman-1.24-centos7-install-test
foreman-1.24-debian9-install-test
foreman-1.24-debian10-install-test
foreman-1.24-ubuntu1604-install-test
foreman-1.24-ubuntu1804-install-test
foreman-2.0-centos7-install-test
foreman-2.0-debian9-install-test
foreman-2.0-debian10-install-test
foreman-2.0-ubuntu1604-install-test
foreman-2.0-ubuntu1804-install-test
foreman-ci-centos-simple-test
foreman-discovery-image-build
foreman-katello-3.13-centos7-install-test
foreman-katello-3.13-centos7-upgrade-test
foreman-katello-3.14-centos7-install-test
foreman-katello-3.14-centos7-upgrade-test
foreman-katello-3.15-centos7-install-test
foreman-katello-3.15-centos7-upgrade-test
foreman-katello-devel-centos7-install-test
foreman-katello-nightly-centos7-install-test
foreman-katello-nightly-centos7-upgrade-test
foreman-katello-nightly-centos8-install-test
foreman-luna-nightly-centos7-install-test
foreman-luna-nightly-centos7-upgrade-test
foreman-luna-nightly-centos8-install-test
foreman-nightly-centos7-install-test
foreman-nightly-centos7-upgrade-test
foreman-nightly-centos8-install-test
foreman-nightly-debian10-install-test
foreman-nightly-debian10-upgrade-test
foreman-nightly-ubuntu1804-install-test
foreman-nightly-ubuntu1804-upgrade-test
```

ehelms:
I debated going with `foreman-foreman` to match the rest of the job conventions given `foreman-` is enforced at the front as a namespace and today we use it for both namespace and indicator that it's foreman core.

ehelms:
I re-used the `parameters` tactic to allow the job-template to control everything and not have to repeat ourselves in groovy files. I generally do not like this approach since these jobs are "hard-coded" to their version and type and the parameters give a sense that they can be changed. I could not find any other way to include content from the job-template definition into the groovy layer.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
```diff
@@ -1,45 +1,80 @@
...
 - job-template:
-    name: 'foreman-katello-{version}-test'
-    project-type: pipeline
+    name: foreman-katello-{version}-{distro}-{action}-test
+    project-type: 'pipeline'
     sandbox: true
+    parameters:
+      - string:
+          name: expected_version
+          default: 'UNKNOWN'
```
ekohl, line 9: `UNKNOWN` will cause the tests to fail. It should be empty if you want to skip it:
https://github.com/theforeman/forklift/blob/0f03bebf38c6941aecb50923585c9f7515aa3fed/bats/fb-test-foreman.bats#L6-L10

In the Foreman pipelines this is ok because it's called with it:
https://github.com/theforeman/theforeman-rel-eng/blob/2d82d085b4a49a723fad33cdeb7ee90c599dd5d1/release_pipeline#L5

We can set it to `2.1.0-develop` on nightly.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
```diff
@@ -0,0 +1,75 @@
...
+    distro:
+      - centos7
+      - debian9
+      - debian10
+      - ubuntu1604
+      - ubuntu1804
+    version:
+      - '1.23'
+      - '1.24'
+      - '2.0'
```
ekohl, line 44: Some combos of this are invalid. See https://github.com/theforeman/forklift/blob/master/vagrant/config/versions.yaml as well. Like Debian 9 / Ubuntu 16.04 + Foreman 2.0.

ekohl:
> I re-used the `parameters` tactic to allow the job-template to control everything and not have to repeat ourselves in groovy files. I generally do not like this approach since these jobs are "hard-coded" to their version and type and the parameters give a sense that they can be changed. I could not find any other way to include content from the job-template definition into the groovy layer.

I've also considered this. There are also properties, maybe those are better?

ehelms:
The way to do it with properties it to use [Env Inject](https://plugins.jenkins.io/envinject/) and create environment variables.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
```diff
@@ -1,45 +1,80 @@
...
 - job-template:
-    name: 'foreman-katello-{version}-test'
-    project-type: pipeline
+    name: foreman-katello-{version}-{distro}-{action}-test
+    project-type: 'pipeline'
     sandbox: true
+    parameters:
+      - string:
+          name: expected_version
+          default: 'nightly'
```
ekohl, line 9: For it to pass, I think this should match https://github.com/theforeman/foreman/blob/develop/VERSION (and foreman-installer and foreman-proxy which have the same content):
```suggestion
          default: '2.1-develop'
```

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-luna-test.yml
```diff
@@ -1,39 +1,53 @@
 ---
 - job-template:
-    name: foreman-{foreman_version}-{distro}-release-test
+    name: foreman-luna-{version}-{distro}-{action}-test
     project-type: 'pipeline'
     sandbox: true
     parameters:
       - string:
           name: expected_version
           default: 'UNKNOWN'
```
ekohl, line 9:
```suggestion
          default: '2.1.0-develop'
```

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -1,3 +1,5 @@
+playBook = pipelineVars(action: params.action, type: params.type, version: params.version, os: params.distro, extra_vars: ['foreman_expected_version': params.expected_version])
```
ekohl, line 1: Don't you need to remove `playbook = playBookVars()` on line 6 if you define it here?

ehelms:
Updated job list:

```
foreman-1.23-centos7-install-test
foreman-1.23-debian9-install-test
foreman-1.23-ubuntu1604-install-test
foreman-1.23-ubuntu1804-install-test
foreman-1.24-centos7-install-test
foreman-1.24-debian9-install-test
foreman-1.24-debian10-install-test
foreman-1.24-ubuntu1604-install-test
foreman-1.24-ubuntu1804-install-test
foreman-2.0-centos7-install-test
foreman-2.0-debian10-install-test
foreman-2.0-ubuntu1804-install-test
foreman-ci-centos-simple-test
foreman-discovery-image-build
foreman-katello-3.13-centos7-install-test
foreman-katello-3.13-centos7-upgrade-test
foreman-katello-3.14-centos7-install-test
foreman-katello-3.14-centos7-upgrade-test
foreman-katello-3.15-centos7-install-test
foreman-katello-3.15-centos7-upgrade-test
foreman-katello-devel-centos7-install-test
foreman-katello-nightly-centos7-install-test
foreman-katello-nightly-centos7-upgrade-test
foreman-katello-nightly-centos8-install-test
foreman-luna-nightly-centos7-install-test
foreman-luna-nightly-centos7-upgrade-test
foreman-luna-nightly-centos8-install-test
foreman-nightly-centos7-install-test
foreman-nightly-centos7-upgrade-test
foreman-nightly-centos8-install-test
foreman-nightly-debian10-install-test
foreman-nightly-debian10-upgrade-test
foreman-nightly-ubuntu1804-install-test
foreman-nightly-ubuntu1804-upgrade-test
```

ekohl:
I still struggle with the expected version so I'm go to write out my thoughts.

First of all, it should really be read as `expected_foreman_version`.

For Foreman jobs it's ok because we can hardcode nightly (we have to update during branching anyway to add the x.y release jobs). Release jobs are called with the correct version (https://github.com/theforeman/theforeman-rel-eng/blob/2d82d085b4a49a723fad33cdeb7ee90c599dd5d1/ci-jobs#L137-L141) via scripts as well. So I'm not too worried about that. They also provide value because they ensure the correct packages have been released in the test.

For Katello and Luna, it's different. Nothing ensures the versions are correctly set. Again, nightly can be hardcoded, but there's nothing for release jobs. We can copy the code in the rel eng tools, but what's the value? Katello might care about the Foreman x.y version, but Foreman x.y.z isn't that relevant. That's why previously it was empty, so it was skipped.

So in summary: I think Foreman needs a default (`UNKNOWN` is good as a general template). Then nightly should override the default via yaml files. release jobs should be invoked with a specific value "by hand". Katello and luna do not benefit from this check and should be set to an empty string.

Does that help?

ehelms:
Bringing congruence, and the pain that job-templates translating to groovy can cause sometimes pushed to the addition of the parameter. So I just removed it where it doesn't belong now.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
```diff
@@ -0,0 +1,102 @@
...
+    dsl:
+      !include-raw:
+        - pipelines/lib/foremanCentosJob.groovy{empty}
+        - pipelines/lib/pipelines.groovy{empty}
+        - ../theforeman.org/pipelines/lib/duffy.groovy{empty}
+        - ../theforeman.org/pipelines/lib/ansible.groovy{empty}
+        - ../theforeman.org/pipelines/lib/foreman_infra.groovy{empty}
+
+- project:
+    name: foreman-release-test-1.23
```
ekohl, line 31: 1.23 is no longer supported now that 2.0 is GA so can be dropped. Perhaps that we should do that first in master?

ehelms:
Job sets:

```
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.23-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.23-debian9-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.23-ubuntu1604-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.23-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-debian9-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-ubuntu1604-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-ci-centos-simple-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-discovery-image-build'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.13-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.13-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.14-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.14-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.15-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.15-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-devel-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-debian10-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-ubuntu1804-upgrade-test'
```

ehelms:
@evgeni I'd appreciate your thoughts on this re-work.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
```diff
@@ -1,46 +1,78 @@
...
-      - 'foreman-katello-{version}-test'
-      - 'foreman-katello-upgrade-{version}-test'
+      - 'foreman-katello-{version}-{distro}-{action}-test'
     empty: ''
     version:
       - '3.13'
       - '3.14'
       - '3.15'
       - '3.16'
+    distro:
```
evgeni, line 37: wouldn't it make more sense to group by distro?

so centos7 would include all releases and nightly for install and upgrade
and centos8 would be 3.16 and nightly, install only?

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
```diff
@@ -0,0 +1,102 @@
...
+      - debian10
+      - ubuntu1604
+      - ubuntu1804
+    version:
+      - '1.24'
+    action:
+      - install
+
+- project:
+    name: foreman-release-test-2.0
```
evgeni, line 62: missing 2.1 ;)

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
evgeni: or group by "install-and-upgrade" and "install-only", so we can move 3.16 to upgradeable on el8 at some point?

ehelms: I could go that route. I think it tends to just be easier to think in terms of versions of the software rather than in distributions? Unless I misunderstand your question.

evgeni: I think what bothers me is that there is no 3.16-el8 job because 3.16 is in the "release" bucket and that has no el8 defined at all. so my second comment is probably better aligned with your original idea, it would look like this:

```yaml
- project:
    name: katello-install-el7
    jobs:
      - 'foreman-katello-{version}-{distro}-{action}-test'
    empty: ''
    version:
      - '3.14'
      - '3.15'
      - '3.16'
      - 'nightly'
      - 'devel'
    distro:
      - centos7
    action:
      - install

- project:
    name: katello-install-el8
    jobs:
      - 'foreman-katello-{version}-{distro}-{action}-test'
    empty: ''
    version:
      - '3.16'
      - 'nightly'
    distro:
      - centos8
    action:
      - install

- project:
    name: katello-upgrade
    jobs:
      - 'foreman-katello-{version}-{distro}-{action}-test'
    empty: ''
    version:
      - '3.14'
      - '3.15'
      - '3.16'
      - 'nightly'
    distro:
      - centos7
    action:
      - upgrade
```

would also fold the `devel` one into the generic install-el7 target.

ehelms:
I tried a modified approach to Katello but focused still on version. How does that feel? If still yucky, I'll move to OS based.

```
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-debian9-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-ubuntu1604-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-1.24-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.0-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.1-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.1-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.1-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-2.1-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-ci-centos-simple-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-discovery-image-build'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.14-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.14-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.15-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.15-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-3.16-centos8-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-devel-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-katello-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-luna-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos7-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos7-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-centos8-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-debian10-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-debian10-upgrade-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-ubuntu1804-install-test'
DEBUG:jenkins_jobs.builder:Writing XML to '/tmp/jobs/foreman-nightly-ubuntu1804-upgrade-test'
```

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
```diff
@@ -36,8 +36,8 @@ pipeline {
             steps {
                 script {
                     runCicoJobsInParallel([
-                        ['name': 'centos7 install test', 'job': "foreman-katello-${katello_version}-test"],
-                        ['name': 'centos7 upgrade test', 'job': "foreman-katello-upgrade-${katello_version}-test"]
+                        ['name': 'centos7 install test', 'job': "foreman-katello-${katello_version}-install-test"],
```
evgeni, line 39: this should also run el8 install test I guess? but then how would this step handle the fact that el8 tests don't exist for some versions?

evgeni:
I think this improves the current situation a lot (and the version-groovy files have been a pain).

I'd be OK with merging this as-is as it basically only refactors the current jobs (and adds katello nightly el8) and then think once more about the "matrix" we want to execute in terms of "which versions on which OS are only installable, and which upgradeable"

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
ehelms: Good find. I've made this dynamic now based on configuration.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/katello/3.16.groovy
```diff
@@ -3,3 +3,10 @@ def katello_version = '3.16'
 def foreman_el_releases = [
     'el7'
 ]
+def katello_server_distros = [
+    'centos7',
+    'centos8'
```
ekohl, line 8: This is odd since `foreman_el_releases` doesn't include el8 which means the RPMs are not being published. Is this intentional?

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-luna-test.yml
```diff
@@ -1,41 +1,50 @@
...
+          default: '{version}'
       - string:
           name: distro
           default: '{distro}'
+      - string:
+          name: action
+          default: '{action}'
+      - string:
+          name: type
+          default: 'katello'
```
ekohl, line 18: Isn't luna its own type?

evgeni: It is. Just wanted to comment the same. Also the luna groovy files can be deleted now

evgeni:
No, that's not yucki anymore. :+1:

ehelms:
I had meant to only include EL8 for the ci.centos.org jobs right now (to help inform the re-factor). If I include them in the Katello pipelines themselves they will fail (due to yum bats tests) and block EL7.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
```diff
@@ -1,46 +1,91 @@
...
+- project:
+    name: katello-release-3.16
+    jobs:
+      - 'foreman-katello-{version}-{distro}-{action}-test'
+    empty: ''
+    version:
       - '3.16'
+    distro:
+      - centos7
+      - centos8
```
evgeni, line 50: I thought you wanted to leave el8 out for now?
But more so: does that combo means it will try a 3.14→3.15→3.16 upgrade on el8 (and fail?)

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
```diff
@@ -0,0 +1,102 @@
+---
+- job-template:
+    name: foreman-{version}-{distro}-{action}-test
+    project-type: 'pipeline'
+    sandbox: true
+    parameters:
+      - string:
+          name: expected_version
+          default: 'UNKNOWN'
```
evgeni, line 9: this default will break nightly tests that you now have folded into the same test definition.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
```diff
@@ -35,10 +35,21 @@ pipeline {
 
             steps {
                 script {
-                    runCicoJobsInParallel([
-                        ['name': 'centos7 install test', 'job': "foreman-katello-${katello_version}-test"],
-                        ['name': 'centos7 upgrade test', 'job': "foreman-katello-upgrade-${katello_version}-test"]
-                    ])
+                    def pipelines = katello_server_distros.collect { os ->
+                        [
+                            name: os,
```
evgeni, line 40: should this be `"${os}-install"` and `"${os}-upgrade"` below?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/luna.groovy
```diff
@@ -14,8 +14,9 @@ pipeline {
             steps {
                 script {
                     runCicoJobsInParallel([
-                        ['name': 'Install test', 'job': 'foreman-luna-nightly-test'],
-                        ['name': 'Upgrade test', 'job': 'foreman-luna-upgrade-nightly-test']
+                        ['name': 'centos7 install test', 'job': 'foreman-luna-centos7-nightly-install-test'],
```
evgeni, line 17: any reason not to use the same "dynamic" thing as in katello?

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-katello-test.yml
evgeni: no, it will be just defined, not run, good good

evgeni:
I'm all for merging this after my last comments are addressed and iterating later for any more changes.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
ehelms: What part breaks? (maybe I can fix it?)

evgeni: We have a test in `forklift` that verifies that the expected version is the one installed:
https://github.com/theforeman/forklift/blob/42a15959db28e1bf6b51e3982952094ec97dff70/bats/fb-test-foreman.bats#L6-L10

It only triggers if the environment variable is actually set, but my reading of this line and https://github.com/theforeman/foreman-infra/pull/1355/files#diff-0bccc46806d9b5ebbc8e2aacf5a8717eR1 suggests it will be always set.

ehelms: Perhaps we just set this for nightly to the version that is currently set in develop? Unify that test across releases and nightly.

evgeni: This is also what @ekohl suggested in https://github.com/theforeman/foreman-infra/pull/1355#discussion_r419144553 -- sounds good to me

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
```diff
@@ -0,0 +1,102 @@
+---
+- job-template:
+    name: foreman-{version}-{distro}-{action}-test
+    project-type: 'pipeline'
+    sandbox: true
+    parameters:
+      - string:
+          name: expected_version
+          default: '2.2.0-develop'
```
evgeni, line 9: I've been wondering if we could somehow make this "automatic" to avoid forgetting to change that during branching, but Jenkins has no knowledge what the "next" version will be. So I'm OK with leaving this for now.

evgeni:
@ekohl any objections from you on the recent changes?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
```diff
@@ -35,10 +35,21 @@ pipeline {
...
+                            job: "foreman-katello-${katello_version}-${os}-install-test"
+                        ]
+                    }
+
+                    pipelines += katello_upgrade_server_distros.collect { os ->
+                        [
+                            name: "${os}-upgrade",
+                            job: "foreman-katello-${katello_version}-${os}-upgrade-test"
+                        ]
+                    }
```
ekohl, line 50: Can't we have a config data structure like this (syntax may not be correct):
```groovy
def pipelines = [
    'install': ['centos7'],
    'upgrade': ['centos7']
]
```

I think this makes the variables more self documenting and this pipeline code more generic. We can also introduce a helper (shared with luna) that has the signature:
```groovy
run_pipelines_in_cico(type, version, pipelines)
```

Katello would call it as `run_pipelines_in_cico('katello', katello_version, pipelines)` and luna as `run_pipelines_in_cico('luna', 'nightly', pipelines)`.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
ekohl: There's still the issue that this is hard to do with Katello. When releasing Katello 3.15.z, do we know if the Foreman version is 2.0.0 or 2.0.1 at that point? Do we care, or can we use 2.0.*? In the end it's grep so that may work.

ekohl:
Oh, to be clear, I don't mind merging this but just expect that it will be incomplete.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
ehelms: This entry here is only used for the Foreman pipelines. The Katello job definitions don't include this parameter. So we can align on similar pattern latter on but I believe this will not affect Katello.

ekohl: Ah right. I expected a single template that was used for Foreman, Katello and Luna. Given they all use the exact same DSL and only use different parameters.

ehelms: That would be fully true *if* I change the name of the foreman jobs to be (https://github.com/theforeman/foreman-infra/pull/1355#issuecomment-623159701), e.g:

foreman-foreman-nightly-centos7-install-test

As a reminder, we have to include `foreman-` at the beginning of job names to namespace the jobs within ci.centos.org, what follows after is our convention. I know `foreman-foreman` looks weird but if we follow the same convention for all it does make it easier to have a single definition.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/katello.groovy
ehelms: I'll take a look at this, I don't want to merge till Monday when more devs are around in case I break things anyway.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-test.yml
evgeni: I'm happy with foreman-foreman, aligns well with our ansible stuff

ekohl: Then I'd suggest to consider `foreman-pipeline-{type}-{version}-{os}-{pipeline}`

ehelms: Another spot it helps, the function I added (https://github.com/theforeman/foreman-infra/pull/1355/files#diff-06c5ebd36e8d5d56d40a13f779b10235R100) will require modified logic to handle Foreman pipelines without strictly sticking to the convention.

ehelms: To ensure clarity, the job names would be:

foreman-pipeline-foreman-nightly-centos7-upgrade
foreman-pipeline-foreman-nightly-centos7-install
foreman-pipeline-katello-nightly-centos7-upgrade
foreman-pipeline-katello-nightly-centos8-install
foreman-pipeline-luna-nightly-centos8-install

ehelms:
Latest update job set naming:

foreman-ci-centos-simple-test
foreman-discovery-image-build
foreman-pipeline-foreman-1.24-centos7-install
foreman-pipeline-foreman-1.24-debian9-install
foreman-pipeline-foreman-1.24-debian10-install
foreman-pipeline-foreman-1.24-ubuntu1604-install
foreman-pipeline-foreman-1.24-ubuntu1804-install
foreman-pipeline-foreman-2.0-centos7-install
foreman-pipeline-foreman-2.0-debian10-install
foreman-pipeline-foreman-2.0-ubuntu1804-install
foreman-pipeline-foreman-2.1-centos7-install
foreman-pipeline-foreman-2.1-centos8-install
foreman-pipeline-foreman-2.1-debian10-install
foreman-pipeline-foreman-2.1-ubuntu1804-install
foreman-pipeline-foreman-nightly-centos7-install
foreman-pipeline-foreman-nightly-centos7-upgrade
foreman-pipeline-foreman-nightly-centos8-install
foreman-pipeline-foreman-nightly-debian10-install
foreman-pipeline-foreman-nightly-debian10-upgrade
foreman-pipeline-foreman-nightly-ubuntu1804-install
foreman-pipeline-foreman-nightly-ubuntu1804-upgrade
foreman-pipeline-katello-3.14-centos7-install
foreman-pipeline-katello-3.14-centos7-upgrade
foreman-pipeline-katello-3.15-centos7-install
foreman-pipeline-katello-3.15-centos7-upgrade
foreman-pipeline-katello-3.16-centos7-install
foreman-pipeline-katello-3.16-centos7-upgrade
foreman-pipeline-katello-3.16-centos8-install
foreman-pipeline-katello-3.16-centos8-upgrade
foreman-pipeline-katello-devel-centos7-install
foreman-pipeline-katello-nightly-centos7-install
foreman-pipeline-katello-nightly-centos7-upgrade
foreman-pipeline-katello-nightly-centos8-install
foreman-pipeline-luna-nightly-centos7-install
foreman-pipeline-luna-nightly-centos7-upgrade
foreman-pipeline-luna-nightly-centos8-install

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/katello/nightly.groovy
```diff
@@ -3,3 +3,7 @@ def katello_version = 'nightly'
 def foreman_el_releases = [
     'el7'
 ]
+def pipelines = [
```
ehelms, line 6: Added this new variable to each set of variables.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -90,6 +90,21 @@ def runCicoJobsInParallel(jobs) {
     }
 }
 
+runCicoPipelines(project, version, config) {
```
ehelms, line 93: Added this function to generate the pipelines array with correct naming based on simple input.

ekohl: This is indeed what I had in mind.

puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-pipeline-template.yml
```diff
@@ -1,41 +1,28 @@
...
 - job-template:
-    name: foreman-{foreman_version}-{distro}-release-test
+    name: foreman-pipeline-{type}-{version}-{distro}-{action}
     project-type: 'pipeline'
     sandbox: true
     parameters:
       - string:
           name: expected_version
-          default: 'UNKNOWN'
+          default: '2.2'
```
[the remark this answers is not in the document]
ehelms: I understood @evgeni assumption here was that it was a grep in the bats tests so any substring here would work.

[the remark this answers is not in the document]
evgeni: Mine? I think it was Ewoud. Either way:

https://github.com/theforeman/forklift/blob/master/bats/fb-test-foreman.bats#L9 uses -x which means "whole line match", so it wouldn't match.

You'd need to default to `2\.2.*` or so (btw, we never eskape those dots). Or an empty string and thus make the yest skip?

puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -1,8 +1,7 @@
+playBook = pipelineVars(action: params.action, type: params.type, version: params.version, os: params.distro, extra_vars: ['foreman_expected_version': params.expected_version ?: null])
```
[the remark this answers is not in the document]
ehelms: This information is used in the `post` section so I think global to the file is needed.

ehelms:
Merging, and testing changes today. I will submit additional fixes if any are needed or revert if it all blows up :)

PR #1362, opened 2020-05-13
Update mash script for 2.1

ehelms:


koji/collection-mash-split.py
```diff
@@ -403,7 +403,6 @@ def main():
             "rhel7": "el7",
             "rhel6": "el6",
             "rhel5": "el5",
-            "fedora29": "fc29",
```
ekohl, line 406: This also removes it from all previously released versions. I'm not sure that's correct.

ehelms: Ahh right, ok added a condition to add it for relevant releases.

PR #1363, opened 2020-05-13
Add 2.1 jobs

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_2_1_stable.yaml
```diff
@@ -0,0 +1,40 @@
...
+          name: ruby
+          values:
+            - 2.5
+            - 2.6
+      - axis:
+          type: user-defined
+          name: database
+          values:
+            - postgresql
+            - sqlite3
```
tbrisker, line 25: This is probably not going to work

mmoll: GH-1364

PR #1365, opened 2020-05-19
add centos8 to centos ci jobs

evgeni:


evgeni:
we have jobs for older ones, so that one won't hurt. we need to clean up at some point.

PR #1368, opened 2020-05-22
Add el8 to katello pipeline

pcreech:
It was requested to keep this change to nightly at least for now


ehelms:
This is only part of the change, the ci.centos.org jobs also need created and defined to handle centos8. My PR (https://github.com/theforeman/foreman-infra/pull/1355/files) attempts to do this while making it require less places to add things.

PR #1370, opened 2020-05-27
Add discovery 16.1 branch

lzap:
SSIA ty


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
```diff
@@ -4,5 +4,7 @@
     branch:
       - develop:
           foreman_branch: develop
+      - 16.1-stable:
+          foreman_branch: 16.1-stable
```
ekohl, line 8: I'm guessing this is for 2.1:
```suggestion
          foreman_branch: 2.1-stable
```

PR #1372, opened 2020-05-29
Silence deprecation warnings from Redmine

ekohl:
By introducing a wrapper function that silences the deprecation warnings, my inbox should be a lot more quiet. The proper solution is to upgrade Redmine but that's on the roadmap.


mmoll:
merged, bedankt @ekohl!

PR #1373, opened 2020-05-30
drop i386 builds for Debian completely

mmoll:


puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/debian/execute_pbuilder_dependency.sh
```diff
@@ -52,9 +52,6 @@ fi
 if grep -qe "Architecture:\s\+any" debian/control; then
   if [ $arch != x86 ]; then
```
ekohl, line 53: I think this can be unified with the above `if [ $arch = x86 ]; then`

mmoll: indeed.

PR #1377, opened 2020-06-05
disable arm builds for 2.1 and nightly

evgeni:


evgeni:
@mmoll I have no idea what I am doing here, so please help :)

mmoll:
In these two jobs filtering needs to be done for `arch` and basically the existing filter needs to get extended by one that comes back true for `(arch = "armv8" && (version == "2.0" || version == "1.24")`. To make things easier, 1.22 and 1.23 can get IMHO removed anyway.

mmoll:
> To make things easier, 1.22 and 1.23 can get IMHO removed anyway.

See GH-1378

mmoll:
needs a rebase now :)

evgeni:
@mmoll done, also simplified the focal exclude, now that we don't build armv8 at all in nightly

PR #1378, opened 2020-06-06
Remove Foreman 1.23 and Katello 3.13 configurations

mmoll:


koji/collection-mash-split.py
```diff
@@ -412,10 +412,10 @@ def main():
         else:
             git_tag = "rpm/{}".format(version)
 
-            if version in ('1.22', '1.23', '1.24'):
+            if version in ('1.24'):
```
ekohl, line 415: You're changing it from a tuple to a string. Either 
```suggestion
            if version in ('1.24',):
```
or
```suggestion
            if version in ['1.24']:
```
or
```suggestion
            if version == '1.24':
```
are valid solutions

mmoll: fixed.

PR #1379, opened 2020-06-08
Remove foreman_remote_execution CI jobs

adamruzicka:


mmoll:
Should the PR tests also get removed?

mmoll:
Merged, díky @adamruzicka.

Somebody:tm: should remove the webhooks in the Github repo.

adamruzicka:
Disabled the webhook

tbrisker:
@adamruzicka why did you remove the build test job? iiuc, the GH action only replaces the PR tests. And considering that the PR action runs on ubuntu, i'm worried this might lead to failures on centos we'll only find when building the packages if build test aren't run after merge.

PR #1385, opened 2020-06-19
Set default expected foreman version to explicit nightly version

ehelms:


mmoll:
merged, thanks @ehelms!

PR #1386, opened 2020-06-22
Fix luna nightly

ehelms:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/luna-nightly-rpm-pipeline.yaml
```diff
@@ -10,6 +10,7 @@
           result: success
     dsl:
       !include-raw:
+        - pipelines/vars/katello/nightly.groovy
```
tbrisker, line 13: should luna have it's own vars?

ehelms: Doesn't need to but I can go that route

tbrisker: ok, let's start with this and split it only if needed

PR #1387, opened 2020-06-24
enable upgrade jobs for 2.0 and 2.1

evgeni:


puppet/modules/jenkins_job_builder/files/centos.org/jobs/foreman-pipelines.yml
```diff
@@ -47,6 +48,7 @@
       - '2.1'
     action:
       - install
+      - upgrade
```
evgeni, line 51: this will also generate el8 upgrade jobs on ci.centos.org, but nothing will ever trigger them :)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman/2.0.groovy
```diff
@@ -17,5 +17,9 @@ def pipelines = [
         'centos7',
         'debian10',
         'ubuntu1804'
+    ],
+    'upgrade': [
+        'centos7',
+        'ubuntu1804'
```
evgeni, line 23: no debian10 here, as upgrading to 2.0 means 1.23→1.24→2.0 and there is no 1.23 on debian10.

ekohl: I thought https://github.com/theforeman/forklift/blob/cee201325469f2228195068fda7ea0be68048313/roles/forklift_versions/lookup_plugins/forklift_versions.py#L97-L98 dealt with this by duplicating the first entry until it's there. So it'll test 1.24 -> 1.24 -> 2.0. https://github.com/theforeman/forklift/blob/cee201325469f2228195068fda7ea0be68048313/pipelines/upgrade/05-server_to_intermediate.yaml#L21 then skips 1.24 -> 1.24 upgrade.

So as long as there's n-1, it's valid to list.

evgeni: This handles the case when the previous version is not defined at all. In this case it's defined, but just not for that specific OS, and that's something `forklift_versions` does not check.

ekohl: Perhaps we can add that check by passing in the OS to the lookup?

evgeni: bleh, more work ;-)

evgeni: https://github.com/theforeman/forklift/pull/1172

evgeni:
@ekohl good call, added!

PR #1388, opened 2020-06-24
correct type in non-foreman pipelines

evgeni:


evgeni:
travis is slow, and I want this in,sorry

PR #1389, opened 2020-06-24
pass expected version from ci.foreman to ci.centos again

evgeni:


evgeni:
Mhh, no, that needs a ?null too

PR #1391, opened 2020-06-25
clean old ci.centos.org jobs automatically

evgeni:
this updates the URL of the jenkins node, as you can reach the API also
*below* a view, but then are limited to that view -- exactly what we
need to delete only our jobs :)


evgeni:
@zjhuntin :eyes:

evgeni:
(yes, I tested that, it works)

PR #1394, opened 2020-07-08
Use EL8 nodes on CentOS CI

ekohl:
This is a step in dropping Ruby 2.0 support in forklift. Rather than migrating to SCL or similar, this has the benefit that we get the latest technologies altogether.

It's entirely untested but came from https://github.com/theforeman/forklift/pull/1179#discussion_r451463667

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
@@ -7,6 +7,7 @@
       cico:
         action: get
         api_key: "{{ api_key }}"
+        release: 8
       register: cico_data
 
     - name: 'Retry geting list of nodes'


evgeni:
The number of EL8 nodes in CICO is significantly lower than EL7 (https://wiki.centos.org/QaWiki/CI/Duffy) so I think this might not be a good idea *right now*

ehelms:
I agree with @evgeni -- the cost here could be slower or backed up pipelines due to the smaller ready pool on `ci.centos.org`.

ekohl:
I wasn't aware of that. Does it make sense to reach out to the CentOS folks to ask their opinion? Perhaps if we indicate why we're using EL7 now, they can take that into account. There's a good chance that they see everyone using EL7 and little demand for EL8, keeping the status quo.

evgeni:
@ekohl absolutely. Do we have a direct contact there, or shall we just open a bug on bugs.centos.org?

ekohl:
I'm not aware of a direct contact.

ekohl:
Closing for now.

PR #1397, opened 2020-07-21
Do not test foreman-tasks on Jenkins

ezr-ondrej:
https://github.com/theforeman/foreman-tasks/pull/574 is introducing the GH actions, so we don't need to test `foreman-tasks` on Jenkins anymore.


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/plugins/foreman-tasks.yaml
```diff
@@ -1,8 +0,0 @@
-- project:
```
ekohl, line 1: This one runs after merging something. Is that something you wish to drop? The Github Actions don't test that.

ezr-ondrej:
I've kept the job file then :-1:

adamruzicka:
Could we get this in?

PR #1398, opened 2020-07-28
run yum reposync also on web01.osuosl

evgeni:


evgeni:
`web01.osuosl` has #1333 deployed (manually), and I'd like to test syncing yum to it in a production-like setup

PR #1399, opened 2020-07-29
the freight jobs use ruby, so let's install it

evgeni:


puppet/modules/freight/manifests/user.pp
```diff
@@ -8,6 +8,8 @@
 ) {
   require freight
 
+  ensure_packages(['ruby'])
```
ekohl, line 11: Does it make sense to add `require => Package['ruby']` to `file { "/etc/cron.daily/${user}":` so we know why it is we need it? It also means we don't deploy it unless Ruby could actually be installed.

evgeni: Probably

evgeni: done

PR #1401, opened 2020-08-03
Update stable version to 2.1

tbrisker:


tbrisker:
added in https://github.com/theforeman/tool_belt/pull/291 (along with some other stuff i keep forgetting to do)

PR #1405, opened 2020-08-11
deploy deb packages to old and new web host

evgeni:
note: should not be merged before 2.2 RC1 is out.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
@@ -23,6 +23,7 @@ void push_rpms_katello(version) {
 void push_debs_direct(os, repo) {
     sshagent(['freight']) {
         sh "ssh freight@web02.theforeman.org deploy ${os} ${repo}"
+        sh "ssh freight@web01.osuosl.theforeman.org deploy ${os} ${repo}"
     }
 }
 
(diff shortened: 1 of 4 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/release.groovy
```diff
@@ -23,6 +23,7 @@ void push_rpms_katello(version) {
 void push_debs_direct(os, repo) {
     sshagent(['freight']) {
         sh "ssh freight@web02.theforeman.org deploy ${os} ${repo}"
+        sh "ssh freight@web01.osuosl.theforeman.org deploy ${os} ${repo}"
```
ekohl, line 26: Does it make sense to do this in parallel?

evgeni:
Yeah. But I'd prefer one breakage at a time ;)

puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -13,13 +13,15 @@ DEB_PATH=./debian/${os}/build-${project}
...
-TARGET_PATH="${USER}@${HOST}.theforeman.org:rsync_cache/${os}/${COMPONENT}/"
+for HOST in HOSTS; do
+  # The path is important, as freight_rsync (which is run on the web node for incoming
+  # transfers) will parse the path to figure out the repo to send debs to.
+  TARGET_PATH="${USER}@${HOST}.theforeman.org:rsync_cache/${os}/${COMPONENT}/"
 
-if ls $DEB_PATH/*deb >/dev/null 2>&1; then
-  /usr/bin/rsync -avPx $DEB_PATH/*deb $TARGET_PATH
-fi
+  if ls $DEB_PATH/*deb >/dev/null 2>&1; then
```
ekohl, line 24: Perhaps this if should be outside of the for loop? Same applies to other scripts.

evgeni: technically yes, but I didn't want to also change the logic flow of the script, and the loop needs to be outside of `TARGET_PATH` as that contains `HOST`.

ekohl: We should really rewrite these to Jenkins pipelines. But I'm ok with merging it as is.

evgeni: Absolutely!

PR #1412, opened 2020-08-13
Retry getting a node from ci.centos

ehelms:


ehelms:
Based on seeing failures typically related to the get, this could be due to multiple requests happening in parallel. So let's add some retries.

puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
```diff
@@ -8,10 +8,13 @@
         action: get
         api_key: "{{ api_key }}"
       register: cico_get_data
+      retries: 5
+      delay: 5
+      until: cico_get_data.results.ssid
```
evgeni, line 13: So https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met is super sparse, what happens when the `until` is not true-ish when the retries are used up? Will it fail *this* task, or will we get the `fail` task in the next lines?

Doesn't matter, technically, just curious :)

ehelms: I think this task will fail. One way to find out!

PR #1414, opened 2020-08-18
make stagingdeb homedir world-readable

evgeni:
the main deb user needs to be able to read it.

secrets (ssh, gpg) are still secure as their folders have stricter perms

### DIFF
--- a/puppet/modules/web/manifests/vhost/deb.pp
+++ b/puppet/modules/web/manifests/vhost/deb.pp
@@ -20,6 +20,8 @@ class web::vhost::deb (
     # script to handle deployment too
     secure_ssh::receiver_setup { $user:
       user           => $user,
+      groups         => ['freightstage'],
+      homedir        => $home,
       foreman_search => 'host.hostgroup = Debian and (name = external_ip4 or name = external_ip6)',
       script_content => template('freight/rsync_main.erb'),
       ssh_key_name   => "rsync_${user}_key",
@@ -35,5 +37,7 @@ class web::vhost::deb (
       mode    => '0700',
       content => template('freight/deploy_debs.erb'),
     }
+
+    User <| title == 'freightstage' |> -> User <| title == $user |>
   }
 }
--- a/puppet/modules/web/manifests/vhost/stagingdeb.pp
+++ b/puppet/modules/web/manifests/vhost/stagingdeb.pp
@@ -18,6 +18,8 @@ class web::vhost::stagingdeb(
   if $setup_receiver {
     secure_ssh::rsync::receiver_setup { $user:
       user           => $user,
+      homedir        => $home,
+      homedir_mode   => '0750',
       foreman_search => 'host.hostgroup = Debian and (name = external_ip4 or name = external_ip6)',
       script_content => template('freight/rsync.erb'),
     }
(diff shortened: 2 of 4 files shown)


puppet/modules/web/manifests/vhost/deb.pp
```diff
@@ -20,6 +20,7 @@
     # script to handle deployment too
     secure_ssh::receiver_setup { $user:
       user           => $user,
+      homedir        => $home,
```
evgeni, line 23: this is just to be explicit, as we define $home here, and `secure_ssh` uses the same "logic" to calculate the homedir

puppet/modules/web/manifests/vhost/stagingdeb.pp
```diff
@@ -18,8 +18,11 @@
   if $setup_receiver {
     secure_ssh::rsync::receiver_setup { $user:
       user           => $user,
+      homedir        => $home,
       foreman_search => 'host.hostgroup = Debian and (name = external_ip4 or name = external_ip6)',
       script_content => template('freight/rsync.erb'),
     }
+
+    File <| title == $home |> { mode => '0755' }
```
evgeni, line 26: technically, this could be 0750 if we add the `freight` user into the `freightstage` group, but I was lazy.

evgeni: which I did now

evgeni:
and now it even works ;-)

puppet/modules/web/manifests/vhost/deb.pp
```diff
@@ -35,5 +36,8 @@
       mode    => '0700',
       content => template('freight/deploy_debs.erb'),
     }
+
+    User <| title == 'freightstage' |> -> User <| title == $user |>
+    User <| title == $user |> { groups +> ['freightstage'] }
```
ekohl, line 41: Rather than setting it this way, should `secure_ssh::receiver_setup` gain a parameter to set the groups on the user?

puppet/modules/web/manifests/vhost/stagingdeb.pp
```diff
@@ -18,8 +18,11 @@
   if $setup_receiver {
     secure_ssh::rsync::receiver_setup { $user:
       user           => $user,
+      homedir        => $home,
       foreman_search => 'host.hostgroup = Debian and (name = external_ip4 or name = external_ip6)',
       script_content => template('freight/rsync.erb'),
     }
+
+    File <| title == $home |> { mode => '0750' }
```
ekohl, line 26: Should `secure_ssh::receiver_setup` gain a `Stdlib::Filemode $homedir_mode = '0700'` parameter and then override it in this file?

evgeni: Right, it's our module anyways. Good idea.

evgeni:
@ekohl  implemented

puppet/modules/web/manifests/vhost/deb.pp
```diff
@@ -35,5 +37,7 @@
       mode    => '0700',
       content => template('freight/deploy_debs.erb'),
     }
+
+    User <| title == 'freightstage' |> -> User <| title == $user |>
```
ekohl, line 41: An alternative to let Puppet figure this out itself is to manage the group explicitly in the receiver_setup. If you choose to go that route, you should set `gid` on the user resource as well to be explicit.

evgeni:
I prefer 700, as some systems default to jot creating a primary group for users and put all into "users", which would make 750 not as secure as possible

PR #1417, opened 2020-08-18
Add Katello 3.17

ehelms:


puppet/modules/jenkins_job_builder/files/centos.org/jobs/katello-pipelines.yml
```diff
@@ -16,15 +16,15 @@
       - katello
 
 - project:
-    name: katello-release-3.16
+    name: katello-release-3.16-3.17
```
pcreech, line 19: Um.... Should 3.16 still be here?

ehelms: yep, 3.16 and 3.17 have the same job configuration, this is just a naming thing

pcreech: ah, k.  WFM

ekohl: Now that you dropped centos8, there actually isn't a difference between 3.14, 3.15, 3.16 and 3.17, right?

PR #1416, opened 2020-08-18
Add Jenkins docs

ehelms:
Replaces:

 * https://projects.theforeman.org/projects/foreman/wiki/Jenkins

This one is a bit rough still since it covers Jenkins and jobs. I'm going to split out anything job related to it's down document.

PR #1419, opened 2020-08-18
Add Table of Contents with links

ehelms:
To be added once the majority of new docs have been added to `docs/`


ehelms:
This can go in as soon as https://github.com/theforeman/foreman-infra/pull/1416 is in

PR #1421, opened 2020-08-18
Add Webserver docs

ehelms:


docs/webserver.md
```diff
@@ -0,0 +1,21 @@
...
+* downloads.theforeman.org
+* stagingdeb.theforeman.org
+* yum.theforeman.org
+
+/var/www is mounted on a separate 100GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
+
+It has the following customisations:
+
+* firewalld is configured with TCP ports 22, 80, 443 and 873 open - should be Puppetised
+* /home/freight* has go+x to permit the deb deploy script (running under the freight user) to read both freight and freightstage config files - should be rolled into secure_ssh or freight Puppet module
```
ekohl, line 17: @evgeni ran into this and addressed in https://github.com/theforeman/foreman-infra/pull/1414
```suggestion
```

docs/webserver.md
```diff
@@ -0,0 +1,21 @@
...
+* slave01's SSH key is added to permit yum uploads - should be moved to a separate secure_ssh user
```
ekohl, line 18: I think this is also no longer an issue

docs/webserver.md
```diff
@@ -0,0 +1,21 @@
+# Webserver
+
+The main web server hosts:
```
ekohl, line 3:
```suggestion
The main web server (web02.rackspace.theforeman.org) hosts:
```

docs/webserver.md
evgeni: This is fixed in puppet by adding the main user to the stage group. We might want to document that fact here.

docs/webserver.md
evgeni: On the new box it's a normal user, not root

docs/webserver.md
```diff
@@ -0,0 +1,21 @@
...
+
+* theforeman.org, www.theforeman.org
+* deb.theforeman.org
+* debugs.theforeman.org
+* downloads.theforeman.org
+* stagingdeb.theforeman.org
+* yum.theforeman.org
+
+/var/www is mounted on a separate 100GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
```
evgeni, line 12:
```suggestion
/var/www is mounted on a separate 140GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
```

docs/webserver.md
evgeni, line 3: as we only have one webserver, I don't see a reason to call that one "main"

evgeni: I think we should have a table at the beginning of the page, with all the technical details, and not call out the hostname here.

smth like:
| | web02.rackspace.theforeman.org |
| - | - |
| type | OpenStack VM |
| OS | CentOS 7 |
| CPUs | 2 |
| RAM | 2GB |
| Storage | /dev/xvda (40GB): root, /dev/xvdb (100GB) + /dev/xvdc (50GB): data |

ekohl: I've been wondering about that - we have that info in Foreman so is it really that relevant? I always like it when the reasons are documented rather than the result.

evgeni: While I agree the data is in Foreman, I kinda like some basics being available here so that it's available when Foreman is not.

ehelms:
Added a round of updates. @evgeni please call out anything I should add regrading the configuration of the webserver worth noting based on the changes you've been making.

Perhaps I should add a link at the top to the puppet module that controls configuration for the webserver? (which one is that?)

Any services running on the webserver worth calling out for folks to be aware of?

docs/webserver.md
```diff
@@ -0,0 +1,39 @@
...
+/var/www is mounted on a separate 140GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
+
+## Firewall
+
+firewalld is configured with TCP ports:
+
+* 22
+* 80
+* 443
+* 873 open
```
ekohl, line 33: The open here is a bit odd.

docs/webserver.md
```diff
@@ -0,0 +1,39 @@
...
+* stagingdeb.theforeman.org
+* yum.theforeman.org
+
+## Volumes
+
+/var/www is mounted on a separate 140GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
+
+## Firewall
+
+firewalld is configured with TCP ports:
```
ekohl, line 28: It's good to mention that this is not Puppetized

docs/webserver.md
```diff
@@ -0,0 +1,39 @@
...
+* 22
+* 80
+* 443
+* 873 open
+
+## Other
+
+* freight and freightstage users have private auto-signing GPG key imported
+
+In case of maintenance, a template page and config file snippet are under /var/www/503.  The config should be copied into each vhost.
```
ekohl, line 39: I'm not sure this still works with Fastly as a CDN in front.

docs/webserver.md
```diff
@@ -0,0 +1,39 @@
...
+firewalld is configured with TCP ports:
+
+* 22
+* 80
+* 443
+* 873 open
+
+## Other
+
+* freight and freightstage users have private auto-signing GPG key imported
```
ekohl, line 37: Also good to mention that this is not Puppetized

ekohl (commented):
https://github.com/theforeman/foreman-infra/pull/1333 does refactor things a bit, but https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/profiles/manifests/web.pp.

You should also mention that it serves rsync.theforeman.org. This is a separate domain because it's not fronted by the Fastly CDN, which yum, downloads and stagingdebs are.

docs/webserver.md
```diff
@@ -0,0 +1,41 @@
...
+## Domains
+
+These domains are all hosted on the webserver.
+
+* theforeman.org, www.theforeman.org
+* deb.theforeman.org
+* debugs.theforeman.org
+* downloads.theforeman.org
+* stagingdeb.theforeman.org
+* yum.theforeman.org
```
evgeni, line 21: these three are fronted by Fastly, and web02 is only the backend, wonder if we can somehow write that down nicely?

docs/webserver.md
```diff
@@ -0,0 +1,41 @@
...
+* rsync.theforeman.org
+
+## Volumes
+
+/var/www is mounted on a separate 140GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
+
+## Firewall
+
+firewalld is manually configured (non-puppetized) with TCP ports:
```
evgeni, line 30: There is no firewall *on* the new host (but OpenStack has one)

docs/webserver.md
evgeni: If the backend serves 50x, Fastly will serve old content.

docs/webserver.md
ehelms: Do you want this to reflect the new webserver before being merged or merge as is for the old one and do a round of updates?

evgeni: I think workflow wise it's better to change that when the new machine becomes prod, as that's what we're documenting here after all.

docs/webserver.md
```diff
@@ -0,0 +1,51 @@
...
+* 22
+* 80
+* 443
+* 873
+
+## Other
+
+* freight and freightstage users have private auto-signing GPG key imported manually (non-puppetized)
+
+In case of maintenance, a template page and config file snippet are under /var/www/503.  The config should be copied into each vhost.
```
evgeni, line 51: this is oddly formatted, should it be another bullet point in the `Other` list?

docs/webserver.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Fastly CDN
+
+A Fastly CDN exists that sits in front of:
+
+* downloads.theforeman.org
+* stagingdeb.theforeman.org
+* yum.theforeman.org
+
+For these, the webserver acts as a backend while the content is served from the Fastly CDN to users.
```
evgeni, line 32: There is also https://projects.theforeman.org/projects/foreman/wiki/Infrastructure_CDN which is a tad outdated and there is a bit of ansible to deploy that setup in `foreman-infra/ansible/` (mostly undocumented)

I wonder if we should fold that completely into webserver.md (I can do as soon as we merge this PR)?

evgeni:
IMHO this is a good start and (the one formatting issue I noted) can be merged.

evgeni (approved):
I've opened an issue for the cdn bits

PR #1422, opened 2020-08-19
add pulpcore mash to script

evgeni:


koji/collection-mash-split.py
```diff
@@ -470,6 +470,14 @@ def main():
                 el8_candlepin.name = 'katello-candlepin-{}-el8'.format(version)
                 mashes.append(el8_candlepin)
 
+    elif collection == 'pulpcore':
+        CONFIG["gitloc"] = "https://github.com/theforeman/pulpcore-packaging.git"
+        git_tag = "rpm/{}".format(version)
```
ekohl, line 475: This suggests there may be deb packaging in the future. Or is this only to mirror foreman-packaging?

evgeni: To mirror f-packaging. The alternative would be to call the branches `PULPCORE-X.Y` or something. I don't mind either way.

evgeni: Mirroring has the benefit that we know the Jenkins jobs can parse that ;)

ekohl: I'm ok with that. Do you intend to have a master branch similar to foreman-packaging which just serves a README about the layout (and is the default branch) or will you keep updating the default branch over time to the latest version?

evgeni: Good question I didn't think about yet.

Keeping the latest branch as default sounds good.

ekohl (approved):
Please capture that policy somewhere. Either in the pulpcore-packaging's README.md or in some procedure in tool_belt come to mind.

evgeni:
> Please capture that policy somewhere. Either in the pulpcore-packaging's README.md or in some procedure in tool_belt come to mind.

ack, noted down a todo item to do so.

PR #1423, opened 2020-08-19
add pulpcore-packaging test and release jobs

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pulpcore_packaging_release.groovy
```diff
@@ -0,0 +1,74 @@
+def packages_to_build
```
evgeni, line 1: this file should not be here but folded with `foreman_packaging_release.groovy`, but I was too lazy to parametrize it properly

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/pulpcore-packaging-rpm-pr-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+    name: pulpcore-packaging-rpm-pr-test
+    project-type: pipeline
+    concurrent: true
+    sandbox: true
+    properties:
+      - github:
+          url: https://github.com/theforeman/pulpcore-packaging
+    triggers:
+      - github_pr_rpm:
+          context: 'rpm'
```
evgeni, line 12: wonder if this should be `pulpcore` instead?

ekohl: Since you have `rpm/x.y`, I think it's fine to keep it that way.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pulpcore_packaging_release.groovy
```diff
@@ -0,0 +1,74 @@
...
+                    userRemoteConfigs: [
+                        [url: 'https://github.com/theforeman/pulpcore-packaging']
+                    ]
+                ])
+
+            }
+        }
+        stage('Find packages') {
+            steps {
+                copyArtifacts(projectName: env.JOB_NAME, optional: true)
```
ehelms, line 29: What is getting copied in Pulpcore case?

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging/pulpcore-packaging-rpm-pr-test.yaml
ehelms: Agreed, given the choice of `rpm/` pattern for branches, this follows on.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pulpcore_packaging_release.groovy
evgeni: That's your code to copy the "commit" file from the last green job.

ehelms: Touche.

PR #1428, opened 2020-08-20
Retry provisioning of a ci.centos.org node

ehelms:
Based on:

```
+ ansible-playbook -i localhost provision.yml
 [WARNING]: Unable to parse /home/foreman/workspace/foreman-pipeline-
katello-3.17-centos7-install/foreman-
infra/puppet/modules/jenkins_job_builder/files/centos.org/ansible/localhost as
an inventory source
 [WARNING]: No inventory was parsed, only implicit localhost is available
 [WARNING]: Could not match supplied host pattern, ignoring: all
 [WARNING]: provided hosts list is empty, only localhost is available

PLAY [localhost] ***************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [Get a new node] **********************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'cico_get_data.results.ssid' failed. The error was: error while evaluating conditional (cico_get_data.results.ssid): 'dict object' has no attribute 'results'"}

PLAY RECAP *********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1   
```

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
+++ b/puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
@@ -7,10 +7,8 @@
       cico:
         action: get
         api_key: "{{ api_key }}"
+        retry_count: 5
       register: cico_get_data
-      retries: 5
-      delay: 5
-      until: cico_get_data.results.ssid
 
     - name: 'Fail if no SSID returned'
       fail:


puppet/modules/jenkins_job_builder/files/centos.org/ansible/provision.yml
```diff
@@ -10,7 +10,7 @@
       register: cico_get_data
       retries: 5
       delay: 5
-      until: cico_get_data.results.ssid
+      until: cico_get_data.results and cico_get_data.results.ssid
```
evgeni, line 13: So I tried this, and it needs to be `cico_get_data.results is defined and cico_get_data.results.ssid` to work. But this also will only work on Ansible < 2.8, as since 2.8 `results` is a reserved keyword and the element is getting renamed to `ansible_module_results` with the following warning:

```
[WARNING]: Found internal 'results' key in module return, renamed to 'ansible_module_results'.
```

evgeni: Opened a PR against cicoclient for the later: https://github.com/CentOS/python-cicoclient/pull/24

evgeni: While hacking the module, I noticed it has a `retry_count` parameter, that defaults to 1, maybe setting this instead of making Ansible do the work for us is nicer?

ehelms:
Like this @evgeni ?

evgeni:
> Like this @evgeni ?

Yeah, that looks like what I thought.

If it works is a different story ;)

evgeni (approved):
The PR title doesn't match anymore, hut ACK on the change

ekohl (approved):

PR #1431, opened 2020-08-25
collect debug data from the proxy too

evgeni:


evgeni:
yes, this works for foreman, where we don't test proxies, too, as this is just a limit for an ansible playbook. the proxy will just not match there.

ekohl (approved):

PR #1441, opened 2020-09-08
allow running ansible playbooks in a virtualenv

evgeni:
We had implemented that once already, but then reverted because it
behaved weirdly. I can't reproduce the bad behaviour anymore, so let's
try this again.

Revert "Revert "allow running ansible playbooks in a virtualenv""

This reverts commit ad573df4439ebb33f1c91e29b51815297accb338.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
@@ -32,7 +32,7 @@ def runPlaybook(args) {
         command.push("-e @${sensitive_extra_vars_file}")
     }
 
-    command_string = "${command.join(' ')}"
+    command_string = command.join(' ')
 
     wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
         if (venv) {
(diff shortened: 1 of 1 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -31,8 +32,14 @@ def runPlaybook(args) {
...
     }
 
+    command_string = "${command.join(' ')}"
+
     wrap([$class: 'AnsiColorBuildWrapper', colorMapName: "xterm"]) {
-        sh "${command.join(' ')}"
+        if (venv) {
+            virtEnv(venv, command_string)
+        } else {
+            sh "${command_string}"
```
ekohl, line 41: Does this work:
```suggestion
            sh command_string
```

evgeni: probably

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/ansible.groovy
```diff
@@ -31,8 +32,14 @@ def runPlaybook(args) {
         command.push("-e @${sensitive_extra_vars_file}")
     }
 
+    command_string = "${command.join(' ')}"
```
evgeni, line 35:
```suggestion
    command_string = command.join(' ')
```

this also ;)

ekohl (approved):

ekohl:
Do you want to squash?

evgeni:
:tennis:  squashed

PR #1444, opened 2020-09-09
fix scoping of the stringifyed command

evgeni:
otherwise, when running multiple `runPlaybook` with `parallel`, the first
invocation wins and you end up with the same command executed multuple
times instead of the individual ones.


evgeni:
Remember how I said I could not reproduce the original issue why we reverted the `ansible-venv` feature the first time? Well, I wasn't testing well enough. But this is the bug we faced and this is the fix we need.

ehelms (approved):

PR #1445, opened 2020-09-09
don't add foreman changes to katello source release changes

evgeni:
otherwise the built RPM might have the wrong hash in the version


ekohl (approved):
I think there are multiple places where this makes sense. Wasn't aware this was an option.

PR #1446, opened 2020-09-09
fix ansible syntax, the filter is called length

evgeni:


evgeni:
fix for https://github.com/theforeman/foreman-infra/pull/1442

pcreech (approved):
ACK

PR #1447, opened 2020-09-10
verify version/release for all pulpcore branches

evgeni:
pulpcore has no rpm/develop that we pick from.


ekohl (approved):
Untested, but it looks good. `ghprbGhRepository` does appear to be the variable for the repo name and I don't spot any typos.

PR #1449, opened 2020-09-14
download all artifacts from ci.centos.org pipelines

evgeni:


evgeni:
this should "fix" the quick rotation of console logs and other artifacts on ci.centos.org

ekohl:
Is consoleText the plain text version?

evgeni:
Yes it is

PR #1451, opened 2020-09-14
deploy downloads as dedicated user

evgeni:


puppet/modules/web/files/rsync_downloads.sh
```diff
@@ -0,0 +1,36 @@
...
+rsync\ --server*)
+  # Permit transfer
+  $SSH_ORIGINAL_COMMAND
+;;
+*)
+echo "Rejected"
+;;
+esac
+# ERB highlighting looks terrible in this script...
+# vim: set ft=sh : #
```
ekohl, line 36: Is this really still needed? Now that it's in files, it doesn't have an ERB extension.

puppet/modules/web/manifests/vhost/downloads.pp
```diff
@@ -36,8 +48,8 @@
 
   file { "${downloads_directory}/HEADER.html":
     ensure => file,
-    owner  => 'root',
-    group  => 'root',
+    owner  => $user,
+    group  => $user,
```
ekohl, line 52: I think it's still fine to keep this owned by root. This signals that it's managed by Puppet, rather than the user.

ekohl (commented):
Nice!

puppet/modules/web/files/rsync_downloads.sh
evgeni: yeah, copy paste error

puppet/modules/web/manifests/vhost/downloads.pp
evgeni: agreed

evgeni:
this will currently break both downloads and discovery deployment, as it blocks non-rsync things which those use.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
```diff
@@ -87,8 +87,8 @@ void build_tarball(project, version, ruby_ver) {
         withRVM(["${rake} pkg:generate_source"], ruby_ver, project)
 
         sshagent(['deploy-downloads']) {
-            sh "ssh root@web02.rackspace.theforeman.org \"mkdir -p ${base_dir}/\" || true"
-            sh "rsync -v --ignore-existing pkg/* root@web02.rackspace.theforeman.org:${base_dir}/"
+            sh "ssh downloads@web02.rackspace.theforeman.org \"mkdir -p ${base_dir}/\" || true"
```
evgeni, line 90: what is this `mkdir` for? `rsync` will happily create it for us if it doesn't exist.

ekohl: Perhaps in the past it didn't? I'd be fine with removing it and seeing when it breaks. We'll probably see that in 2.3 branching when we get to it.

puppet/modules/web/files/rsync_downloads.sh
```diff
@@ -0,0 +1,41 @@
...
+;;
+*\|*)
+echo "Rejected"
+;;
+rsync\ --server*)
+  # Permit transfer
+  $SSH_ORIGINAL_COMMAND
+;;
+update-discovery-latest-release)
+  DISCOVERY_RELEASES="/var/www/vhosts/downloads/htdocs/discovery/releases"
```
ekohl, line 32: I'm slightly leaning to templating and using `$downloads_directory` here.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/tarballsRelease.groovy
ekohl: Perhaps in the past it didn't? I'd be fine with removing it and seeing when it breaks. We'll probably see that in 2.3 branching when we get to it.

puppet/modules/web/templates/rsync_downloads.sh.erb
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+set -e
+
+case "$SSH_ORIGINAL_COMMAND" in
+*\&*)
+echo "Rejected"
```
ekohl, line 7: Not sure I looked at this earlier, but could you indent the `echo` lines?

evgeni: dude, I just copied that from some other file, but sure :)

ekohl: Are you suggesting that not all files are in perfect condition?

ekohl (approved):
Let's try this.

puppet/modules/web/templates/rsync_downloads.sh.erb
evgeni: I'd never suggest such a blasphemy!

PR #1452, opened 2020-09-15
drop 1.24 and its special cases

evgeni:


evgeni:
@tbrisker 1.24 is dead, right?

tbrisker (approved):

evgeni:
so good we have tests XD

PR #1453, opened 2020-09-15
only mash pulpcore for non-nightly and el8 for nightly

evgeni:


koji/collection-mash-split.py
```diff
@@ -443,23 +443,19 @@ def main():
...
+        if version in ['3.15', '3.16', '3.17']:
             pulpcore = MashConfig(collection, version, "pulpcore-el7", "pulpcore-el7", "pulpcore/el7")
             # The default builds katello-{}-pulpcore-el7
             pulpcore.name = 'katello-pulpcore-{}-el7'.format(version)
             mashes.append(pulpcore)
 
-            if version not in ['3.15', '3.16', '3.17']:
-                el8_katello = MashConfig(collection, version, "el8", "el8", "katello/el8")
-                mashes.append(el8_katello)
+        if version in ['nightly']:
```
ekohl, line 452: Do we want to make it future proof and assume all versions after 3.17 will mash on EL8?
```suggestion
        else:
```

evgeni: I'd rather not use `else` as that section has logically nothing in common with the `if` before. The `if` in 446 is "do we have an pulpcore subrepo" and here it's "do we also build el8".

ekohl: I kind of considered the same, but on the other hand, it is based on version. The alternative to make it future proof is
```suggestion
        if version not in ['3.15', '3.16', '3.17']:
```

evgeni: :+1:

ekohl (approved):

PR #1454, opened 2020-09-15
use relative path when archiving artifacts

evgeni:
otherwise archiveArtifacts rejects the path

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy
@@ -56,7 +56,8 @@ def runIndividualCicoJob(job_name, number = 0, job_parameters = null, job_extra_
     } finally {
         script {
             if (extra_vars['jenkins_artifacts_directory']) {
-                archiveArtifacts artifacts: "${extra_vars['jenkins_artifacts_directory']}/**", allowEmptyArchive: true
+                def relative_artifacts_directory = extra_vars['jenkins_artifacts_directory'].replace("${env.WORKSPACE}/", '')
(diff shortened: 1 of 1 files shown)


ekohl (commented):
The default is relative:

https://github.com/theforeman/foreman-infra/blob/5b3037739b6f46f3847711c17a6bf2cffaf1a69f/puppet/modules/jenkins_job_builder/files/centos.org/ansible/jenkins_job.yml#L11

Here it's set to an absolute path:

https://github.com/theforeman/foreman-infra/blob/5b3037739b6f46f3847711c17a6bf2cffaf1a69f/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/foreman-discovery-image.groovy#L21-L25

However, nowhere else it's actually set. Isn't that the real problem?

evgeni:
It's set here:
https://github.com/theforeman/foreman-infra/blob/35c62a87cd9f1a78ea8ac95a5452e8104a590528/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/foreman_infra.groovy#L108-L111

And if you look at the error was:
```
09:18:44  Archiving artifacts
09:18:44  ‘/home/jenkins/workspace/foreman-2.2-release-pipeline/artifacts/foreman-pipeline-foreman-2.2-ubuntu1804-upgrade//**’ doesn’t match anything, but ‘artifacts/foreman-pipeline-foreman-2.2-ubuntu1804-upgrade//**’ does. Perhaps that’s what you mean?
09:18:44  No artifacts found that match the file pattern "/home/jenkins/workspace/foreman-2.2-release-pipeline/artifacts/foreman-pipeline-foreman-2.2-ubuntu1804-upgrade//**". Configuration error?
```

evgeni:
I could try not defining the path as absolute to begin with?

ekohl (approved):
Let's try this.

PR #1455, opened 2020-09-15
remove all user debs after 7 days

evgeni:
the old logic always kept "one version" of every package, which can
result in broken PR builds if a user does the following:
1. build package_a with a bug and never builds a fixed version
2. tries to build package_b that depends on package_a


evgeni:
This PR is presented to you by @adamruzicka having an broken `ruby-foreman-tasks` in his plugins folder, and then trying to build `ruby-foreman-ansible` that depends on tasks. :woman_shrugging:

ekohl (approved):

PR #1456, opened 2020-09-18
Revert "remove all user debs after 7 days"

evgeni:
This reverts commit 5b3037739b6f46f3847711c17a6bf2cffaf1a69f.

This wiped more nightly bits than I wanted.


evgeni:
Merging, as that causes data loss right now.

PR #1457, opened 2020-09-22
use new deploy code also for pulpcore and client

evgeni:


evgeni:
of course #1333 was not complete XD

ekohl (approved):

PR #1458, opened 2020-09-22
fix awk call in deploy-yumrepo

evgeni:


evgeni:
and of course I didn't test `purgecdn` part

evgeni:
you can't use single quotes inside an `awk` expression because why would you

ekohl (approved):
So this never worked?

evgeni:
the old capistrano code worked, because it had double quotes, your translation didn't as you used single quotes.

evgeni:
I just did comment that part in the web01 deployment, as I didn't want to purge twice

PR #1460, opened 2020-09-28
flip fastly to OSUOSL

evgeni:


evgeni:
this has been deployed, so it's only here for documenting the status

ekohl (approved):

PR #1461, opened 2020-09-28
flip jenkins to OSUOSL

evgeni:


evgeni:
going to merge this so that no deploys happen on the old host anymore

PR #1462, opened 2020-09-28
update webserver docs after migration

evgeni:


docs/webserver.md
```diff
@@ -56,16 +56,16 @@ Alternatively one can use `p2.shared.global.fastly.net` for an IPv4-only setup.
 
 ## Volumes
 
-/var/www is mounted on a separate 140GB block device via LVM.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
+/var/www is mounted on a separate 140GB block device.  /var/www/freight* contain the staging areas for freight (deb), and /var/www/vhosts contain the web roots themselves.
```
ekohl, line 59: Why didn't we make it LVM? Won't it be hard to extend it in the future now?

ekohl (approved):
:+1: since it's documenting the current state but I have my doubts about using `/dev/sdb` raw like this.

docs/webserver.md
evgeni: You can grow the disk in OpenStack and then just issue a `resize2fs /dev/sdb` and done :)

PR #1466, opened 2020-10-01
add deb.tfm.o to fastly config

evgeni:


evgeni:
deb.tfm and stagingdeb.tfm share the same apache config in terms of caching, so anything in `/pool` should be cached for 30 days, and anything in `/dists` for 2 minutes, that means: the actual files: 30 days, package indexes: 2 minutes.
https://github.com/theforeman/foreman-infra/blob/54c6416090e81aac0ae0d93951377ecbce9a5142/puppet/modules/freight/manifests/user.pp#L44-L57

this is in sync with what we have for RPM:
https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/manifests/vhost/yum.pp#L12-L29

ekohl (approved):

PR #1472, opened 2020-10-05
Add string extraction and update for Foreman

upadhyeammit:
This add the functionality to automate string extractions and pulling new translations into Foreman from Transifex. The job is suppose to run every 7th day of month at 10 AM Asia/Kolkata timezone.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
@@ -1,0 +1,87 @@
+pipeline {
+  agent { label 'admin && sshkey' }
+  environment {
+  ruby_version = '2.5'
+  triggers {
+  	 TZ='Asia/Kolkata'
+  	 cron('0 10 */7 * *')
+  }
+  }
+  stages {
+    stage('Setup Environment') {
+      steps {
+        git url: 'https://github.com/theforeman/foreman', branch: 'develop'
+        script {
+          sh 'git checkout -b i18n-updates'
+          configureRVM(ruby_version)
+          withRVM(['bundle install --jobs=5 --retry=5'], ruby_version)
+          sh 'cp config/database.yml.example config/database.yml'
+        }
+      }
+    }
+    stage('Extract strings and update translations'){
+      steps {
+        script {
+          withCredentials([string(credentialsId: 'transifex-api-token', variable: 'TX_TOKEN')]) {
+            sh 'make -C locale/ tx-update'
+          }
+        }
+      }
+    }
+    stage('Test new strings'){
+      steps {
+        script {
+          sh 'make -C locale/ all-mo'
+        }
+      }
+    }
+    stage('Create PR with updated strings') {
+      steps {
+        script {
+          sh 'git commit --amend i18n - extracting new, pulling from tx(jenkins-auto-update)'
+          sh 'git branch -D i18n-updates'
+        }
+      }
+    }
+  }
+}
+
+def gemset(name = null) {
+
+    def base_name = env.BUILD_TAG
+
+    if (EXECUTOR_NUMBER != '0') {
+        base_name += '-' + EXECUTOR_NUMBER
+    }
+
+    if (name) {
+        base_name += '-' + name.replace(".", "-")
+    }
+
+    base_name
+}
+
+def configureRVM(ruby = '2.0', name = '') {
+    emptyGemset(ruby, name)
+    withRVM(["gem install bundler -v '< 2.0'"], ruby, name)
+}
+
+def emptyGemset(ruby = '2.0', name = '') {
+    withRVM(["rvm gemset empty ${gemset(name)} --force"], ruby, name)
+}
+
+def cleanupRVM(ruby = '2.0', name = '') {
+    withRVM(["rvm gemset delete ${gemset(name)} --force"], ruby, name)
+}
+
+def withRVM(commands, ruby = '2.0', name = '') {
+
+    commands = commands.join("\n")
+    echo commands.toString()
+
+    sh """#!/bin/bash -l
+        set +e
+        rvm use ruby-${ruby}@${gemset(name)} --create
+        ${commands}
+    """
+}
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
@@ -1,18 +1,18 @@
 pipeline {
   agent { label 'admin && sshkey' }
   environment {
-  ruby_version = '2.5'
-  triggers {
-  	 TZ='Asia/Kolkata'
-  	 cron('0 10 */7 * *')
+    ruby_version = '2.5'
   }
+  triggers {
+    cron('30 4 */7 * *')
   }
   stages {
     stage('Setup Environment') {
       steps {
         git url: 'https://github.com/theforeman/foreman', branch: 'develop'
         script {
-          sh 'git checkout -b i18n-updates'
+          sh '(git show-branch i18n-updates &>/dev/null) && (git checkout i18n-updates) || (git checkout -b i18n-updates)'
+          sh 'git rebase origin/develop'
           configureRVM(ruby_version)
           withRVM(['bundle install --jobs=5 --retry=5'], ruby_version)
           sh 'cp config/database.yml.example config/database.yml'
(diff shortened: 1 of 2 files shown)


upadhyeammit:
Few questions on this implementation:
1. I want to delete the git branch i18n-updates on unsucessful job run. This would allow to run next one without any problems.
2. I am not sure if git push will work out of the box as it requires credentials or ssh key.
3. Want to have same job for Katello and all plugins under theforeman org. So is it wise to have separate job for Katello and plugins ? (3 of them)
4. When creating pipeline for plugins, should I have n no of steps for n no of plugins ? or it should be handled in some other way ? How I can update list of plugins dynamically in job, if possible ?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,87 @@
...
+          withRVM(['bundle install --jobs=5 --retry=5'], ruby_version)
+          sh 'cp config/database.yml.example config/database.yml'
+        }
+      }
+    }
+    stage('Extract strings and update translations'){
+      steps {
+        script {
+          withCredentials([string(credentialsId: 'transifex-api-token', variable: 'TX_TOKEN')]) {
+            sh 'make -C locale/ tx-update'
```
ehelms, line 26: Do the make commands use the Ruby environment?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,87 @@
+pipeline {
+  agent { label 'admin && sshkey' }
+  environment {
+  ruby_version = '2.5'
+  triggers {
+  	 TZ='Asia/Kolkata'
+  	 cron('0 10 */7 * *')
+  }
```
ehelms, line 8: Indent within the environment block

ehelms:
You will need a JJB job definition for this to create a job, e.g. https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/katello-nightly-rpm-pipeline.yaml

ehelms:
> * I want to delete the git branch i18n-updates on unsucessful job run. This would allow to run next one without any problems.

You could keep the branch around and just force push to it every time the job runs. That has the benefit of, if a PR was left open, it will just update the PR.

> * I am not sure if git push will work out of the box as it requires credentials or ssh key.

Correct. A bot user with credentials setup in Jenkins will need to be used to orchestrate this. We have one today that is used for some actions like prprocessor I believe.

> * Want to have same job for Katello and all plugins under theforeman org. So is it wise to have separate job for Katello and plugins ? (3 of them)

Different jobs, that way you can easily identify failures or last ran.

ekohl:
> > * I want to delete the git branch i18n-updates on unsucessful job run. This would allow to run next one without any problems.
> 
> You could keep the branch around and just force push to it every time the job runs. That has the benefit of, if a PR was left open, it will just update the PR.

Github has an option in your account to delete merged branches automatically.

> > * I am not sure if git push will work out of the box as it requires credentials or ssh key.
> 
> Correct. A bot user with credentials setup in Jenkins will need to be used to orchestrate this. We have one today that is used for some actions like prprocessor I believe.

Yes, we do have an account. We need to be careful with credentials though. Right now we don't have a good way to only share credentials with certain jobs AFAIK. While with JJB it's not a problem, it may be a problem in the future (https://community.theforeman.org/t/jenkins-github-app/20244).

Given anyone can create a PR, I'm wondering if we should start off with a separate account. The bot account has write permissions on various repos and if those credentials leak, it'll be a bigger problem than some anonymous account.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+    commands = commands.join("\n")
+    echo commands.toString()
+
+    sh """#!/bin/bash -l
+        set +e
+        rvm use ruby-${ruby}@${gemset(name)} --create
+        ${commands}
+    """
+}
```
ekohl, line 87: We have this in a shared library that you can include

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
upadhyeammit: changed this to run in RVM env

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
upadhyeammit: done

upadhyeammit:
> You will need a JJB job definition for this to create a job, e.g. https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/katello-nightly-rpm-pipeline.yaml

created job definition for this one

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
upadhyeammit: i have included methods in JJB, that should work ?

      !include-raw:
        - pipelines/deploy/i18n-string-extraction-and-update.groovy
        - pipelines/lib/rvm.groovy

upadhyeammit:
If I try this on Jenkins directly the it fails with,
~~~
ERROR: Could not find credentials entry with ID 'transifex-api-token'
Finished: FAILURE
~~~
I think id I am using is correct ?

upadhyeammit:
> > * I want to delete the git branch i18n-updates on unsucessful job run. This would allow to run next one without any problems.
> 
> You could keep the branch around and just force push to it every time the job runs. That has the benefit of, if a PR was left open, it will just update the PR.
> 
> > * I am not sure if git push will work out of the box as it requires credentials or ssh key.
> 
> Correct. A bot user with credentials setup in Jenkins will need to be used to orchestrate this. We have one today that is used for some actions like prprocessor I believe.
> 
> > * Want to have same job for Katello and all plugins under theforeman org. So is it wise to have separate job for Katello and plugins ? (3 of them)
> 
> Different jobs, that way you can easily identify failures or last ran.

Do you mean separate job for each plugin ? or like 3 jobs ?

upadhyeammit:
> > > * I want to delete the git branch i18n-updates on unsucessful job run. This would allow to run next one without any problems.
> > 
> > 
> > You could keep the branch around and just force push to it every time the job runs. That has the benefit of, if a PR was left open, it will just update the PR.
> 
> Github has an option in your account to delete merged branches automatically.
> 
> > > * I am not sure if git push will work out of the box as it requires credentials or ssh key.
> > 
> > 
> > Correct. A bot user with credentials setup in Jenkins will need to be used to orchestrate this. We have one today that is used for some actions like prprocessor I believe.
> 
> Yes, we do have an account. We need to be careful with credentials though. Right now we don't have a good way to only share credentials with certain jobs AFAIK. While with JJB it's not a problem, it may be a problem in the future (https://community.theforeman.org/t/jenkins-github-app/20244).
> 
> Given anyone can create a PR, I'm wondering if we should start off with a separate account. The bot account has write permissions on various repos and if those credentials leak, it'll be a bigger problem than some anonymous account.

Okay, so should we have separate account for i18n tasks ?

upadhyeammit:
@ehelms @ekohl changes done as per suggestions and added few questions, request to check.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,48 @@
...
+        script {
+          configureRVM(ruby_version)
+          withRVM(['make -C locale/ all-mo'], ruby_version)
+        }
+      }
+    }
+    stage('Create PR with updated strings') {
+      steps {
+        script {
+          sh "git commit --amend -m 'i18n - extracting new, pulling from tx(jenkins-auto-update)'"
```
ekohl, line 42: If (on line 14) it was checked out as a fresh branch, this will update the existing commit instead of creating a new one.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,48 @@
...
+          sh 'git push origin i18n-updates -f'
```
ekohl, line 43: I don't like giving Jenkins direct push permission to our upstream repository. You will at least need to do something with credentials since you don't want the key to always be present. That would make it trivial for any PR to push anything to any branch.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/i18n-string-extraction-and-update-pipeline.yaml
```diff
@@ -0,0 +1,10 @@
+- job:
+    name: i18n-string-extraction-and-update-pipeline
+    project-type: pipeline
+    sandbox: true
+    triggers:
+      - timed: '30 4 */7 * *'
```
ekohl, line 6: I don't think you need to specify it twice (both here and groovy). Not sure which one would take precedence or would be preferred.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/deploy/i18n-string-extraction-and-update.groovy
```diff
@@ -0,0 +1,48 @@
...
+        }
+      }
+    }
+  }
+}
```
ekohl, line 48: Please use newlines at the end of the file.

ekohl (changes requested):

ehelms:

PR #1476, opened 2020-10-26
Drop kafo and kafo_parsers masters tests

ehelms:


ekohl (commented):
We should also codify this in the kafo gemspec with a minimal required Ruby version. It would mean a major new version, of the gem but I think that's fair at this point. Then we can also remove all the RUBY_VERSION conditionals.

ehelms:
I am changing the tactic with this PR to just drop the test of the master branch of kafo and kafo_parsers. Given they do not see a lot of change, and that we do not have nightly releases of Kafo I think it's OK to reduce the number of jobs to PR jobs and let release PRs indicate the status when doing a release.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_parsers_master.yml
```diff
@@ -1,38 +0,0 @@
...
-            - 2.0.0
-            - 2.1
-            - 2.3
-      - axis:
-          type: user-defined
-          name: puppet
-          values:
-            - 4.10.0
-    builders:
-      - shell: !include-raw: scripts/test/test_kafo_parsers_master.sh
```
ekohl, line 32: This file is now unused.

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/test_kafo_master.yaml
```diff
@@ -1,42 +0,0 @@
...
-          - '4.10'
-          - '5.0'
-          - '6.0'
-    execution-strategy:
-      combination-filter: '!(ruby ==~ /^2\.[0-3].*/ && puppet ==~ /^[56]\..*/)'
-      touchstone:
-        expr: 'ruby=="2.5" && puppet=="6.0"'
-        result: 'stable'
-    builders:
-      - test_kafo
```
ekohl, line 37: This builder is now unused

ekohl (changes requested):

ekohl (approved):

PR #1478, opened 2020-10-29
Add ci.theforeman.org jobs to the update jobs job

ehelms:
The credentials will need to be created and added if the structure is agreed upon. After this PR, we can then:

 * move the creation of this job to the puppet module
 * move the bulk of the JJB configurations to a top level directory

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
@@ -17,6 +17,14 @@ pipeline {
             }
         }
 
+        stage('Update ci.theforeman.org jobs') {
+            steps {
+                withCredentials([string(credentialsId: 'theforeman-jenkins', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
+                    virtEnv('jjb-venv', "cd ./ci/theforeman.org && jenkins-jobs --conf ./foreman_jenkins.ini --user ${env.USERNAME} --password '${env.PASSWORD}' update --delete-old -r .")
+                }
+            }
+        }
+
         stage('Update ci.centos.org jobs') {
             steps {
                 withCredentials([string(credentialsId: 'centos-jenkins', variable: 'PASSWORD')]) {
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/foreman_jenkins.ini
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/foreman_jenkins.ini
@@ -1,0 +1,10 @@
+[job_builder]
+ignore_cache=True
+keep_descriptions=False
+include_path=.
+recursive=True
+allow_duplicates=False
+
+[jenkins]
+url=https://ci.theforeman.org/
+query_plugins_info=False


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -17,6 +17,14 @@ pipeline {
             }
         }
 
+        stage('Update ci.theforeman.org jobs') {
+            steps {
+                withCredentials([string(credentialsId: 'theforeman-jenkins', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
+                    virtEnv('jjb-venv', "cd ./ci/theforeman.org && jenkins-jobs --conf ./foreman_jenkins.ini --user ${env.USERNAME} --password '${env.PASSWORD}' update --delete-old -r ./jobs")
```
evgeni, line 23:
```suggestion
                    virtEnv('jjb-venv', "cd ./ci/theforeman.org && jenkins-jobs --conf ./foreman_jenkins.ini --user ${env.USERNAME} --password '${env.PASSWORD}' update --delete-old -r .")
```

the directory layout is not 100% identical between the tfm and the centos jenkins

evgeni (changes requested):
Another thing that I wonder about the following:

You added `--delete-old` here, but we also have https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/jenkins_job_builder/files/theforeman.org/unmanaged_jobs.rb that is run via cron 
https://github.com/theforeman/foreman-infra/blob/3d24ecf3695dec14a68458467e075fe651d63b89/puppet/modules/jenkins_job_builder/manifests/config.pp#L40-L46

and then we have another cron that runs `update --delete-old`:
https://github.com/theforeman/foreman-infra/blob/3d24ecf3695dec14a68458467e075fe651d63b89/puppet/modules/jenkins_job_builder/manifests/config.pp#L25-L31

and finally we have the old deployment that has no `--delete-old` and gets triggered by puppet whenever stuff changes:
https://github.com/theforeman/foreman-infra/blob/3d24ecf3695dec14a68458467e075fe651d63b89/puppet/modules/jenkins_job_builder/manifests/config.pp#L40-L46

My understanding is that your plan is to have Puppet only deploy the `foreman-infra-jenkins-jobs-update` job (by calling `jjb update foreman-infra-jenkins-jobs-update` or similar) and then let Jenkins do the rest.

Now my questions:
* Should `foreman-infra-jenkins-jobs-update` get a timed trigger and thus replace the cron job too?
* Do we really need the `unmanaged_jobs.rb` thing? Doesn't `--delete-old` take care of that?

ehelms:
I was thinking of a prepare and use the new for a while and then remove the old strategy. 

> Should foreman-infra-jenkins-jobs-update get a timed trigger and thus replace the cron job too?

For handling the periodic removal of jobs that are not tracked in git rather than relying on merges to this repository to go find them? That probably makes sense when we replace the cron job. I'd like to do that in two separate steps unless there is a good reason not to.

evgeni (changes requested):
the comment about the folder structure (https://github.com/theforeman/foreman-infra/pull/1478/files#r544066487) still stands and needs addressing

evgeni:
> For handling the periodic removal of jobs that are not tracked in git rather than relying on merges to this repository to go find them? That probably makes sense when we replace the cron job. I'd like to do that in two separate steps unless there is a good reason not to.

Absolutely, was just curious if you have a "roadmap" besides this PR :)

evgeni (approved):

PR #1479, opened 2020-10-29
Add rubocop step to kafo PR tests

ehelms:


ehelms:
Requires https://github.com/theforeman/kafo/pull/304 to be merged first

ekohl (changes requested):
In https://github.com/theforeman/foreman-infra/commit/3592a770d600fea73ae3b2b3ac6dadc47304f9fa I converted the Smart Proxy to use Rubocop as well. The most important part that it adds there is that it also publishes the results. I chose to only run it once before everything else. This has the benefit that you don't run the whole test suite if Rubocop fails. It was also done to avoid publishing the same thing multiple times.

ehelms:

PR #1483, opened 2020-10-29
Switch to GH actions for testing

ehelms:
This is my first time writing GH actions by hand and I have based this on other projects in our community thus this will likely need some iteration to get right. I have already pushed empty files to allow iterating on the results of the tests.

This is motivated by Travis becoming increasingly slow and GH actions providing more concurrency. We also should not use Jenkins here given this repository bootstrap's Jenkins.


.github/workflows/puppet.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-${{ matrix.ruby }}-gems-unit-${{ hashFiles('**/Gemfile.lock') }}
```
evgeni, line 23: we don't have a `Gemfile.lock` in the repo, so this cache will never hit. See https://github.com/theforeman/foreman-installer-modulesync/pull/106#issuecomment-718022711 for a similar discussion.

.github/workflows/puppet.yml
```diff
@@ -0,0 +1,47 @@
...
+            if [[ ! -f g10k-linux-amd64 ]] ; then
+              wget https://github.com/xorpaul/g10k/releases/download/v0.8.9/g10k-linux-amd64.zip
+            fi
+            unzip g10k-linux-amd64.zip
+          fi
+
+          ./g10k -quiet -cachedir .g10k/cache -puppetfile -puppetfilelocation puppet/Puppetfile_forge -moduledir puppet/forge_modules
+
+          cd puppet
+          bundle exec rake syntax
```
evgeni, line 45:
```
---> syntax:hiera:yaml
ERROR: Failed to parse vendor/bundle/ruby/2.4.0/gems/puppet-6.19.1/spec/fixtures/unit/data_providers/environments/hiera_bad_syntax_yaml/data/bad.yaml: (vendor/bundle/ruby/2.4.0/gems/puppet-6.19.1/spec/fixtures/unit/data_providers/environments/hiera_bad_syntax_yaml/data/bad.yaml): found unexpected ':' while scanning a plain scalar at line 3 column 1
```

Seems it tries to parse the installed files and chokes on an explicitly bad one?

.github/workflows/puppet.yml
```diff
@@ -0,0 +1,47 @@
+name: Unit tests
```
evgeni, line 1: it feels weird to have two files, both having `name: Unit tests`, and one `unit` job, but doing different things.

IMHO this should be:
1. one file called `main.yml` or `tests.yml` or whatever
2. having a useful name (or just "CI")
3. having two jobs defined: `puppet` and `jenkins`.

.github/workflows/puppet.yml
ehelms: Guess I need to find a way to ignore that path or to put that bundle path somewhere else

ehelms:
Switched to a `main.yml` with two sections. Applied the cache strategy @evgeni linked and fixed the errors by placing the bundle in `~/.vendor` as well as it checks out the submodules now for puppet.

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            DIR="."
+            ( cd $d && jenkins-jobs -l debug test -r $DIR )
+            ( cd $d && jenkins-jobs test -r $DIR --config-xml -o output && find output -name config.xml -exec python ../jenkins-lint.py --xml {} + && rm -rf output )
+          done
+
+          for d in puppet/modules/jenkins_job_builder/files/centos.org; do
+            DIR="jobs"
+            ( cd $d && jenkins-jobs -l debug test -r $DIR )
+            ( cd $d && jenkins-jobs test -r $DIR --config-xml -o output && find output -name config.xml -exec python ../jenkins-lint.py --xml {} + && rm -rf output )
+          done
```
evgeni, line 44: the `for d in <path>` blocks seem rather unnecessary as they only contain one item each, but then again the two could be easily folded together if we could set `DIR` more nicely.

shouldn't block this PR tho, as I think that part is copied from travis.

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.4"
+    name: Puppet
+    steps:
+      - uses: actions/checkout@v2
+      - name: Checkout submodules
+        uses: textbook/git-checkout-submodule-action@master
```
evgeni, line 56:
```suggestion
      - uses: actions/checkout@v2
        with:
          submodules: true
```

.github/workflows/main.yml
ehelms: Yes, taken straight from Travis. So I like re-factoring afterwards rather than before.

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y yajl-tools
+          gem install bundler
+          cd puppet
+          bundle config path ~/.vendor/bundle
+          bundle lock
```
ekohl, line 67: Locking wasn't needed in my experience.

ehelms: How did you create the lock file to be used in the cache step without this?

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Run tests
+        run: |
+          if [[ ! -f g10k ]] ; then
+            if [[ ! -f g10k-linux-amd64 ]] ; then
+              wget https://github.com/xorpaul/g10k/releases/download/v0.8.9/g10k-linux-amd64.zip
+            fi
+            unzip g10k-linux-amd64.zip
+          fi
+
+          ./g10k -quiet -cachedir .g10k/cache -puppetfile -puppetfilelocation puppet/Puppetfile_forge -moduledir puppet/forge_modules
```
ekohl, line 87: Might be good to do this as a separate step since it's not really part of running tests, really just installing modules. Also, might be nice to cache `.g10k/cache`

.github/workflows/main.yml
ekohl: You don't? https://github.com/theforeman/foreman-installer-modulesync/pull/106#issuecomment-720644253 explains why it's not needed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,92 @@
...
+            ( cd $d && jenkins-jobs -l debug test -r $DIR )
+            ( cd $d && jenkins-jobs test -r $DIR --config-xml -o output && find output -name config.xml -exec python ../jenkins-lint.py --xml {} + && rm -rf output )
+          done
+  puppet:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.4"
```
ekohl, line 51: Puppet 6 includes Ruby 2.5.
```suggestion
          - "2.5"
```

It's also confusing to read `Puppet (2.4)` in the checks.

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: '2.7'
+      - name: Restore pip cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
ekohl, line 23: We don't have a `requirements.txt`. I think that means the cache will never be rebuilt.

@evgeni would a solution to be to add a `pip freeze > requirements-lock.txt` at the end of the install phase?

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+      fail-fast: false
+    name: Puppet
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Setup ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.6'
```
ekohl, line 54: 2.6 is an odd version since there's no Puppet version that ships it. Probably doesn't matter too much, but using 2.5 ensures you don't rely on any Ruby 2.6 features.

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+
+          for d in puppet/modules/jenkins_job_builder/files/centos.org; do
+            DIR="jobs"
+            ( cd $d && jenkins-jobs -l debug test -r $DIR )
+            ( cd $d && jenkins-jobs test -r $DIR --config-xml -o output && find output -name config.xml -exec python ../jenkins-lint.py --xml {} + && rm -rf output )
+          done
+  puppet:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
ekohl, line 45: Fail fast is only relevant with a matrix build, but it's no longer a matrix.

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+          sudo apt-get update && sudo apt-get install -y yajl-tools
+          gem install bundler
+          cd puppet
+          bundle config path ~/.vendor/bundle
+      - uses: actions/cache@v2
+        with:
+          path: ~/.vendor/bundle
+          key: ${{ runner.os }}-${{ matrix.ruby }}-gems-unit-
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.ruby }}-gems-unit-
```
ekohl, line 66: There is no more matrix.

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
+name: CI
+
+on: pull_request
+
+jobs:
+  jjb:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
ekohl, line 9: This is only relevant for a matrix

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+          cd puppet
+          bundle install --jobs $(nproc) --retry 3
+      - name: Setup g10k
+        run: |
+          if [[ ! -f g10k ]] ; then
+            if [[ ! -f g10k-linux-amd64 ]] ; then
+              wget https://github.com/xorpaul/g10k/releases/download/v0.8.9/g10k-linux-amd64.zip
+            fi
+            unzip g10k-linux-amd64.zip
+          fi
```
ekohl, line 78: We don't store this in git so it's safe to simplify it:
```suggestion
          wget https://github.com/xorpaul/g10k/releases/download/v0.8.9/g10k-linux-amd64.zip
          unzip g10k-linux-amd64.zip
```

For the Travis scripts I made it idempotent because I ran them locally as well.

.github/workflows/main.yml
evgeni: Technically? Yeah.

But we also could just dump the three packages on the next step into a `requirements.txt` and use that? That would produce a rather static hash, as the non-versioned deps don't change much, but I think that's still OK?

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+on: pull_request
+
+jobs:
+  jjb:
+    runs-on: ubuntu-latest
+    name: Jenkins Job Builder
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v1
```
evgeni, line 12:
```suggestion
        uses: actions/setup-python@v2
```

can you add a dependabot config like https://github.com/theforeman/obal/blob/master/.github/dependabot.yml so that actions get bumped automatically?

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+          sudo apt-get update && sudo apt-get install -y yajl-tools
+          gem install bundler
+          cd puppet
+          bundle config path ~/.vendor/bundle
+      - uses: actions/cache@v2
+        with:
+          path: ~/.vendor/bundle
+          key: ${{ runner.os }}-gems-unit-
+          restore-keys: |
+            ${{ runner.os }}-gems-unit-
```
evgeni, line 62: this seems redundant, no `hashFiles()` at all here? then the cache will always be saved as `Linux-gems-unit-`

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+          submodules: true
+      - name: Setup ruby
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.5'
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y yajl-tools
+          gem install bundler
+          cd puppet
```
evgeni, line 55: you can set a `defaults.run.working-directory` to `puppet` so that you don't need to `cd` all the time:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun

.github/workflows/main.yml
```diff
@@ -0,0 +1,78 @@
...
+          wget https://github.com/xorpaul/g10k/releases/download/v0.8.9/g10k-linux-amd64.zip
+          unzip g10k-linux-amd64.zip
+
+          ./g10k -quiet -cachedir .g10k/cache -puppetfile -puppetfilelocation puppet/Puppetfile_forge -moduledir puppet/forge_modules
+      - name: Run tests
+        run: |
+          cd puppet
+          bundle exec rake syntax
+          bundle exec rake lint
+          bundle exec rake spec
```
evgeni, line 78: would it make sense to have these as separate steps, so that it's easier to see which step failed?

.github/workflows/main.yml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  jjb:
+    runs-on: ubuntu-latest
+    name: Jenkins Job Builder
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
```
ekohl, line 14: I think we should use Python 3.

.github/workflows/main.yml
```diff
@@ -0,0 +1,83 @@
...
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      - name: Install dependencies
+        run: |
+          echo 'jenkins-job-builder' >> requirements.txt
+          echo 'lxml' >> requirements.txt
+          echo 'requests' >> requirements.txt
```
ekohl, line 26: Can't we store `requirements.txt` in git as a file?

.github/workflows/main.yml
```diff
@@ -0,0 +1,83 @@
...
+      - uses: actions/checkout@v2
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+      - name: Restore pip cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
```
ekohl, line 19: Can you include the Python version in the cache key? That way the cache is automatically invalidated when we switch the Python version.

.github/workflows/main.yml
```diff
@@ -0,0 +1,83 @@
...
+          ruby-version: '2.5'
+      - name: Install dependencies
+        run: |
+          sudo apt-get update && sudo apt-get install -y yajl-tools
+          gem install bundler
+          bundle config path ~/.vendor/bundle
+      - uses: actions/cache@v2
+        with:
+          path: ~/.vendor/bundle
+          key: ${{ runner.os }}-gems-unit-
```
ekohl, line 63: This should include the Ruby version and hashfiles
```suggestion
          key: ${{ runner.os }}-gems-2.5-${{ hashFiles('**/Gemfile.lock') }}
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,83 @@
...
+          restore-keys: |
+            ${{ runner.os }}-gems-unit-
+      - name: Bundle install
+        run: |
+          bundle install --jobs $(nproc) --retry 3
```
ekohl, line 68: If it's a single command, there's no need for the multi-line syntax
```suggestion
        run: bundle install --jobs $(nproc) --retry 3
```

.github/workflows/main.yml
ehelms: If there is never a `Gemfile.lock` at this point in the steps, I do not get why I would include hashing it?

.github/workflows/main.yml
ehelms: Truth. I find it nice to read though.

.github/workflows/main.yml
ehelms: We could. @evgeni suggested this method, y'all figure it out and tell me what you want.

.github/workflows/main.yml
ekohl: I think that `bundle install` generates a `Gemfile.lock` and in the post step it will create a bundle with the correct hash.

ehelms: That makes sense. When it runs the next time, since there is no `Gemfile.lock` created yet at this step, it will use `${{ runner.os }}-gems-2.5-` as the lookup key right? So the hash of the `Gemfile.lock` seems to be pointless without creating the lockfile in the step above it.

.github/workflows/main.yml
evgeni: I did not, I meant to have the file in git. ;)

.github/workflows/main.yml
evgeni: The cached wheels are version agnostic, so why bother?

ekohl: I suspected lxml was version dependent. It's not?

evgeni: lxml is, indeed, but then the filename in the cache is different too and it'll be updated from the internet. no idea what's saving more, to be honest.

ekohl: I suspect that if you use a common cache, it can grow and grow. Not sure if there's anything that purges old wheels from the cache, but switching Python version could be a good enough thing to force that.

tbrisker:
What's the status here? i'd love to get rid of travis, it's holding up a release yet again...

ekohl (approved):
Going to merge this now to get rid of Travis. If it's incomplete, we can fix it later.

PR #1484, opened 2020-11-02
don't merge plugins with old stuff

evgeni:


ekohl (changes requested):
For the regular Foreman repos we have two different places. For releases, it's here:
https://github.com/theforeman/foreman-infra/blob/84b31986cb70235a66228e75a16dd737cc858521/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/foreman-pipeline.groovy#L48

For nightly it's here:
https://github.com/theforeman/foreman-infra/blob/84b31986cb70235a66228e75a16dd737cc858521/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/foreman-rpm.groovy#L50-L51

That explains why it's different there.

For plugins we always use the release model. A better solution is to set `merge = foreman_version != 'nightly'`.

ekohl (approved):
Might be nice to describe in the commit message why we do this.

ekohl (approved):
Thanks

PR #1487, opened 2020-11-02
Unify Foreman RPM release process

ekohl:
In most places the elX convention is used, but foreman repos are the exception in using RHEL/X. This change aligns all which uses. It doesn't contain any code to move the existing directories. While it's technically not needed and it'll work, it would waste a lot of disk space.

It then standardizes all pushes by using a common function to reduce duplication.

### DIFF
--- a/koji/collection-mash-split.py
+++ b/koji/collection-mash-split.py
@@ -398,7 +398,7 @@ def main():
         else:
             extras = []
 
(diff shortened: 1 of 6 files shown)


ehelms:

ekohl:
Replaced by https://github.com/theforeman/jenkins-jobs/pull/265

PR #1488, opened 2020-11-02
Foreman 2.3 mash updates

pcreech:
dropping el5 support

### DIFF
--- a/koji/collection-mash-split.py
+++ b/koji/collection-mash-split.py
@@ -407,10 +407,12 @@ def main():
             "el8": "el8",
             "rhel7": "el7",
             "rhel6": "el6",
-            "rhel5": "el5",
             "sles11": "sles11",
             "sles12": "sles12",
         }
+
+        if version in ["2.0", "2.1", "2.2"]:
+            dists["rhel5"] = "el5"
 
         if version == "nightly":
             git_tag = "rpm/develop"


pcreech:
Tested LooseVersion logic:
```
>>> from distutils import version
>>> version.LooseVersion("2.2")
LooseVersion ('2.2')
>>> version.LooseVersion("2.2") < version.LooseVersion("2.3")
True
>>> version.LooseVersion("2.3") < version.LooseVersion("2.3")
False
>>> version.LooseVersion("nightly")
LooseVersion ('nightly')
>>> version.LooseVersion("nightly") < version.LooseVersion("2.3")
False
>>> version.LooseVersion("nightly") < version.LooseVersion("2.3")
```

koji/collection-mash-split.py
```diff
@@ -407,11 +408,13 @@ def main():
             "el8": "el8",
             "rhel7": "el7",
             "rhel6": "el6",
-            "rhel5": "el5",
             "sles11": "sles11",
             "sles12": "sles12",
         }
 
+        if LooseVersion(version) < LooseVersion("2.3"):
```
evgeni, line 415: In the past we've been just doing `if version in ['2.0', '2.1', '2.2']:`

But using `LooseVersion` is also valid.

tbrisker (approved):

koji/collection-mash-split.py
pcreech: My specific thinking was, reading the code says "We are dropping el5 in 2.3", instead of "we are still shipping el5 in 2.0, 2.1, 2.2"

ekohl: Since I was the person who wrote those `if version in ...` parts I can tell you that it was easy to do it this way when I unified the mash-script-per-version into this single script. I don't have a particular preference, but we have a bunch of these comparisons. It'd be nice if the styles were the same.

pcreech: Personally, I think utilizing LooseVersion opens up some avenues for minimizing need to manage arrays of versions, so i'll propose doing one of these, and let me know which one you would like to move forward with:

a) switch all comparisions in this pr
b) leave this pr as is, open a new pr fixing the other comparisions
c) switch the above comparison to the "in array", and open a new pr switching all the comparisons.

ekohl: I'm leaning to option c.

pcreech: updated.

ekohl (approved):

ekohl:
Feel free to merge whenever. Travis doesn't matter since there's no CI on it. I just didn't merge it because it needs to be deployed manually.

pcreech:
updated /usr/local/bin/collection-mash-split.py with master koji/collection-mash-split.py

PR #1490, opened 2020-11-02
Switch to using LooseVersion version comparisons

pcreech:
There are various places in the code that have "in [array]" syntax.  This switches to using version comparison logic, to minimize the need to maintain version matricies in the conditionals, and replace them with simple greater than, less than, equal to logic.

### DIFF
--- a/koji/collection-mash-split.py
+++ b/koji/collection-mash-split.py
@@ -11,6 +11,7 @@ import difflib
 import time
 import re
 import sys
+from distutils.version import LooseVersion
 
 import koji
 import kobo.log
@@ -390,7 +391,7 @@ def main():
             git_tag = "rpm/{}".format(version)
 
         CONFIG["comps_path"] = "/mnt/koji/mash/comps/foreman"
-        if collection == 'foreman' and version in ['2.0', '2.1']:
+        if collection == 'foreman' and LooseVersion(version) <= LooseVersion('2.1'):
             CONFIG["extras_baseloc"] = "%(tag)s:extras"
             CONFIG["extras_path"] = "/mnt/koji/mash/extras/foreman"
             extras = ["extras-foreman-rhel7"]
@@ -399,7 +400,7 @@ def main():
 
         mashes = [MashConfig(collection, version, "rhel7-dist", "rhel7", "RHEL/7", extras)]
 
-        if version not in ['2.0']:
+        if LooseVersion(version) > LooseVersion('2.0'):
             mashes.append(MashConfig(collection, version, "el8", "el8", "RHEL/8"))
 
     elif collection == "foreman-client":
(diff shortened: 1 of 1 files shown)


pcreech:
Tested LooseVersion logic:
```
>>> from distutils import version
>>> version.LooseVersion("2.2")
LooseVersion ('2.2')
>>> version.LooseVersion("2.2") < version.LooseVersion("2.3")
True
>>> version.LooseVersion("2.3") < version.LooseVersion("2.3")
False
>>> version.LooseVersion("nightly")
LooseVersion ('nightly')
>>> version.LooseVersion("nightly") < version.LooseVersion("2.3")
False
>>> version.LooseVersion("nightly") < version.LooseVersion("2.3")
```

pcreech:
@evgeni @ekohl Here's the subsequent PR to switch to using LooseVersion.  Primarily i'm in favor of it because it reduces the amount of places in the code we have to maintain version matrices, and simplifies version comparison logic.

koji/collection-mash-split.py
```diff
@@ -399,7 +400,7 @@ def main():
 
         mashes = [MashConfig(collection, version, "rhel7-dist", "rhel7", "RHEL/7", extras)]
 
-        if version not in ['2.0']:
+        if LooseVersion(version) > LooseVersion('2.0'):
```
ekohl, line 403: Is this also true for `nightly`? Given 2.0 is now EOL, should we first submit a PR that cleans up 2.0?

ekohl (approved):
I think this is indeed easier to read.

koji/collection-mash-split.py
pcreech: I think that's kinda chicken/egg, as I'm working now on the cleanup PRs.  So i'll just take it into account, one way or another.

evgeni:
I'm fine with this. But we also should drop all branches that test for <= 2.0.

koji/collection-mash-split.py
ekohl: My logic is always to made the changes in an order that results in the smallest diffs. In this case that means first cleaning up all the 2.0 logic and then introducing `LooseVersion`. I adopted this because it makes tracking regressions easier and now apply it to all my patches.

koji/collection-mash-split.py
```diff
@@ -411,15 +412,15 @@ def main():
...
+        if LooseVersion(version) < LooseVersion("2.3"):
             dists["rhel5"] = "el5"
 
         if version == "nightly":
             git_tag = "rpm/develop"
         else:
             git_tag = "rpm/{}".format(version)
 
-            if version in ['2.0']:
+            if LooseVersion(version) == LooseVersion('2.0'):
```
tbrisker, line 423: Pretty sure we can drop the 2.0 stuff already

pcreech:
There is a broader "remove EOL" step in the branching docs.  I would prefer to handle the removal of 2.0 stuff in this file then if we can.  That way it's grouped in with those changes.

PR #1493, opened 2020-11-04
increase max_connections to 10 for rsync downloads

evgeni:
@alexjfisher reported having issues to rsync due to "max connections reached"


alexjfisher (approved):
That'll be a great help.  Thanks!

ekohl (approved):

PR #1494, opened 2020-11-04
Bump Katello timeouts to 3 hours

jturel:


tbrisker (approved):

ekohl:
FYI, I'm :-1: on this and would like to revert this. There were processes running for more than a week that took up resources. They probably slowed it down. If we keep extending limits, we don't find them.

PR #1496, opened 2020-11-04
[WIP] Add stages for containers to pipelines

xprazak2:
Adds additional stage that triggers image
builds when pipelines run.

Needs https://github.com/theforeman/grapple/pull/2/files, therefore wip.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
@@ -1,0 +1,26 @@
+import groovy.json.JsonOutput
+
+def createTimestamp() {
+    (new Date()).format("yyyyMMdd-HHmm", TimeZone.getTimeZone('UTC'))
+}
+
+def triggerGithubBuilder(project_name, version, version_build) {
+    def inputs = JsonOutput.toJson([
+        ref: 'master',
+        inputs: [
+            project_name: project_name,
+            version: version,
+            push: 'true',
+            registry: 'xyz',
+            version_build: version_build
+        ]
+    ])
+
+    println inputs
+
+    httpRequest httpMode: 'POST',
+                authentication: 'github_username_token',
+                customHeaders: [[name: 'Accept', value: 'application/vnd.github.v3+json']],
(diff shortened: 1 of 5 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
```diff
@@ -0,0 +1,26 @@
...
+}
+
+def triggerGithubBuilder(project_name, version, version_build) {
+    def inputs = JsonOutput.toJson([
+        ref: 'master',
+        inputs: [
+            project_name: project_name,
+            version: version,
+            push: 'true',
+            registry: 'xyz',
```
xprazak2, line 14: Change to a real registry name

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
```diff
@@ -0,0 +1,26 @@
...
+            version_build: version_build
+        ]
+    ])
+
+    def command_parts = [
+        "curl -X POST -H 'Accept: application/vnd.github.v3+json' -u $GITHUB_USER:$GITHUB_TOKEN",
```
ekohl, line 20: Careful here, it may get logged. There's a secrets module as well to store these things.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
```diff
@@ -0,0 +1,26 @@
...
+        "https://api.github.com/repos/xprazak2/image_builds/actions/workflows/build_image.yml/dispatches",
+        '-d \'' + inputs + '\''
```
ekohl, line 22: Perhaps we should use https://plugins.jenkins.io/http_request/ instead so there's no risk of an incorrectly escaped body?

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/lib/containers.groovy
xprazak2: I know, it uses secrets and they were not leaked when I tested: 

secrets

ehelms:

xprazak2:
Sure, I will open this against new repository.

PR #1498, opened 2020-11-05
Drop ARM builds

ekohl:
Version 2.0 was the last that supported these, but that's now EOL. This allows us to clean up the unsupported builders.

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_dependency.yaml
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_dependency.yaml
@@ -23,14 +23,13 @@
           name: arch
           values:
           - x86
-          - armv8
       - axis:
           type: slave
           name: label
           values:
           - debian
     execution-strategy:
-      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.2" || version == "2.1" || version == "2.0" ) && os == "focal") && !((version == "2.2" || version == "2.1" || version == "nightly") && arch == "armv8")'
+      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.2" || version == "2.1" || version == "2.0" ) && os == "focal")'
     builders:
       - build_deb_dependency
     publishers:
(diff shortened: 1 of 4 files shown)


puppet/modules/freight/templates/freight.conf.erb
```diff
@@ -17,5 +17,5 @@ GPG="563278F6"
 # Follow symlinks in VARLIB
 SYMLINKS="on"
 
-# x86 plus supported ARM platforms
-ARCHS="i386 amd64 armhf arm64"
+# x86
+ARCHS="i386 amd64"
```
ekohl, line 21: I debated this, not sure if this is the right one

tbrisker: I think we can also remove i386, iirc that was for 32 bit builds

puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/packaging_build_deb_dependency.yaml
```diff
@@ -23,14 +23,13 @@
...
           - x86
-          - armv8
       - axis:
           type: slave
           name: label
           values:
           - debian
     execution-strategy:
-      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.2" || version == "2.1" || version == "2.0" ) && os == "focal") && !((version == "2.2" || version == "2.1" || version == "nightly") && arch == "armv8")'
+      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.2" || version == "2.1" || version == "2.0" ) && os == "focal")'
```
tbrisker, line 32: I think this also need to change to add 2.3 since we don't have focal yet?

ekohl: This PR is only about dropping ARM.

puppet/modules/freight/templates/freight.conf.erb
ekohl: Not sure, but looking at https://deb.theforeman.org/pool/buster/nightly/r/ruby-journald-native/ we still build these. I'm a bit unsure about removing them. I also don't know if this removes the arm builds from all the existing repos.

@evgeni any idea?

evgeni: I would expect it does, so I'd leave it here.

ekohl:
Anything else that should be done for this?

tbrisker (approved):

PR #1501, opened 2020-11-05
merge pulpcore rpms

evgeni:
this allows for older versions of the plugins to remain in the repo


ekohl:
In https://github.com/theforeman/foreman-infra/pull/1487 I made some changes to `push_rpms` to allow keeping old files. Maybe worth including that already?

evgeni:
Could do, if you want to. Essentially picking https://github.com/theforeman/foreman-infra/pull/1487/commits/51f421e5e27fb4f4b40757af41fb40bb11a33f35?

ekohl:
Well, I think the `push_foreman_rpms` change depends on the previous commit so that should be left out.

evgeni:
@ekohl ah, now I understood. so like this?

ekohl (approved):
Yes exactly.

PR #1505, opened 2020-11-09
purge the contents of "latest" from the CDN when it changes

evgeni:


evgeni:
@tbrisker @ekohl this should fix the "we need to manually purge 'latest' from the CDN" after we made a release.

however, this does not take into account https://yum.theforeman.org/latest, as this never changes :(

puppet/modules/web/manifests/init.pp
```diff
@@ -62,6 +62,22 @@
     source => 'puppet:///modules/web/filter_apache_stats.sh',
   }
 
+  file { '/usr/local/bin/fastly-purge':
+    ensure => file,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
+    source => 'puppet:///modules/web/fastly-purge.sh',
```
ekohl, line 70: It's preferred to inline files nowadays since it's faster in most cases. With source there's an additional HTTP request while with content it's part of the catalog. The performance difference shifts to the HTTP request above a certain size, but these scripts are small.
```suggestion
    content => file('web/fastly-purge.sh'),
```

evgeni: TIL

ekohl (approved):

PR #1506, opened 2020-11-10
configure expire defaults for web vhost

evgeni:


puppet/modules/web/manifests/vhost/web.pp
```diff
@@ -33,9 +34,43 @@
     { 'rewrite_rule' => ["^/api/latest(.*) /api/${stable}\$1 [R,L]"] },
   ]
 
+  $directory_config = [
+    {
+      path            => $web_directory,
+      options         => ['Indexes', 'FollowSymLinks', 'MultiViews'],
+      expires_active  => 'on',
+      expires_default => 'access plus 1 day',
```
ekohl, line 42: Doesn't this mean that it may take up to a day to publish a new manual? Perhaps the website deploy script also needs a tool to purge the caches?

evgeni: Depends on the definition of "new". A brand new one will be instant (there is nothing in the cache). An updated one, yes, that'll take a day if we don't purge.

But yes, we can extend deploy-web with a "smart" purge the following sub-paths to be sure.

evgeni: I've now changed that to 2 hours, which I think is reasonable, as I don't *expect* these pages to be updated *too* often.

I'll work on the smart purge nevertheless.

evgeni: step 1: https://github.com/theforeman/foreman-infra/pull/1530

ekohl (approved):

PR #1508, opened 2020-11-11
use runCicoPipelines in foreman-{deb,rpm} nightly pipelines

evgeni:


evgeni:
this should also finally collect artifacts from foreman jobs, not only katello :see_no_evil:

ekohl (approved):
In https://github.com/theforeman/foreman-infra/pull/1487 I also added a TODO for this, but this is a good improvement already.

PR #1510, opened 2020-11-11
Rewrite Smart Proxy PR testing to a pipeline

ekohl:
The main motivation for this was to get rid of the checkstyle plugin. It now rewrites to warnings-ng.

To keep things easy, it gets rid of the jenkins:rubocop task. This means the task will work on all branches without updating them all.

It only runs Rubocop on Ruby 2.7 because recordIssues has problems when it runs multiple times. This has the side effect of not running the tests on 2.7 when Rubocop fails. This means the number of tests fluctuates.


ekohl:
Note that currently https://ci.theforeman.org/job/smart-proxy-pr-test/ exists, but this is not guaranteed.

Another thing is that this also works for stable branches, but then it would also use the Ruby versions that develop should support. However, the current jobs also suffer from that.

tbrisker (commented):
I think stopping tests if rubocop fails is fine, we do the same on other jobs as well (since rubocop failures often lead to a quick push with fixes, the tests will trigger again anyways). otherwise, maybe we could run rubocop in parallel to the tests? Or split it to a separate check?

ekohl:
I did think about doing a clone first, run rubocop and then do the matrix testing. I can take a stab at that.

ezr-ondrej:
> I did think about doing a clone first, run rubocop and then do the matrix testing. I can take a stab at that.

I think it's the best aproach here. IMO RuboCop should be the first to run, as fast as possible and if it fails, do not run any other tests. As Tomer mentioned, RuboCop fail should be followed by quick repush with fixes. (If it's not, reviewers tend to point the RuboCop failures first anyway).

ekohl:
So what would you suggest?
```
clone -> bundle install with only the test group -> run rubocop -> (for each ruby version: clone -> bundle install -> run tests)
```

ezr-ondrej:
> ```
> clone -> bundle install with only the test group -> run rubocop -> (for each ruby version: clone -> bundle install -> run tests)
> ```

Yes! That's exactly I was thinking! :)
I'd even want to make a rubocop group (for core) to minimize the gemset as much as possible :)

ekohl:
Updated. I wish we could easily make the matrix configurable, but right now I don't see a good way. For now it's not that relevant anyway.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -0,0 +1,106 @@
+pipeline {
+    agent none
+	parameters {
+		string(name: 'ruby_versions', defaultValue: '2.5,2.6,2.7', description: 'Ruby versions to test with. Last value is used for Rubocop')
```
tbrisker, line 4: I guess this isn't used right now?

ekohl: Oops, no. That was an attempt that didn't work.

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -0,0 +1,103 @@
...
+                        post {
+                            always {
+                                junit testResults: 'jenkins/reports/unit/*.xml'
+                            }
+                        }
+                    }
+                }
+                post {
+                    always {
+                        archiveArtifacts artifacts: 'Gemfile.lock'
```
tbrisker, line 95: Perhaps if you move this to inside the matrix you could archive all of them, similar to how junit works? might need to rename them first to a different name for every version.

ekohl: This should already be part of the matrix. It probably needs to be renamed thought. However, I would like to get this merged first so I can continue with removing the checkstyle plugin.

tbrisker (approved):
makes sense, let's iterate on this

PR #1514, opened 2020-11-12
Turn on Katello EL8 nightly pipeline

ehelms:


ekohl (approved):
Currently the separate proxy is disabled on the EL8 pipeline but I'd like to enable that too. Have you tried that already?

ekohl:
Also, can we get 2.3 / 3.18 to support EL8? Since we just branched it feels like we're very close to that.

ehelms:
I think it only requires two fixes I know of:

 * https://github.com/Katello/katello/pull/9016
 * https://github.com/Katello/katello/pull/9024

@jlsherrill @ianballou y'all planning to pull those into 3.18? Want to try for EL8 support for it?

ehelms:
I am testing the proxy being enabled and will open a Forklift PR if that is successful.

PR #1515, opened 2020-11-12
Ensure jobs cannot configure global bundler config

ehelms:
When testing new job configs, global bundle configurations can be
accidentally created and break existing jobs. This change ensures
that this file does not exist on the Jenkins nodes.

Something in the past couple of days started creating the following:

```
---
BUNDLE_WITHOUT: "development"
```

This leads to some tests breaking such as Kafo PRs: https://ci.theforeman.org/job/kafo-pr-test/201/consoleFull

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -55,6 +55,19 @@ class slave (
     ensure  => absent,
   }
 
+  file { "${homedir}/.bundle":
+    ensure  => directory,
+    owner   => root,
+    group   => root,
+  }
+
+  file { "${homedir}/.bundle/config":
+    ensure  => file,
+    owner   => root,
+    group   => root,
+    content => '',
+  }
+
   # Build dependencies
   $libxml2_dev = $facts['os']['family'] ? {
     'RedHat' => 'libxml2-devel',


wbclark (approved):
This looks straightforward. You could save a few lines if you desire by adding this path to the array on L54

ekohl (commented):
This is IMHO not a real solution. A job can create it and then and other would run. That creates a very unreliable method. Puppet only runs every 30 minutes so nothing is guaranteed to work.

I did experiment with `bundle config` when rewriting the Smart Proxy job and I wonder if that triggered it.

ehelms:
That may have caused it, @ekohl . The timing is right. However, I'll take a 30 minute window to clean up accidental issues like this over spending nearly an entire day dumb founded.

ekohl:
A better solution would be to make it owned by root and have an empty content if this is a problem. That would really prevent one build from interfering with another.

ehelms:
This now creates an empty file owned by root. I'd like to do this at this level to prevent having to find all of the global configs as I am still running into issues.

puppet/modules/slave/manifests/init.pp
```diff
@@ -55,6 +55,13 @@
     ensure  => absent,
   }
 
+  file { "${homedir}/.bundle/config":
```
ekohl, line 58: I think you may need to ensure the directory as well. Otherwise you get errors about that.

puppet/modules/slave/manifests/init.pp
```diff
@@ -55,6 +55,19 @@
     ensure  => absent,
   }
 
+  file { "${homedir}/.bundle":
+    ensure  => directory,
+    owner   => root,
```
ekohl, line 60: Making this owned by root means bundler can't create any files, like `~/.bundle/cache` and `~/.bundle/ruby`. That will likely cause problems.

ehelms:
Turns out, whatever was done on the nodes caused this to also happen within job folders and `deleteDir` does not clear out hidden files. I'll just close this and clean up manually, and we'll have to keep an eye out for any other jobs that run into problems. @ekohl keep in mind whatever you did to cause this, so we don't trigger it again :)

PR #1517, opened 2020-11-13
run pulpcore pipelines

evgeni:


puppet/modules/jenkins_job_builder/files/centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -25,7 +26,8 @@ pipeline {
                 runPlaybook(
                     playbook: 'playbooks/setup_forklift.yml',
                     inventory: cico_inventory('./'),
-                    options: ['-b']
+                    options: ['-b'],
+                    extraVars: ['forklift_install_from_galaxy': install_from_galaxy]
```
ekohl, line 30: Doesn't this mean that we now set this variable for every forklift job? I think it would be cleaner to determine extraVars either as an empty array or hardcoded set to true if `params.type == 'pulpcore'`.

evgeni: every forklift setup job. but yes. While I see no harm in setting it to `false` for every job but pulpcore, I can also go and refactor that into setting extravars based on type

ekohl (approved):
Feel free to merge and monitor when you do.

PR #1497, opened 2020-11-04
Remove obsolete versions as of foreman 2.3

pcreech:

PR #1519, opened 2020-11-13
Jenkins node updates

ekohl:
* Drop MySQL from nodes
* Update RVM form 1.29.9 to 1.29.10
* Ordering fixes
* Experimental conditionals for EL8 nodes
* Vagrant config to allow spinning up nodes

### DIFF
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -236,7 +236,6 @@ class slave (
   include slave::mysql
   include slave::postgresql
   slave::db_config { 'sqlite3': }
-  slave::db_config { 'postgresql': }
 
   # RVM
   include slave::rvm
--- a/puppet/modules/slave/manifests/init.pp
+++ b/puppet/modules/slave/manifests/init.pp
@@ -235,7 +235,10 @@ class slave (
   # Databases
   include slave::mysql
   include slave::postgresql
-  slave::db_config { 'sqlite3': }
+  # sqlite3 support was dropped in Foreman 2.1
+  slave::db_config { 'sqlite3':
+    ensure => absent,
+  }
 
   # RVM
   include slave::rvm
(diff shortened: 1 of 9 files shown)


puppet/modules/slave/manifests/init.pp
```diff
@@ -229,10 +233,9 @@
   include slave::java
 
   # Databases
-  include slave::mysql, slave::postgresql
-  slave::db_config { 'mysql': }
+  include slave::mysql
```
ehelms, line 236: This is still used by something?

ekohl: If you notice the actual class you notice that it now ensures the package and db config are absent. That means after this has been rolled out we can remove it after a while. I assumed that a package removal stops the service. It doesn't remove the actual storage (DBs etc).

puppet/modules/slave/manifests/init.pp
```diff
@@ -229,10 +233,9 @@
...
+  include slave::postgresql
   slave::db_config { 'sqlite3': }
```
evgeni, line 238: I think this can be `ensure: absent` these days

ekohl: SQLite is still used in the 2.0 jobs. https://github.com/theforeman/foreman-infra/pull/1497 would probably solve it, but that PR is slow to be merged.

evgeni: it's merged now.

ekohl:
Rebased

puppet/modules/slave/manifests/init.pp
ekohl: Added as a separate commit now.

puppet/modules/slave/manifests/init.pp
evgeni: add a "#TODO remove when all nodes have been cleaned"?

evgeni (approved):

PR #1522, opened 2020-11-19
mash katello 3.18 el8

evgeni:


evgeni:
this explicitly does *not* enable pushing of el8 bits to the prod repo yet

ekohl (approved):

PR #1523, opened 2020-11-27
update focal exclude to have 2.3 and drop 2.0

evgeni:


ekohl (commented):
Normally this would be done as part of https://github.com/theforeman/foreman-infra/pull/1497 around branching.

Are there more places where this is needed?

evgeni:
I didn't see more, no. We don't build core/plugins for focal at all yet, just dependencies.

ekohl (approved):

PR #1526, opened 2020-12-02
plugins pipeline

evgeni:


puppet/modules/jenkins_job_builder/files/theforeman.org/yaml/jobs/pipeline/foreman-release-pipelines.yaml
```diff
@@ -52,12 +52,28 @@
...
+    sandbox: true
+    dsl:
+      !include-raw:
+        - 'pipelines/vars/foreman/{version}.groovy'
+        - 'pipelines/release/pipelines/plugins-pipeline.groovy{empty}'
+        - 'pipelines/lib/rvm.groovy{empty}'
+        - 'pipelines/lib/ansible.groovy{empty}'
+        - 'pipelines/lib/obal.groovy{empty}'
+        - 'pipelines/lib/rpm_packaging.groovy{empty}'
+        - 'pipelines/lib/release.groovy{empty}'
```
ekohl, line 67: I think that for these only `ansible.groovy` is needed since it only calls `runCicoPipelines` (which depends on `runPlaybook`)

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/plugins-pipeline.groovy
```diff
@@ -0,0 +1,22 @@
...
+    options {
+        timestamps()
+        timeout(time: 2, unit: 'HOURS')
+        disableConcurrentBuilds()
+        ansiColor('xterm')
+    }
+
+    stages {
+        stage('Install Test') {
+            agent { label 'el' }
```
ekohl, line 13: It only really depends on nodes that have the secret `centos-jenkins`

ekohl (changes requested):

puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/release/pipelines/plugins-pipeline.groovy
evgeni: so "any"? but `el` is what *all* the "run cico" stages are tied to, so maybe we should clean them up once in a follow up?

ekohl: That is indeed what I was thinking about.

evgeni: updated

evgeni:
@ekohl this has been updated.

ekohl (approved):

PR #1528, opened 2020-12-07
shorten DB names to not hit postgresql limits

evgeni:
now that we don't really have more than one DB, there is no need to have the db type in the gemset name


evgeni:
this *might* fix the source-release failure we've been seeing

ekohl (approved):

PR #1531, opened 2020-12-08
Fix foreman-installer tests that rely on checking for log

ehelms:
In newer Kafo versions, --help does not generate an initial empty
log file and thus we should not rely on it for test validation. The
fact that --help does not error should be sufficient.


ehelms:
@tbrisker this is to fix the error we have been seeing on develop and 2.3-stable

ekohl (approved):

PR #1542, opened 2021-01-11
don't expect version in plugins pipe

evgeni:
the plugins don't affect the version anyways


evgeni:
@ekohl this should fix the error you've been seeing

ekohl (approved):

PR #1547, opened 2021-01-13
fallback to "all" boxes, instead of none

evgeni:
otherwise you get errors like this:

    $ ansible-playbook playbooks/collect_debug.yml -l   -e pipeline_version=3.7 -e pipeline_os=centos8 -e pipeline_type=pulpcore -e foreman_expected_version=
    ansible-playbook: error: argument -l/--limit: expected one argument

which is not what we want as a fallback :)


ekohl (approved):
I've also wondered if `all` should always be used, just to simplify things. Git blame will show I'm the author of these lines, but prior to that it was on vars files. Ideally we would drop all of these. Alternatively I'm fine with letting forklift define a group and collect the group. My goal would be to have the minimal amount of code that needs to be kept in sync. Thoughts on that?

evgeni:
Yeah, this is probably a good next step.

I think the limit in the past was because we only wanted to collect debug from the "server" box, but in reality we want all of them.

Let's see how it turns out in the pulpcore case, and then we can generalize from there.

PR #1549, opened 2021-01-14
trigger nightly package builds in post

evgeni:
package release takes the last successfull source release as the source
for the artifacts, but when we trigger in a stage step, the source
release is not finished yet and package release will take the *previous*
successfull source release as the source, thus not actually building
what it was triggered for.


evgeni:
FWIW, I don't actually *know* whether this will fix the issue, but at least it has a higher chance to do the right thing.

ekohl (approved):

PR #1554, opened 2021-01-21
Only create JJB update job via puppet module

ehelms:
Now that https://ci.theforeman.org/job/foreman-infra-jenkins-jobs-update/ is handling updates for ci.theforeman.org and ci.centos.org the JJB puppet module need only deploy `foreman-infra-jenkins-jobs-update` itself.


puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -22,29 +22,13 @@
     notify  => Exec["jenkins-jobs-update-${config_name}"],
   }
 
-  cron { "jenkins-jobs-update-${config_name}-delete-old":
```
evgeni, line 25: those should be `ensure => absent` to really remove them.

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -22,29 +22,13 @@
...
-
   exec { "jenkins-jobs-update-${config_name}":
-    command => "jenkins-jobs --conf ${inifile} update ${directory}/${config_name} > /var/cache/jjb.xml",
+    command => "jenkins-jobs --conf ${inifile} update ${directory}/${config_name} foreman-infra-jenkins-jobs-update",
     timeout => $jenkins_jobs_update_timeout,
     path    => '/bin:/usr/bin:/usr/local/bin',
     require => File[$inifile],
   }
 
-  cron { "remove-unmanaged-jobs-${config_name}":
```
evgeni, line 40: those should be `ensure => absent` to really remove them.

evgeni (changes requested):
we can probably also remove the `unmanaged_jobs.rb` file deployment

puppet/modules/jenkins_job_builder/manifests/config.pp
ehelms: Ahh right, the ole absent then remove cycle. I forget that some(every)times.

evgeni (approved):

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -23,6 +23,7 @@
   }
 
   cron { "jenkins-jobs-update-${config_name}-delete-old":
```
ekohl, line 25: I won't miss this daily email from cron, but what is now responsible for deleting old (unmanaged) jobs?

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -31,13 +32,14 @@
   }
 
   exec { "jenkins-jobs-update-${config_name}":
-    command => "jenkins-jobs --conf ${inifile} update ${directory}/${config_name} > /var/cache/jjb.xml",
+    command => "jenkins-jobs --conf ${inifile} update ${directory}/${config_name} foreman-infra-jenkins-job-update",
```
ekohl, line 35: :-1: please revert. This is a defined type and should be generic. If anything, it should be a parameter.

ehelms: I missed that, gets confusing for me which things are generic and which things are specific for our infra. https://github.com/theforeman/foreman-infra/pull/1556

puppet/modules/jenkins_job_builder/manifests/config.pp
ehelms: Is an unmanaged job a job created out of band of JJB?

evgeni: we call `update --delete-old` which removes all jobs that are not currently defined in jjb

ekohl: Can be. It also deleted jobs that were removed from our JJB config.

PR #1557, opened 2021-01-21
Git checkout of jenkins jobs

ehelms:
I am thinking of an order of operation of changes to be:

 1) Switch to doing a git checkout of the jobs and creating the update job from that directory
 2) Change symlink from `ci/` to `jenkins-jobs/` and update any code
 3) Move the jobs to `jenkins-jobs/` and unlink `jenkins-jobs/` directory

I can do this all in one go if desired, but I was trying to think of small testable chunks to reduce potential infrastructure impacts.


ekohl (commented):
We've talked about splitting off the JJB configs to a separate repo. I'd like to hold off on this PR until we make a choice there.

ehelms:
This PR converts to doing a git checkout. I think having those mechanics in place and knowing they work is useful before we move the bits to a new repository. Then we just come back through here and change the checkout target to the new repo but we'll know the mechanics already work.

ehelms:
@ekohl Based on the discourse thread I think we can assume safe to move the jobs to a separate repository. Therefore I'd like to get this change in first, so we know git checkout works then do the move and update.

ehelms:
Could someone please take a look at this? I'd like to get the jobs move resolved.

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -23,29 +28,18 @@
     notify  => Exec["jenkins-jobs-update-${config_name}"],
   }
 
-  cron { "jenkins-jobs-update-${config_name}-delete-old":
-    ensure      => absent,
```
wbclark, line 27: I thought maybe we could miss some cleanup by removing the crons with `ensure => absent` but after checking https://github.com/theforeman/foreman-infra/commit/7495ec8373e6606dfbce94b939fbd215644b9bc9 I understand the logic here

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -6,6 +6,11 @@
   String $password,
   Integer[0] $jenkins_jobs_update_timeout = 600,
   String $command_arguments = 'foreman-infra-jenkins-job-update',
+  String $git_project_name = 'foreman-infra',
+  String $git_repo = 'https://github.com/theforeman/foreman-infra.git',
+  String $git_branch = 'master',
```
wbclark, line 11: It appears that this isn't used anywhere @ehelms . What is the intention?

ehelms: Dropped the parameter

wbclark (approved):

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -23,29 +27,18 @@
...
   }
 
-  cron { "jenkins-jobs-update-${config_name}-delete-old":
-    ensure      => absent,
-    command     => "timeout 1h jenkins-jobs --conf ${inifile} update --delete-old ${directory}/${config_name} > /var/cache/jjb.xml",
-    hour        => 0,
-    minute      => 10,
-    environment => 'PATH=/bin:/usr/bin:/usr/sbin',
-    require     => File[$inifile],
+  git::repo { "jenkins-jobs-${config_name}":
```
evgeni, line 30: does this make the `file { "${directory}/${config_name}": …}` above obsolete? Or is that still used somewhere?

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -23,29 +27,18 @@
...
+    target => "${directory}/${git_project_name}",
+    source => $git_repo,
+    args   => $git_args,
   }
 
   exec { "jenkins-jobs-update-${config_name}":
     command => "jenkins-jobs --conf ${inifile} update ${directory}/${config_name} ${command_arguments}",
```
evgeni, line 37: this seem to contradict the ` cwd     => "${directory}/${git_project_name}/${git_relative_path}/${config_name}",` below? which will win?

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -6,46 +6,28 @@
   String $password,
   Integer[0] $jenkins_jobs_update_timeout = 600,
   String $command_arguments = 'foreman-infra-jenkins-job-update',
+  String $git_project_name = 'foreman-infra',
+  String $git_repo = 'https://github.com/theforeman/foreman-infra.git',
```
ekohl, line 10: I'd prefer to do it right the first time and created https://github.com/theforeman/jenkins-jobs. It also includes the history for git log and git blame.

ekohl (changes requested):

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -6,46 +6,28 @@
...
-  file { "${directory}/${config_name}":
-    ensure  => directory,
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0755',
-    recurse => true,
-    purge   => true,
-    force   => true,
-    source  => "puppet:///modules/jenkins_job_builder/${config_name}",
+  git::repo { "jenkins-jobs-${config_name}":
```
ekohl, line 18: What's the mechanism to keep this repo updated? If you don't a notify can easily revert the job config back to an old state.

ehelms: If I don't notify what?

puppet/modules/jenkins_job_builder/manifests/config.pp
ehelms: I appreciate you making the repository. As the person making the change I did prefer to get the mechanism to work and then orchestrate the moving of the jobs given there are open PRs and such still to reduce the amount of potential churn on all fronts. At this point, I'll just update towards the ultimate goal and hope it works out.

puppet/modules/jenkins_job_builder/manifests/config.pp
ehelms: I suppose this is the way the git repo defines works (https://github.com/theforeman/puppet-git/blob/master/manifests/repo.pp) whereby it won't reclone on each run. How is this normally handled? Do I blow away the repository on each run to enforce the clone?

ekohl: Our module was really designed to be more of a "server" repo: it should be there so you can git push to it (via SSH). We never had to deal with pulling. You can use https://github.com/puppetlabs/puppetlabs-vcsrepo#git with ensure latest.

ehelms:
Now uses vcsrepo

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -6,46 +6,28 @@
...
-  }
-
-  cron { "remove-unmanaged-jobs-${config_name}":
-    ensure      => absent,
-    command     => "ruby ${directory}/${config_name}/unmanaged_jobs.rb ${inifile}",
-    hour        => 1,
-    minute      => 10,
-    environment => 'PATH=/bin:/usr/bin:/usr/sbin',
-    require     => File[$inifile],
+    cwd     => "${directory}/${git_project_name}/${config_name}",
```
ekohl, line 30: What's the benefit of setting `cwd` rather than passing an absolute path in the command? Absolute paths are easier to understand in debug logs since AFAIK `cwd` isn't really logged clearly.

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -6,46 +6,28 @@
...
 
-  file { "${directory}/${config_name}":
-    ensure  => directory,
-    owner   => 'root',
-    group   => 'root',
-    mode    => '0755',
-    recurse => true,
-    purge   => true,
-    force   => true,
-    source  => "puppet:///modules/jenkins_job_builder/${config_name}",
```
ekohl, line 22: You can remove these files now. Probably best in a separate commit but can be in the same PR.

ekohl (commented):
We'll also need to update https://ci.theforeman.org/view/Deploy/job/foreman-infra-jenkins-jobs-update/ to reflect this.

ehelms:
Updated over here -- https://github.com/theforeman/jenkins-jobs/pull/1 since this is a chicken-egg problem.

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -5,49 +5,30 @@
...
-  cron { "jenkins-jobs-update-${config_name}-delete-old":
-    ensure      => absent,
-    command     => "timeout 1h jenkins-jobs --conf ${inifile} update --delete-old ${directory}/${config_name} > /var/cache/jjb.xml",
-    hour        => 0,
-    minute      => 10,
-    environment => 'PATH=/bin:/usr/bin:/usr/sbin',
-    require     => File[$inifile],
-  }
-
   exec { "jenkins-jobs-update-${config_name}":
```
ekohl, line 25: Does it make sense to add `refreshonly => true` so it only runs i the `vcsrepo` resource changes?

ekohl (approved):

ekohl (approved):

PR #1560, opened 2021-01-26
remove old debian/ubuntu pbuilder configs

evgeni:


ekohl (approved):
There is still a cleanup cron that has `ensure => file`. Should that remain?

evgeni:
No, probably not. The `pbuilder` class should cleanup itself.

PR #1561, opened 2021-01-26
add debian jenkins node to Vagrant

evgeni:


evgeni:
@ekohl I wasn't sure what the best way was to install Puppet here, first I tried if/else in the original provisioner, then I overwrote it in the debian box. Neither feels too clean, but here we are :)

Vagrantfile
```diff
@@ -50,6 +50,22 @@ Vagrant.configure("2") do |config|
...
+    override.vm.hostname = "jenkins-node-debian10"
+    override.vm.box = "debian/buster64"
+
+    override.vm.provider "libvirt" do |libvirt|
+      libvirt.memory = "4096"
+    end
+    config.vm.provision "install puppet", type: "shell", inline: <<-SHELL
+      . /etc/os-release
+      wget https://apt.puppet.com/puppet6-release-${VERSION_CODENAME}.deb
+      apt-get install -y ./puppet6-release-${VERSION_CODENAME}.deb
```
ekohl, line 63: TIL: apt-get can install local files

Vagrantfile
```diff
@@ -50,6 +50,22 @@ Vagrant.configure("2") do |config|
...
 
+  config.vm.define "jenkins-node-debian10" do |override|
+    override.vm.hostname = "jenkins-node-debian10"
+    override.vm.box = "debian/buster64"
+
+    override.vm.provider "libvirt" do |libvirt|
+      libvirt.memory = "4096"
+    end
+    config.vm.provision "install puppet", type: "shell", inline: <<-SHELL
+      . /etc/os-release
```
ekohl, line 61: I like `/etc/os-release` :)

ekohl (approved):

Vagrantfile
evgeni: Yeah, and contrary to `dpkg` it will resolve dependencies when doing so ;)

PR #1563, opened 2021-01-26
Add management of Koji external repositories

ehelms:
Note, I have not tried to clean anything up here but to first capture what already exists.

### DIFF
--- a/koji/playbooks/external_repos.yml
+++ b/koji/playbooks/external_repos.yml
@@ -1,0 +1,13 @@
+---
+- hosts: localhost
+  gather_facts: false
+  vars_files:
+    - ../vars/external_repos.yml
+  tasks:
+    - name: Configure Koji External Repositories
+      ktdreyer.koji_ansible.koji_external_repo:
+        name: "{{ item.name }}"
+        url: "{{ item.url }}"
+        state: "{{ item.state | default('present') }}"
+        koji: 'koji'
+      loop: "{{ external_repos }}"
--- a/koji/vars/external_repos.yml
+++ b/koji/vars/external_repos.yml
@@ -1,0 +1,130 @@
+---
+external_repos:
+  - name: puppetlabs-pc1-fedora-24
+    url: file:///mnt/koji/external-repos/www/puppetlabsf24-$arch/RPMS.pc1/
+  - name: rhel-7.0-server
+    url: file:///mnt/koji/external-repos/rhel7-$arch/7.0/
+  - name: epel-7
+    url: file:///mnt/koji/external-repos/www/epel7-$arch/RPMS.epel/
+  - name: fedora-23
+    url: file:///mnt/koji/external-repos/www/fedora23-$arch/RPMS.os/
+  - name: fedora-23-updates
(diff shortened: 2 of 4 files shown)


evgeni:
when run, this will result in no changes, right?

ehelms:
That is the idea, I tested it with one to start with:

```
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] **************************************************************************************************************************************************************************************************

TASK [ktdreyer.koji_ansible.koji_external_repo] *******************************************************************************************************************************************************************
ok: [localhost] => (item={'name': 'puppetlabs-pc1-fedora-24', 'url': 'file:///mnt/koji/external-repos/www/puppetlabsf24-$arch/RPMS.pc1/'})

PLAY RECAP ********************************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

koji/vars/external_repos.yml
```diff
@@ -0,0 +1,128 @@
...
+  - name: puppetlabs-pc1-fedora-24
+    url: file:///mnt/koji/external-repos/www/puppetlabsf24-$arch/RPMS.pc1/
+  - name: rhel-7.0-server
+    url: file:///mnt/koji/external-repos/rhel7-$arch/7.0/
+  - name: epel-7
+    url: file:///mnt/koji/external-repos/www/epel7-$arch/RPMS.epel/
+  - name: fedora-23
+    url: file:///mnt/koji/external-repos/www/fedora23-$arch/RPMS.os/
+  - name: fedora-23-updates
+    url: file:///mnt/koji/external-repos/www/fedora23-$arch/RPMS.updates/
```
ekohl, line 12: Do we still have F23? I thought we cleaned that up a while back. Same with the other older Fedora versions.

koji/playbooks/external_repos.yml
```diff
@@ -0,0 +1,12 @@
+---
+- hosts: localhost
+  gather_facts: false
+  vars_files:
+    - ../vars/external_repos.yml
+  tasks:
+    - ktdreyer.koji_ansible.koji_external_repo:
```
evgeni, line 7:
```suggestion
    - name: Configure Koji External Repositories
      ktdreyer.koji_ansible.koji_external_repo:
```

koji/vars/external_repos.yml
ehelms: Nope, but see description :)

ekohl: I just assumed we actually removed the repos from disk.

ehelms: pretty sure we did, but given how manual its all been, the external repos in Koji were not cleaned up. I wonder what we should do about tags that these external repos are attached to as well (when we remove them).

ehelms:
Added new epel7 repository to the list

ehelms:
Further thoughts here? I wondered about whether I should add code that removes unmanaged tags to enforce management through this interface.

evgeni:
I think if that represents the current status quo, let's merge it (read: ACK), and then iterate on the cleanup later (read: in a direct followup)

PR #1566, opened 2021-01-27
Debian cleanup

evgeni:


ekohl (approved):

puppet/modules/slave/manifests/packaging/debian.pp
```diff
@@ -41,76 +41,6 @@
...
-          aptcontent => "deb $debian_mirror stretch main non-free contrib\ndeb-src $debian_mirror stretch main non-free contrib\n";
-        'xenial':
-          ensure     => present,
-          arch       => 'arm64',
-          release    => 'xenial',
-          apturl     => 'http://ports.ubuntu.com/ubuntu-ports',
-          aptcontent => "deb http://ports.ubuntu.com/ubuntu-ports xenial main restricted universe\ndeb-src http://ports.ubuntu.com/ubuntu-ports xenial main restricted universe\n";
       }
     }
     default: {}
```
ekohl, line 46: Should the whole case statement be removed if we only have one arch anyway?

evgeni: I wasn't sure. Maybe? We can revert that too if we ever have arm again :)

PR #1567, opened 2021-01-27
support using cowbuilder instead of pbuilder

evgeni:


evgeni:
draft, because it needs fixes in `camptocamp-pbuilder` to work reasonably

puppet/modules/slave/manifests/pbuilder_setup.pp
```diff
@@ -7,31 +7,41 @@
...
+        release   => $release,
+        methodurl => $apturl,
+      }
+      $update_pbuilder = "update_pbuilder_${name}"
+    }
+    'cowbuilder': {
+      pbuilder::cowbuilder { "${name}":
+        ensure     => $ensure,
+        arch       => $arch,
+        dist       => $release,
```
evgeni, line 27: this will become `release` in https://github.com/camptocamp/puppet-pbuilder/pull/13

ekohl:
In 7d3be35e53c0477818689af98f310bdf74957ddb we updated to a release of puppet/pbuilder which should include the required fixes, which would unblock this.

ekohl:
@evgeni I took the liberty to rebase this.

evgeni:
Thanks!

However, as noted on IRC, this PR is from the time where we had our builders on spinning rust, and there cowbuilder makes a speed difference vs plain pbuilder that needs to untar a chroot first. On NVMe that we have today, it's probably not worth the hassle, so I am going to close it.

If anything, I'd probably move to some more modern builder altogether (like sbuild), but that's more work, so not today :woman_shrugging:

PR #1572, opened 2021-01-28
support katello's switch to main branch

jlsherrill:


ekohl:
Why main? Wouldn't it be better to align it to Foreman and use develop if you want to avoid the word master?

jlsherrill:
Hey @ekohl, we discussed this among the katello team and decided to go with the github default, however i think this does warrent wider discussion for a 'foreman' standard default branch name across both katello and foreman orgs.  I'll start up a community post to discuss it and i've switched katello back to master now.  (In retrospect i shouldnt' have done this during the quiet period anyways!!!)

PR #1574, opened 2021-02-01
drop foreman 2.2 and katello 3.17

evgeni:


ehelms:

evgeni:
this now only contains infra changes

ekohl (changes requested):
If you really want to remove it, then you should also drop the `< 3.18` part from line 438. Also, the `>= 3.18` part is then always true and the conditional part can be merged into `mashes` unconditionally.

pcreech:
@evgeni now that Foreman 2.4 is out, do we want to move this out of Draft, or wait for katello 4.0?

evgeni:
waiting on 4.0/last 3.17

evgeni:
there are no more 3.17.z planned, so feel free to ACK and merge :)

PR #1580, opened 2021-02-08
Delete unmanaged Koji repositories

ehelms:


ekohl (approved):

koji/playbooks/external_repos.yml
```diff
@@ -1,13 +1,33 @@
...
+    - name: Extract external repository names
+      set_fact:
+        external_repository_names: "{{ external_repos | map(attribute='name') | list }}"
+
+    - name: Remove unmanaged external repositories
+      koji_external_repo:
+        name: "{{ item }}"
+        state: "absent"
+        koji: 'koji'
+      when: "{{ item not in external_repository_names }}"
```
evgeni, line 32:
```suggestion
      when: item not in external_repository_names
```

does that work too?

koji/playbooks/external_repos.yml
```diff
@@ -1,13 +1,33 @@
...
         url: "{{ item.url }}"
         state: "{{ item.state | default('present') }}"
         koji: 'koji'
       loop: "{{ external_repos }}"
+
+    - name: list of koji external repos from koji
+      koji_call:
+        name: listExternalRepos
+        koji: 'koji'
+      register: call_result
```
evgeni, line 21:
```suggestion
      register: external_repos_result
```

speaking names etc ;)

koji/playbooks/external_repos.yml
```diff
@@ -1,13 +1,33 @@
...
+      set_fact:
+        external_repository_names: "{{ external_repos | map(attribute='name') | list }}"
+
+    - name: Remove unmanaged external repositories
+      koji_external_repo:
+        name: "{{ item }}"
+        state: "absent"
+        koji: 'koji'
+      when: "{{ item not in external_repository_names }}"
+      loop: "{{ call_result['data'] | map(attribute='name') | list }}"
```
evgeni, line 33: would that also work, droping the `when`?

```yaml
loop: "{{ call_result['data'] | map(attribute='name') | reject("in", external_repository_names) | list }}"
```

evgeni:
I wonder if I want this to auto-delete all unmanaged repos. Would it be Puppet that runs every 30 minutes, I'd lean to no, but here… probably fine. (otherwise you'd need to add "state: absent" to the yaml, run, remove completely, which is nonsense too)

ehelms:
I debated auto-delete vs. printing a message and letting the user correct things. If we want to run this via a Jenkins job on merge to auto-manage, then I lean towards auto-delete. If not, messaging could make more sense? auto-delete is I think initially scary until we get mentally comfortable with it and remember to run everything through here. As in theory, everything is re-createale since its managed via git.

evgeni:
Yeah, execute manually or "on merge" sounds fine for deleting.

PR #1581, opened 2021-02-08
Cleanup removed external Koji repositories

ehelms:


koji/vars/external_repos.yml
```diff
@@ -1,61 +1,31 @@
...
-    url: file:///mnt/koji/external-repos/www/fedora20-$arch/RPMS.os/
-  - name: fedora-20-updates
-    url: file:///mnt/koji/external-repos/www/fedora20-$arch/RPMS.updates/
   - name: rhel-6.5-server
     url: file:///mnt/koji/external-repos/rhel6-$arch/6.5/
   - name: rhel-6.5-server-optional
     url: file:///mnt/koji/external-repos/rhel6-$arch/6.5-optional/
-  - name: fedora-22
-    url: file:///mnt/koji/external-repos/www/fedora22-$arch/RPMS.os/
   - name: epel-7-beta
```
ekohl, line 17: Why do we have epel-7-beta? Do we ever use this?

ekohl (commented):
If you remove them here, does it actually remove purge unmanaged entries or do you need to ensure absent on them?

ehelms:
This alone will not clean things up. That is what https://github.com/theforeman/foreman-infra/pull/1580 is for

koji/vars/external_repos.yml
ekohl: Any idea about this?

ehelms: Probably way back in the day we did. I am only not removing it here because I am performing a sequence of removals, with this round being old Fedora. Then I plan to tackle the EL landscape.

evgeni (approved):
this is super formal

PR #1586, opened 2021-03-01
split plugins test pipeline

evgeni:
this splits the pipeline definitions into EL and DEB, and uses that to create two plugin tests pipes

if that works as intended, I plan to also split the regular release pipes like this, following what we do in nightly

### DIFF
--- a/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman/2.2.groovy
+++ b/puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman/2.2.groovy
@@ -12,17 +12,30 @@ def foreman_el_releases = [
     'el8'
 ]
 def foreman_debian_releases = ['buster', 'bionic']
-def pipelines = [
+
+def pipelines_deb = [
     'install': [
-        'centos7',
-        'centos8',
         'debian10',
         'ubuntu1804'
     ],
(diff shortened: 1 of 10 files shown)


puppet/modules/jenkins_job_builder/files/theforeman.org/pipelines/vars/foreman/2.2.groovy
```diff
@@ -12,17 +12,30 @@ def foreman_el_releases = [
...
+    ],
+    'upgrade': [
+        'centos7',
+        'centos8',
+    ]
+]
+
+def pipelines = [
+    'install': pipelines_deb['install'] + pipelines_el['install'],
+    'upgrade': pipelines_deb['upgrade'] + pipelines_el['upgrade'],
```
ekohl, line 40: I think groovy doesn't understand trailing commas. Same in `pipelines_el`. Other than that I do like this organization.

evgeni: It does accept that, I tried :)

ekohl: Ok, that's good to know.

ekohl (approved):

PR #1587, opened 2021-03-01
add archivedeb vhost

evgeni:


puppet/modules/freight/files/archivedeb-HEADER.html
```diff
@@ -0,0 +1,27 @@
+<h1>The Debian Foreman archive repo</h1>
+
+<p>You will find them at:</p>
+
+<pre>deb http://archivedeb.theforeman.org/ &lt;codename&gt; &lt;component&gt;</pre>
+
+<p>Currently we have</p>
```
evgeni, line 7: this probably will need some update once we fill content here :)

puppet/modules/profiles/manifests/web.pp
```diff
@@ -37,6 +37,10 @@
   }
   contain web
 
+  class { 'web::vhost::archivedeb':
+  }
+  contain web::vhost::archivedeb
```
ekohl, line 42: If there are no parameters this is redundant:
```suggestion
  contain web::vhost::archivedeb
```

`contain` behaves like `include` but also makes sure that chaining works so if you have `Class['profiles::web'] -> Class['myclass']` you can be sure that there's `Class['web::vhost::archivedeb'] -> Class['myclas']`. With just `include` it "floats" and can show up anywhere in the dependency graph.

ekohl (approved):

PR #1588, opened 2021-03-01
don't mention old debian releases

evgeni:


ekohl (approved):
Staging is even older:
https://github.com/theforeman/foreman-infra/blob/d2421871570897966d9bba4c3443e9ee8fa24e4e/puppet/modules/freight/files/stagingdeb-HEADER.html#L7-L10

evgeni:
yeah, staging needs even more love ;)

PR #1589, opened 2021-03-01
more archivedeb setup

evgeni:


ekohl (approved):

puppet/modules/profiles/manifests/web.pp
```diff
@@ -37,6 +37,9 @@
   }
   contain web
 
+  class { 'web::vhost::archivedeb':
+    setup_receiver => $setup_receiver,
```
ekohl, line 41: Wait, this parameter doesn't exist

ekohl (changes requested):

puppet/modules/profiles/manifests/web.pp
evgeni: sure does :see_no_evil:

ekohl (approved):

PR #1594, opened 2021-03-26
purge katello-repos-latest from the CDN when publishing

evgeni:


ekohl (approved):

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -73,7 +73,7 @@ purgecdn() {
 	set +f
 	for d in "${REPO_PATH}"/*; do
 		purge_base="https://yum.theforeman.org/${REPO_DEST}/$(basename $d)"
-		fastly-purge ${purge_base} foreman-release.rpm foreman-client-release.rpm
+		fastly-purge ${purge_base} foreman-release.rpm foreman-client-release.rpm katello-repos-latest.rpm
```
ekohl, line 76: I haven't tested this, but does it work with katello and its subdirectories? or is it `katello/...`

evgeni: yeah, the purge base is what makes the magic :)

PR #1595, opened 2021-03-26
remove jenkins workspaces and gemsets older than 3 days

evgeni:


ehelms (approved):
I think there is some NPM caching that would be neat to clean up as well every now and again.

evgeni:
> I think there is some NPM caching that would be neat to clean up as well every now and again.

have a path handy where I could look for examples?

ehelms:
> have a path handy where I could look for examples?

```
# du -sh /home/jenkins/.npm/_cacache/
27G	/home/jenkins/.npm/_cacache/
```

puppet/modules/slave/manifests/init.pp
```diff
@@ -275,4 +275,12 @@
     mode    => '0755',
     content => "#!/bin/sh\nps -eo pid,etime,comm | awk '(\$2 ~ /-/ && \$3 ~ /ruby/) { print \$1 }' | xargs kill -9 >/dev/null 2>&1 || true\n", # lint:ignore:140chars
   }
+
+  file { '/etc/cron.daily/jenkins_cleaner':
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => "#!/bin/sh\nfind /home/jenkins/workspace/ /usr/local/rvm/gems/ -maxdepth 1 -mindepth 1 -type d -user jenkins -ctime 3 -print0 | xargs --no-run-if-empty --null rm -rf \n", # lint:ignore:140chars
```
ekohl, line 284: `-ctime 3` matches exactly 3 days, not older than 3.

Also, there are some files that shouldn't be cleaned up. I think these:
```console
$ ls /usr/local/rvm/gems -1
cache
ruby-2.0.0
ruby-2.0.0-p648
ruby-2.1
ruby-2.1.5
ruby-2.2
ruby-2.2.5
ruby-2.3
ruby-2.3.5
ruby-2.4
ruby-2.4.3
ruby-2.5
ruby-2.5.1
ruby-2.6
ruby-2.6.3
ruby-2.7
ruby-2.7.0
ruby-2.7.0@global
```

Also, you can use `-exec`:
```suggestion
    content => "#!/bin/sh\nfind /home/jenkins/workspace/ /usr/local/rvm/gems/ -maxdepth 1 -mindepth 1 -type d -user jenkins -ctime 3 -exec rm -rf {} +\n", # lint:ignore:140chars
```

ekohl (changes requested):

puppet/modules/slave/manifests/init.pp
ekohl: Oh, and another:
```suggestion
    content => "#!/bin/sh\nfind ${workspace} /usr/local/rvm/gems/ -maxdepth 1 -mindepth 1 -type d -user jenkins -ctime 3 -exec rm -rf {}\n", # lint:ignore:140chars
```

evgeni: > Also, there are some files that shouldn't be cleaned up. I think these:

Those are not owned by `jenkins` ;)

ekohl: I wonder if the cronjob should also run as the jenkins user to prevent accidental deletion.

evgeni: I think I applied all suggestions from this thread

evgeni:
@ehelms I'd prefer to handle the node cache in a follow up -- it needs a slightly different invocation ;)

ekohl (approved):

PR #1597, opened 2021-04-01
Update flatten-el8 with creating individual repos per module

pcreech:
This also simplifies deployment


koji/flatten-el8/rhel8-split.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 HOMEDIR=/root/koji-sync/podman/
-BINDIR=/usr/local/bin
+BINDIR=/root/koji-sync/podman/
```
ehelms, line 3: The BINDIR appears to be the same as the HOMEDIR now

koji/flatten-el8/rhel8-split.sh
```diff
@@ -1,6 +1,13 @@
...
+
+dnf install -y createrepo_c libmodulemd librepo python36 python3-hawkey python3-librepo python3-gobject-base "dnf-command(reposync)"
+
+
+pushd /root/koji-sync/podman/repos
```
ehelms, line 8:
```suggestion
pushd $HOMEDIR/repos
```

koji/flatten-el8/rhel8-split.sh
pcreech: Yes, this is intentional, as the executable is referenced $BINDIR/splitter.py.  (_could_ probably just $HOMEDIR it to minimize redundancy)

koji/flatten-el8/test.sh
```diff
@@ -1,4 +1,3 @@
-podman run --privileged -v /home/koji-sync/podman:/root/koji-sync/podman centos:8 /bin/sh /root/koji-sync/podman/script.sh
-
-
-rsync -r --delete -e ssh /home/koji-sync/podman/koji/staged/ root@koji.katello.org:/mnt/koji/releases/split/yum/koji-modules/koji/staged/ -v
+podman run --privileged -v ./:/root/koji-sync/podman centos:8 /bin/sh /root/koji-sync/podman/rhel8-split.sh
```
ehelms, line 1: Should this file have a shebang as the first line?

pcreech: Technically, probably.

koji/flatten-el8/README.md
```diff
@@ -1 +1,8 @@
 This is a quick hack I put together from what was done by the epel folks in fedora infrastructure to build packages against modules
+
+Usage:  
+
+	`./test.sh `
```
ehelms, line 5: Based on the name of this script, I thought if I ran it things could just execute locally and not make changes over on Koji. Looking at that script it looks like the thing to run to generate updated repositories -- is that accurate? If so, could we rename to better capture that so I don't go and do dumb things :)

pcreech: Yeah, naming could use an update.  The script is intended to be an all in one command.

ekohl: You're mixing 2 styles to make it code in markdown leading to weird rendering:
```suggestion
	./test.sh
```

koji/flatten-el8/rhel8-split.sh
pcreech: this should be fixed now.

koji/flatten-el8/rhel8-split.sh
pcreech: Done

koji/flatten-el8/test.sh
pcreech: Shebang added.

koji/flatten-el8/README.md
pcreech: Updated to a different name, and removed the back ticks.

ehelms (approved):

PR #1601, opened 2021-04-16
Add perl 5.26 modlule flattened external repo

pcreech:


evgeni (approved):

ehelms:
This will need to get added to any relevant tool_belt configurations that need to consume it.

PR #1602, opened 2021-04-23
kill perl:5.30 from the flattened repos

evgeni:
perl 5.26 is in Base, not AppStream, but thanks to our flattened layout,
we inject 5.30, which overrides 5.26 from base and breaks stuff.


evgeni:
@pcreech this means we can drop the `flat-perl26-el8` you added recently, it's empty anyways, as 5.26 is actually in Base, not AppStream. :exploding_head:

PR #1608, opened 2021-05-12
Correct API docs redirect

ekohl:
In c7c69ae4ec77d794f536ffe13bb8ea728dd0dba5 the apidoc redirect was made. However, it copied a line with variables and then changed to single quotes. That meant escaping was no longer needed. The result:

```console
$ curl https://theforeman.org/api/2.4/apidoc/v2/compute_profiles/create.html -si | grep Location
Location: https://apidocs.theforeman.org/foreman/$1
```

### DIFF
--- a/puppet/modules/web/manifests/vhost/web.pp
+++ b/puppet/modules/web/manifests/vhost/web.pp
@@ -27,7 +27,7 @@ class web::vhost::web(
     { 'rewrite_rule' => ['^/events/all.ics https://community.theforeman.org/c/events/l/calendar.ics [R,L]'] },
     { 'rewrite_rule' => ['^/events/? https://community.theforeman.org/c/events/l/calendar [R,L]'] },
     { 'rewrite_rule' => ["^/api/latest(.*) https://apidocs.theforeman.org/foreman/${stable}\$1 [R,L]"] },
-    { 'rewrite_rule' => ['^/api/(.*) https://apidocs.theforeman.org/foreman/\$1 [R,L]'] },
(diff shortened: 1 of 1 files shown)


evgeni (approved):

puppet/modules/web/manifests/vhost/web.pp
```diff
@@ -27,7 +27,7 @@
     { 'rewrite_rule' => ['^/events/all.ics https://community.theforeman.org/c/events/l/calendar.ics [R,L]'] },
     { 'rewrite_rule' => ['^/events/? https://community.theforeman.org/c/events/l/calendar [R,L]'] },
     { 'rewrite_rule' => ["^/api/latest(.*) https://apidocs.theforeman.org/foreman/${stable}\$1 [R,L]"] },
```
lzap, line 29: It's also here where I copied that blindly from, isn't this also incorrect?

lzap: Ah now I see it, okay. Single quotes are one of the worst features in many languages, including Ruby.

lzap (approved):

lzap:
Tests are green.

puppet/modules/web/manifests/vhost/web.pp
evgeni: No, because here it's double quotes, which are processed (so `$foo` becomes the value of `foo`) and you need to escape it to get an actual dollar sign.

(I hate languages which do this difference between single and double quotes, but here we are)

ekohl: Here you can see the Ruby influence on Puppet. And I doubt that at this point you really want to change such a fundamental thing in a language.

PR #1613, opened 2021-06-02
Drop EPEL 6 external koji repository

ehelms:


ekohl (commented):
We have EPEL6 added to our repos, but do we actually need it for the packages we build?

ehelms:
> We have EPEL6 added to our repos, but do we actually need it for the packages we build?

I could go run some mock builds to try and figure out if that is the case for our EL6 client packages.

ehelms:
Drop from tag -- https://github.com/theforeman/tool_belt/pull/373

This now drops epel 6 as my scratch build testing without EPEL 6 passed for the current packages being built there.

ekohl (approved):

PR #1614, opened 2021-06-02
Cleanup unused or no longer supported EL7 external repositories

ehelms:


koji/vars/external_repos.yml
```diff
@@ -8,40 +8,20 @@ external_repos:
   - name: epel-6
     url: file:///mnt/koji/external-repos/www/epel6-$arch/RPMS.epel/
   # EL7 repositories
-  - name: rhel-7.0-server
```
ekohl, line 11: This is used in https://github.com/theforeman/tool_belt/blob/master/configs/pulp/3.nightly.yaml but I think that isn't really used anymore since we have pulpcore, right?

koji/vars/external_repos.yml
```diff
@@ -8,40 +8,20 @@ external_repos:
...
     url: https://dl.fedoraproject.org/pub/epel/7/$arch/
-  - name: centos-sclo-rh-rhel-7
-    url: http://mirror.centos.org/centos/7/sclo/$arch/rh/
-  - name: epel-7-beta
-    url: file:///mnt/koji/external-repos/www/epel7-$arch/RPMS.beta/
-  - name: rhscl-1.1-rhel-7
-    url: file:///mnt/koji/external-repos/rhscl-rhel7-$arch/1.1/
-  - name: rhel-7.0-server-optional
-    url: file:///mnt/koji/external-repos/rhel7-$arch/7.0-optional/
   - name: epel7
```
ekohl, line 13: Where is this repo used? In tool_belt it looks like we use `epel-7` everywhere.

koji/vars/external_repos.yml
ehelms: Ahh yea, it does appear some Pulp 3 nightly tags use this external repo. @evgeni can you confirm if that tag setup is actively used? And should we update it to CentOS?

koji/vars/external_repos.yml
ehelms: Both are currently used:

```
# koji list-external-repos --used --name=epel7
Tag                  Pri Mode       External repo name
-------------------- --- ---------- -------------------------
katello-4.0-rhel7-build 4   koji       epel7
katello-nightly-rhel7-build 10  koji       epel7
```

I think it was added to allow transition to or maybe just testing of using an external URL. We could probably update these two tags to `epel-7` and then update `epel-7` to use a URL.

ekohl: Perhaps @pcreech can enlighten us here.

pcreech: It was a change done initially to not clobber epel-7 with the http url.  Once we were comfortable, we also updated epel-7, but never switched katello back over from epel7.

ehelms: Excellent, I have cleaned up those two tags (seems tool_belt was already set properly) and will push an update that drops this.

koji/vars/external_repos.yml
ehelms: Ok so those Pulp 3 "nightly" tags were quite old and so I have opened to remove that config https://github.com/theforeman/tool_belt/pull/374 then we can drop this repo

ekohl (approved):

PR #1619, opened 2021-06-28
add doc how to handle GPG keys

evgeni:


evgeni:
this should replace https://projects.theforeman.org/projects/foreman/wiki/GPG_Keys

ekohl (commented):
I'm just wondering: does it make sense to keep it here or in theforeman-rel-eng?

evgeni:
I'm fine with either.

ekohl (approved):
Given how we already have docs in this repo, it is probably a good place but a small part in the README of theforeman-rel-eng also feels appropriate. Perhaps with a link to this document.

evgeni:
@ekohl given the new docs in rel-eng, I re-did this, please have another look.

docs/gpg.md
```diff
@@ -0,0 +1,29 @@
+# GPG Keys
+
+## Summary
+
+After our security incident in July 2014, we planned to try and contain the scope of our GPG keys to avoid resigning lots of content if (or rather, when) a key is compromised or has to be revoked.
+
+* General use time based keys: for use with Debian archives, nightly packages etc.  Cycled every two years.
+* Major release (1.5, 1.6 etc) based keys: for use with tarballs, RPMs.  Expiry of one year.
```
ekohl, line 8: I think these examples are technically minor releases. Perhaps we should refer to this as just `Release based keys` instead of `Major release ...`?

docs/gpg.md
```diff
@@ -0,0 +1,29 @@
...
+
+See [Generating a new GPG Key for a X.Y release](https://github.com/theforeman/theforeman-rel-eng/#generating-a-new-gpg-key-for-a-xy-release) and [Generating a new GPG Key for signing the Debian repository](https://github.com/theforeman/theforeman-rel-eng/#generating-a-new-gpg-key-for-signing-the-debian-repository) for documentation how to do so.
+
+## Distributing keys
+
+### Release based keys
+
+RPM users are told in install & upgrade documentation to install foreman-release from the new release, which can contain the keys for that release, making distribution easy.
+
+### Time based keys
```
ekohl, line 22: Is this done via Puppet or still manual? If Puppet, can we link to a similar commit?

docs/gpg.md
evgeni: Yeah, sounds good.

docs/gpg.md
evgeni: "this" being adjusting `freight.conf`? 

7680053bb8fa5f682a41d912ccd89701d0f5b93f - Add 2016 archive key
9f50f62e0fbb120e5c1ce5b36d1d14e1aefb4949 - Remove 2014 archive signing GPG key

ekohl: Basically all steps in this chapter. So also making sure `foreman.asc` is correct on https://deb.theforeman.org/. Ideally that after you've read this, you know how to perform the procedure.

evgeni:
udpated

docs/gpg.md
```diff
@@ -0,0 +1,31 @@
...
+
+Debian archives can be signed with multiple keys (by setting those in `freight.conf`), but key distribution to users is manual right now.
+
+To make our infrastructure aware of the new keys, one needs to:
+
+* Export private key to `freight{,stage,archive}@web01`
```
ekohl, line 28: https://deb.theforeman.org/foreman.asc is a symlink to pubkey.gpg which is owned by freight, so freight takes care of it when promoting once you've exported the private key? Or do you also need to manually upload the public key somewhere?

evgeni: Correct, freight will keep this updated. But there is nothing that keeps it updated on the user system (contrary to foreman-release.rpm)

evgeni:
@ekohl updated with all the details I did today :)

ekohl (approved):

PR #1625, opened 2021-07-15
Manage Koji users

ehelms:

PR #1626, opened 2021-07-15
Disable koji users

ehelms:
Follow on to https://github.com/theforeman/foreman-infra/pull/1625 once the structure there is agreed to to disable inactive Koji users.

### DIFF
--- a/koji/playbooks/users.yml
+++ b/koji/playbooks/users.yml
@@ -11,13 +11,17 @@
         name: "{{ user.name }}"
         state: "enabled"
         permissions: "{{ user.permissions }}"
+        koji: 'koji'
       loop: "{{ enabled_users }}"
-      loop_var: user
+      loop_control:
+        loop_var: user
 
     - name: Set disabled users
       koji_user:
         name: "{{ user.name }}"
         state: "disabled"
         permissions: []
+        koji: 'koji'
       loop: "{{ disabled_users }}"
-      loop_var: user
+      loop_control:
+        loop_var: user
--- a/koji/vars/users.yml
+++ b/koji/vars/users.yml
@@ -1,51 +1,15 @@
 ---
 enabled_users:
-  - name: adprice
-    permissions:
-      - admin
-  - name: akofink
-    permissions:
-      - admin
-  - name: bbouters
-    permissions: []
-  - name: bizhang
-    permissions: []
-  - name: dadavis
-    permissions: []
-  - name: ddolguik
-    permissions:
-      - build
-      - repo
-  - name: dkliban
-    permissions: []
-  - name: dlobatog
-    permissions:
-      - admin
-      - build
-      - repo
   - name: ehelms
     permissions:
       - admin
-      - build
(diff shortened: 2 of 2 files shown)


koji/vars/users.yml
```diff
@@ -109,17 +74,38 @@ enabled_users:
...
-      - admin
   - name: zhunting
     permissions:
       - admin
       - build
       - repo
 
 disabled_users: []
+  - name: adprice
+    permissions: []
```
ekohl, line 85: It feels odd that you have to specify permissions for a disabled user. Can't we default that to an empty array?

ehelms: Fair thought, I hard set that in the playbook here https://github.com/theforeman/foreman-infra/pull/1625

koji/vars/users.yml
```diff
@@ -0,0 +1,97 @@
...
+    permissions:
+      - admin
+  - name: ewoud
+    permissions:
+      - admin
+  - name: jenkins
+    permissions:
+      - build
+      - repo
+  - name: jenkins.pulp
```
ekohl, line 18: Is this still used? Or only when Pulp 2 really is EOL?

ehelms: That was my assumption -- @pcreech do you know?

pcreech: It's the one for "nightly" builds of pulp.  Which, nightly hasn't really worked in a while anyawys.  We can probably kill 'em

ehelms:
Cleaned up and I think it's now an accurate reflection. Once this is merged I plan to go clean up the generated certificates on Koji itself.

koji/playbooks/users.yml
```diff
@@ -11,13 +11,17 @@
         name: "{{ user.name }}"
         state: "enabled"
         permissions: "{{ user.permissions }}"
+        koji: 'koji'
```
ekohl, line 14: Isn't this default?

ekohl (commented):
The list of users to disable looks correct to me. It does look odd to see some users with both `admin` and `build` / `repo` permissions and others don't.

koji/playbooks/users.yml
ehelms: Nope. https://github.com/ktdreyer/koji-ansible#koji-profiles -- check the first line.

ehelms:
Yea. I updated it so either someone is a fully admin or they have build/repo as I think those are two most common profiles.

koji/vars/users.yml
```diff
@@ -78,48 +39,46 @@ enabled_users:
   - name: kojira.katello.org
     permissions:
       - build
+      - repo
   - name: lzapleta
     permissions:
       - admin
```
ekohl, line 45: I recall @lzap wanted to hand over Koji maintenance. Do you want to keep admin access or is it sufficient to have regular build/repo permissions?

ekohl (approved):
I'm ok with starting here (already cleans up a lot).

PR #1627, opened 2021-07-16
[koji] Add management of Koji hosts

ehelms:


ekohl (approved):
I really hope we can at some point integrate this with Foreman so that we can add a host via Foreman and have it register properly.

PR #1629, opened 2021-07-23
add new (2021) debian signing key to freight conf

evgeni:


ekohl (approved):

evgeni:
I have imported that key into the keyring on web01, so it should be good. (famous last words)

PR #1630, opened 2021-07-23
add new (2021) key to pbuilder

evgeni:


ekohl (approved):
Perhaps we should use `/etc/apt/trusted.gpg.d` in a future patch.

PR #1637, opened 2021-08-20
document how to change expiration of a key

evgeni:


ehelms:
After modifying the expiration of the key, does anything need to be changed with respect to the public key?

evgeni:
No. The rest is done by freight.

ehelms (approved):

PR #1638, opened 2021-08-25
Install JQ and mark JSON.sh as the old implementation

ekohl:
This is a preparation for another PR.


evgeni (approved):
this requires EPEL on el7, but we have that anyways

PR #1642, opened 2021-08-27
Add F99printrepos and C10foremanlog hooks for pbuilder

ehelms:
These hooks originated from the Jenkins jobs that would build
debian packages and are added here as they are static for
every build.

### DIFF
--- a/puppet/modules/slave/manifests/pbuilder_setup.pp
+++ b/puppet/modules/slave/manifests/pbuilder_setup.pp
@@ -46,6 +46,18 @@ define slave::pbuilder_setup (
     content => file('slave/pbuilder_D80no-man-db-rebuild'),
   }
 
+  file { "/etc/pbuilder/${name}/hooks/C10foremanlog":
+    ensure  => $ensure,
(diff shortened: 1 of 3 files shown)


puppet/modules/slave/manifests/pbuilder_setup.pp
```diff
@@ -46,6 +46,18 @@
     content => file('slave/pbuilder_D80no-man-db-rebuild'),
   }
 
+  file { "/etc/pbuilder/${name}/hooks/C10foremanlog":
+    ensure  => $ensure,
+    mode    => '0775',
```
evgeni, line 51: IMHO all these files should be `0755`, not `0775`, but given this is not a new thing you introduce in this PR, we might want fix them all in one go later

evgeni (approved):

ekohl:
I'd be OK with this, but we should merge a PR to change the jobs to stop trying to change these files.

ehelms:
Let's merge https://github.com/theforeman/jenkins-jobs/pull/98 with this at the same time

PR #1643, opened 2021-08-30
add hook to dynamically configure repositories

evgeni:


ekohl (commented):
Overall I like this, but this does need a Jenkins job change to benefit from this. Is there already such a PR?

puppet/modules/slave/files/pbuilder_F60addforemanrepo
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Add the right theforeman.org repositories
+
+set -eu
+
+if [ -z "${REPO_SETUP:-}" ]; then
```
ekohl, line 6: Should we make this fail hard if it isn't set? In our builds we basically always want it anyway. Perhaps it should be 
```suggestion
if [ "${REPO_SETUP:-}" = 0]; then
```

puppet/modules/slave/templates/pbuilder_F60addforemanrepo.erb
```diff
@@ -0,0 +1,13 @@
...
+# Add the right theforeman.org repositories
+
+set -eu
+
+if [ -z "${REPO_SETUP:-}" ]; then
+  echo "Skipping automtic repository setup"
+  exit 0
+fi
+
+echo deb http://deb.theforeman.org/ <%= @$release %> ${REPO_VERSION:-nightly} > /etc/apt/sources.list.d/foreman.list
```
ekohl, line 11:
```suggestion
echo deb http://deb.theforeman.org/ <%= @release %> ${REPO_VERSION:-nightly} > /etc/apt/sources.list.d/foreman.list
```

ekohl (changes requested):
Bikeshedding: `FOREMAN_REPO_VERSION`? Or just `FOREMAN_VERSION`?

Can we get rid of `REPO_SETUP`? Remove the default value for `REPO_VERSION` and deal with it that way?

puppet/modules/slave/files/pbuilder_F60addforemanrepo
evgeni: My idea was: don't fail hard *now*, allowing us to deploy this and still play around with the jobs and when we feel comfortable with the actual flow, just drop the `if` altogether.

ekohl: Can we really? The filename is the same as the one in the current jobs so it'll flip back and forth between a job and what Puppet applies.

ekohl:
Should this include https://github.com/theforeman/foreman-infra/pull/1642 as well to make it complete?

puppet/modules/slave/files/pbuilder_F60addforemanrepo
evgeni: implementation details ;)

puppet/modules/slave/templates/pbuilder_F60addforemanrepo.erb
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Add the right theforeman.org repositories
+
+set -eu
+
+if [ -z "${FOREMAN_REPO_SETUP:-}" ]; then
```
ehelms, line 6: What's the case to skip this?

evgeni: https://github.com/theforeman/foreman-infra/pull/1643#discussion_r698433934

My idea was to easily switch between this and the "jenkins sets up things" for now, and remove later.

ekohl (approved):

ehelms (approved):

PR #1644, opened 2021-08-30
Use trusted.gpg.d over apt-key

ekohl:
This means it's no longer needed to install gnupg which makes it slightly faster.

I should note I didn't test this.


ekohl:
Updated

evgeni (approved):

PR #1646, opened 2021-08-30
Make all pbuilder hooks static

ekohl:
It also moves F60addforemanrepo to a static hook by using `$DISTRIBUTION` which according to man pbuilder should be present in every hook.

It also splits all hooks to separate files. Rather than changing the content depending on the preferences, certain files are deployed or not deployed. The whole directory is now managed and unmanaged files are purged. This is possible since Jenkins jobs no longer modify hooks.

Includes https://github.com/theforeman/foreman-infra/pull/1644


puppet/modules/slave/manifests/pbuilder_setup.pp
```diff
@@ -34,34 +34,29 @@
...
   }
 
-  file { "/etc/pbuilder/${name}/hooks/F60addforemanrepo":
-    ensure  => $ensure,
-    mode    => '0775',
-    content => template('slave/pbuilder_F60addforemanrepo.erb'),
+  file { "/etc/pbuilder/${name}/hooks":
+    ensure => directory,
+    mode   => '0775',
+    purge  => true,
```
ekohl, line 40: Note: I haven't checked if there are unmanaged hooks right now.

puppet/modules/slave/manifests/pbuilder_setup.pp
```diff
@@ -34,34 +34,29 @@
...
-    mode    => '0775',
-    content => template('slave/pbuilder_f70.erb'),
-  }
+  $hooks = [
+    'C10foremanlog',
+    'D80no-man-db-rebuild',
+    'F60addforemanrepo',
+    if $backports { 'F65-add-backport-repos' },
+    if $nodesource { 'F66-add-nodesource-repos' },
+    if $puppetlabs { 'F66-add-puppet-repos' },
```
ekohl, line 49: I must admit I haven't tested this just yet, but in theory it should work.

ekohl:
Bleh, looks like this doesn't work since pbuilder already manages the directory and doesn't have a purge parameter.

ekohl:
Includes https://github.com/theforeman/foreman-infra/pull/1650

ehelms (approved):
This looks good to me, I am not the Debian expert though. What sets $DISTRIBUTION?

ekohl:
> What sets $DISTRIBUTION?

https://github.com/theforeman/foreman-infra/pull/1650#issuecomment-911685989

evgeni (approved):

ekohl:
Rebased

puppet/spec/classes/slave_spec.rb
```diff
@@ -17,6 +17,13 @@
         else
           it { is_expected.to contain_users__account('jenkins').with_sudo('') }
         end
+
+        if facts[:osfamily] == 'Debian'
+          it { is_expected.to contain_file('/etc/pbuilder/buster64/hooks/F60addforemanrepos') }
```
evgeni, line 22:
```suggestion
          it { is_expected.to contain_file('/etc/pbuilder/buster64/hooks/F60addforemanrepo') }
```

puppet/spec/classes/slave_spec.rb
```diff
@@ -17,6 +17,13 @@
...
+          it { is_expected.not_to contain_file('/etc/pbuilder/buster64/hooks/F65-add-backport-repos') }
```
evgeni, line 23: this fails?

ekohl: Correct. I changed the code to ensure absent. Now the tests expect the same.

ekohl:
Updated, now with green tests.

evgeni (approved):

PR #1653, opened 2021-09-22
Jenkins node cleanup and Debian 11 support

evgeni:
sorry, I initially wanted to only add a Debian 11 builder node, but then things escalated ;)


puppet/modules/slave/manifests/unittests.pp
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  # Databases
+  include slave::mysql
+  include slave::postgresql
+  # sqlite3 support was dropped in Foreman 2.1
+  slave::db_config { 'sqlite3':
+    ensure => absent,
+  }
```
ekohl, line 168: We've had this for a while now so should we drop it?

puppet/modules/profiles/manifests/jenkins/packaging_node.pp
```diff
@@ -0,0 +1,19 @@
+# @summary A Jenkins packaging node
```
ekohl, line 1: Do we really need this and the unittest wrapper classes? Can't we set the booleans?

ekohl (commented):
For docker, shouldn't we ensure it's absent? Otherwise we have an unused service running on our nodes.

puppet/modules/slave/manifests/unittests.pp
evgeni: yepp, sounds good.

puppet/modules/profiles/manifests/jenkins/packaging_node.pp
evgeni: we certainly can. I was mainly looking at this as "a profile should be something that almost never needs overrides", and overriding the booleans would mean the opposite ;)

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -0,0 +1,172 @@
...
+      ensure => present;
+    $firefox:
+      ensure => present;
+    $libvirt_dev:
+      ensure => present;
+    $libcurl_dev:
+      ensure => present;
+    $sqlite3_dev:
+      ensure => present;
+  }
```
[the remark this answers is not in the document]
evgeni: probably, there was one (ansible) that has `ensure => latest`, but this now lives in a different place

I'd also convert to `ensure_packages` as we go?

puppet/modules/slave/manifests/init.pp
```diff
@@ -6,7 +6,8 @@
   Boolean $unittests = true,
   Boolean $packaging = true,
 ) {
-  $is_el8 = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8'
+  # Java
```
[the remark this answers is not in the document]
evgeni: don't shoot the refactorer! ;)

puppet/modules/profiles/manifests/jenkins/packaging_node.pp
ekohl: Looking at https://puppet.com/docs/pe/2019.2/the_roles_and_profiles_method.html I'd expect that more at the role level. Right now we haven't started with the role level (we barely have a profile layer).

puppet/modules/slave/manifests/unittests.pp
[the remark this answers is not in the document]
ekohl: I'd be ok with that.

puppet/modules/slave/manifests/docker.pp
```diff
@@ -19,11 +19,7 @@
...
     service_overrides_template  => false,
     docker_ce_package_name      => $docker_ce_package_name,
+    ensure                      => absent,
   }
 
-  group { 'docker':
-    ensure => 'present',
-  }
-
-  User<|title == 'jenkins'|>{groups +> ['docker']}
```
evgeni, line 28: I dropped this, but didn't make it remove jenkins from that group, as I didn't know if it won't explode trying to do so?

puppet/modules/slave/manifests/init.pp
```diff
@@ -123,148 +68,25 @@
...
+      koji_certificate => $koji_certificate,
+      uploader         => $uploader,
+      homedir          => $homedir,
+      workspace        => $workspace,
     }
-    default: {}
   }
 
-  if $facts['os']['architecture'] in ['x86_64', 'amd64'] and !$is_el8 {
+  if $facts['os']['architecture'] in ['x86_64', 'amd64'] and !$facts['os']['release']['major'] == '8' {
```
evgeni, line 89: we don't support debian 8, and I moved the `is_el8` definition away out of this class anyways :woman_shrugging:

puppet/modules/slave/manifests/docker.pp
ekohl: I'd expect you can ensure the group to be absent as well.

puppet/modules/slave/manifests/init.pp
```diff
@@ -3,8 +3,10 @@
   Boolean $uploader                     = true,
   Stdlib::Absolutepath $homedir         = '/home/jenkins',
   Stdlib::Absolutepath $workspace       = '/home/jenkins/workspace',
+  Boolean $unittests = true,
```
ekohl, line 6: Do you eventually want to make this
```suggestion
  Boolean $unittests = $facts['os']['family'] == 'RedHat',
```

evgeni: that's an interesting idea. I like it.

puppet/modules/slave/manifests/docker.pp
evgeni: no risk, no fun

ekohl (approved):

PR #1656, opened 2021-09-29
update unattended_upgrades to 6.0.0

evgeni:
this brings Debian 11 support

6.0.0 technically drops support for Debian 9, and we still have one machine running this: discourse, but there are no changes that actually should break Debian 9 -- just removal from upstream CI


evgeni:
hold on, we might just get a release of that thing out :D

evgeni:
green, yay

evgeni:
this escalated again a bit >.<

puppet/Puppetfile_forge
```diff
@@ -16,9 +16,10 @@ mod 'puppet/jenkins', '2.0.0'
 mod 'puppet/letsencrypt', '5.0.0'
 mod 'puppet/nodejs', '8.0.0'
 mod 'puppet/selinux', '3.1.0'
-mod 'puppet/unattended_upgrades', '5.1.0'
+mod 'puppet/unattended_upgrades', '6.0.0'
 mod 'puppetlabs/apache', '5.4.0'
 mod 'puppetlabs/apt', '7.7.0'
+mod 'puppetlabs/augeas_core', '1.1.2'
```
ekohl, line 22: This should only be a testing dependency. All _core modules are bundled in the agent packages (but not the gem)

evgeni: but we also have this here:

https://github.com/theforeman/foreman-infra/blob/1c15a1d3fa616a316795b38b6a7dfeae3cde223e/puppet/Puppetfile_forge#L27

ekohl: That is probably an oversight.

evgeni: updated

ekohl (approved):

PR #1657, opened 2021-09-30
use puppetlabs/postgresql 7.4.1

evgeni:
7.5.0 has broken Sensitive support


evgeni:
current execution fails with:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: no parameter named 'sensitive' (file: /etc/puppetlabs/code/environments/production/forge_modules/postgresql/manifests/server/role.pp, line: 89) on Postgresql_psql[CREATE ROLE foreman ENCRYPTED PASSWORD ****] (file: /etc/puppetlabs/code/environments/production/forge_modules/postgresql/manifests/server/role.pp, line: 89) on node deb-node01.jenkins.osuosl.theforeman.org
```

evgeni:
`puppet generate types --environment $env` fixed it

PR #1663, opened 2021-09-30
include rsync for debian packaging

evgeni:
also update rsync module to the latest release, not that there are many
changes besides docs, but hey


puppet/modules/slave/manifests/packaging/debian.pp
```diff
@@ -6,6 +6,8 @@
   Stdlib::HTTPUrl $debian_mirror = 'http://deb.debian.org/debian/',
   Stdlib::HTTPUrl $ubuntu_mirror = 'http://archive.ubuntu.com/ubuntu/',
 ) {
+  include rsync
```
ekohl, line 9: Should this be part of `freight::uploader` or is it also used before uploading?

evgeni: nah, that's the place!

ekohl (approved):

PR #1688, opened 2021-11-15
Calculate the Katello to Foreman mapping

ekohl:
This means the individual versions don't need to be added to the version map at every branching. Instead, it keeps a small static mapping and calculates the numbers dynamically. In this it assumes Katello's version is one major and 2 minor versions ahead.

### DIFF
--- a/koji/collection-mash-split.py
+++ b/koji/collection-mash-split.py
@@ -373,6 +373,22 @@ def run_mashes(collection, git_tag, mashes):
                        mash_config.compses, git_tag)
 
 
+def get_foreman_version(katello_version):
+    branch_map = {
+        'nightly': 'develop',
+        '4.1': '2.5',
+        '4.0': '2.4',
+        '3.18': '2.3',
+    }
+
+    try:
+        return branch_map[katello_version]
+    except KeyError:
+        major, minor = map(int, katello_version.split('.', 1))
+        # Katello (4 + x).(2 + y) -> Foreman (3 + x).(0 + y)
+        return "{}.{}".format(major - 1, minor - 2)
+
+
 def main():
     try:
         collection = sys.argv[1]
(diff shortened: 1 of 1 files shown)


koji/collection-mash-split.py
```diff
@@ -373,6 +373,22 @@ def run_mashes(collection, git_tag, mashes):
...
+        '4.0': '2.4',
+        '3.18': '2.3',
+    }
+
+    try:
+        return branch_map[katello_version]
+    except KeyError:
+        major, minor = map(int, katello_version.split('.', 1))
+        # Katello (4 + x).(2 + y) -> Foreman (3 + x).(0 + y)
+        return "{}.{}".format(major - 1, minor - 2)
```
evgeni, line 389: This assumes that Katello will keep its 4.y and Foreman its 3.y for a while, which I don't feel like is something we should rely on.

@tbrisker @jlsherrill @ehelms what do y'all think?

ekohl: The counter argument is that now we need to add every version. That means every branching it needs an update. If we do have a major version bump in either project or if there's more than 1 Katello release per Foreman release, we can then deal with it.

evgeni: Yeah, this is my main concern:
Right now, the procedure is easy "go and update the mapping"
With this change, it would become "go and see if the mapping needs updating" and if that wasn't needed for a few releases, people will skip this step in their minds.

That all said, I am not doing releases, so :woman_shrugging:

ekohl: We have [a line in Foreman's branching](https://github.com/theforeman/tool_belt/blob/aec8772665dab55ce4e056f8f5c5c66936e64bd8/procedures/foreman/branch.md.erb#L80) and then there's also [instructions in Katello's branching](https://github.com/theforeman/tool_belt/blob/aec8772665dab55ce4e056f8f5c5c66936e64bd8/procedures/katello/branch.md.erb#L99) but that's outdated and refers to a file that hasn't existed since [March 2020](https://github.com/theforeman/foreman-infra/commit/c64ede30cd542b4adbf3c1da1765efcefbbd8829).

That either means people ignore the procedures or don't update them when they're incorrect. Given I wonder what's best.

ehelms: Ole Murphy says that as soon as we make this change one of the projects will break the assumption :grin: 

We have multiple places in the project that need the mapping of Foreman version to Katello version -- is there some central place we could store it and then access it so we update it in one place?

PR #1689, opened 2021-11-17
Change Devel URL for Centos 8

Odilhao:
Looks like Centos 8.5 don't have the Devel repo ready to use, changing to Vault to ensure that our buildroot runs for EL8


pcreech (approved):
ACK

PR #1697, opened 2021-12-13
automate rss stats collection

evgeni:


tbrisker (approved):
My bash-fu isn't the best but makes sense to me.

puppet/modules/web/files/rss-stat.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+LOGPREFIX="/var/log/httpd/web-https_access_ssl.log-"
+OUTPUT=/var/log/rss-stat
+
+for logfile in ${LOGPREFIX}*; do
+  lognum=$(echo ${logfile}|sed "s#${LOGPREFIX}##")
```
ekohl, line 7: Do we need `sed`?
```suggestion
  lognum=${logfile#${LOGPREFIX}}
```

ekohl (commented):
It looks like we rotate by `YYYYMMDD` (which is good). `lognum` made me thing we still used the old `.n` part which would cause problems here.

puppet/modules/web/files/rss-stat.sh
evgeni: `lognum` is empty with that change :woman_shrugging:

evgeni: (renamed it to `logsuffix` now)

evgeni: > `lognum` is empty with that change woman_shrugging

I'm stupid, typo

puppet/modules/web/manifests/vhost/web.pp
```diff
@@ -89,4 +89,20 @@
...
+
+  # Generate RSS stats
+  file { '/var/log/rss-stat':
+    ensure => directory,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0750',
+  }
+
+  file { '/etc/cron.weekly/rss-stat':
```
ekohl, line 101: This is created via `/etc/logrotate.d/httpd`. I'm not sure I want to modify that, but I wonder if that would be better. I'm a bit worried that if we run this weekly and logrotate runs weekly that it may be a week behind most of the time (because cron would run before logrotate and then wait a week).

Still, we keep 4 weeks of backlog so it won't really be a big problem.

ekohl (approved):

puppet/modules/web/manifests/vhost/web.pp
evgeni: When talking to @tbrisker, we both agreed that making this executed from `logrotate` directly would be cool, but none of us is well-versed in `logrotate` and I always found it rather awful to debug, so didn't want to touch.

And yes, this makes us probably lag by a week, but then again, right now we lag by 3 weeks until @tbrisker does it ;)

PR #1698, opened 2021-12-20
Add two playbooks to generate modules metadata and then add it to an …

ehelms:
…existing repository

Module metadata needs to include the set of RPMs with full NEVRA
to allow dependency resolution. This assumes a base modules metadata
file and then analyzes a mashed repository to find all RPMs to add
dynamically to the metadata. The metadata is then added to the repository
via modifyrepo_c.

The workflow change would be:

 1) Run mash like normal
 2) Run make module metadata
 3) Run add module metadata

### DIFF
--- a/koji/playbooks/add-module-metadata.yaml
+++ b/koji/playbooks/add-module-metadata.yaml
@@ -1,0 +1,6 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Run modifyrepo_c to add module metadata
+      command: "modifyrepo_c --mdtype=modules {{ modules_metadata }} {{ repository_path }}/repodata"
--- a/koji/playbooks/make-modules-metadata.yaml
+++ b/koji/playbooks/make-modules-metadata.yaml
@@ -1,0 +1,31 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Collect RPM names
+      # The epoch number in the NEVRA string is mandatory per spec
+      # https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L668
+      # So we can't just use `%{nevra}` here
+      command: "rpm --query --package {{ repository_path }}/*rpm --queryformat='%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n'"
+      register: rpm_names
+
+    - name: Split RPM names into list
+      set_fact:
+        rpm_list: "{{ rpm_names.stdout_lines }}"
+
+    - name: Read base modules file
+      set_fact:
+        modules: "{{ lookup('file', modules_base) | from_yaml }}"
+
+    - name: Get current date for metadata version
+      set_fact:
+        version: "{{ lookup('pipe', 'date --universal +%Y%m%d%H%M%S') }}"
+
+    - name: Add RPM list to modules
+      set_fact:
+        modules: "{{ modules | combine({'data': {'artifacts': {'rpms': rpm_list}, 'version': (version|int)}}, recursive=True) }}"
+
+    - name: Write modules.yaml
+      copy:
+        dest: "{{ modules_output_path }}/modules.yaml"
+        content: "{{ modules | to_nice_yaml }}"


ehelms:
See https://github.com/theforeman/foreman-packaging/pull/7437 for base module metadata

koji/playbooks/add-module-metadata.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- hosts: localhost
```
ehelms, line 2: This could also be a step in the previous playbook gated by a variable. I, however, chose to start with discrete steps since these are two separate actions.

ehelms:
This *could* also be a chunk of code written in python and stand-alone or built into the mash script. This was inspired by some Satellite ecosystem work and trying to align on a similar process and tooling given the repository generation uses two different tools. We use this custom mash script and a lot of other projects rely on [pungi](https://pagure.io/pungi).

koji/playbooks/make-modules-metadata.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Collect RPM names
+      command: "rpm -qp {{ repository_path }}/*rpm --queryformat=\"%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n\""
```
ekohl, line 6: Isn't this easier? Note that this doesn't include epoch if epoch is 0, but I'd expect the tooling to understand that.
```suggestion
      command: "rpm -qp {{ repository_path }}/*rpm --queryformat='%{nevra}\n'"
```

And in scripts I prefer long options for readability:
```suggestion
      command: "rpm --query --package {{ repository_path }}/*rpm --queryformat='%{nevra}\n'"
```

ehelms: Easier yes, but that will not generate correct information. To quote [the spec](https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L668):

>     artifacts:

        # rpms:
        # RPM artifacts shipped with this module
        # A set of NEVRAs associated with this module. An epoch number in the
        # NEVRA string is mandatory.

ekohl: Thanks! That would make a good comment to prevent anyone in the future trying to "optimize" it :)

koji/playbooks/make-modules-metadata.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Collect RPM names
+      command: "rpm --query --package {{ repository_path }}/*rpm --queryformat=\"%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n\""
```
evgeni, line 6: I don't understand why this works. `command` doesn't use a shell, so the `*.rpm` should never be resolved.

```suggestion
      command: "rpm --query --package {{ repository_path }}/*rpm --queryformat='%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n'"
```

ekohl: Perhaps rpm itself implements globbing?

evgeni: Yes it does, and that scares me.

evgeni:
Some notes, mostly for myself trying to answer the question "are we - using this solution - maneuvering us in a corner where we rely on a customized setup, and can't move to some more modern tooling, like `pingi`":

* `pungi` has `module_defaults_dir` and `createrepo_extra_modulemd` in their [config docs](https://docs.pagure.org/pungi/configuration.html) which seem to allow injecting stuff into the generated compose.

ekohl:
I'm ok with merging this once you feel it's ready @evgeni

evgeni:
I am still pondering where exactly we want to call these from.

ekohl:
I think it should be called right after mashing, which is [collection-mash-split.py](https://github.com/theforeman/foreman-infra/blob/master/koji/collection-mash-split.py). To me it's better to encapsulate the whole mashing in a single command than to expose both and let the Jenkins job call both. It's also this mashing script which has the knowledge about which mashes for which version need to happen. Teaching Jenkins that feels like more duplication.

Having said that, I'm not sure calling Ansible playbooks from a Python script is the best. Personally I'd still consider a native Python implementation since you can do better error handling, but I'll leave that up to you.

koji/playbooks/make-modules-metadata.yaml
```diff
@@ -0,0 +1,31 @@
...
+      set_fact:
+        rpm_list: "{{ rpm_names.stdout_lines }}"
+
+    - name: Read base modules file
+      set_fact:
+        modules: "{{ lookup('file', modules_base) | from_yaml }}"
+
+    - name: Get current date for metadata version
+      set_fact:
+        version: "{{ lookup('pipe', 'date --universal +%Y%m%d%H%M%S') }}"
```
evgeni, line 22: hmm, I think I want to prepend this with the Foreman version, as otherwise a later generated metadata of Foreman 3.0 might look "newer" than the one from 3.1.

(I have no idea what this "version" is used for in DNF, it seems to work just fine with 0, but the spec says it needs to be set, and an unsigned 64bit int can easily fit that information)

ekohl: So would you do something like `(major * 100 + minor) * 10 ^ 14 + date`?

evgeni: something like that, yeah

koji/collection-mash-split.py
```diff
@@ -324,6 +341,9 @@ def handle_comps(self, whole_path, tmp_path, split_path, mash_config, arches, co
                 source = os.path.join(whole_path, mash_config, arch, "os", "Packages")
                 copied.update(self.copyout(comps_pkg_names, source, tmp_target))
                 self.createrepo(tmp_target, comps, checksum=checksum)
+                if modulemd_yaml:
+                    modulemd = self.get_from_git(gitloc, modulemd_baseloc, modulemd_yaml, tmp_target)
```
evgeni, line 345: I wonder if this should be unpacked to `tmp_target` or `comps_path` or something else. `get_from_git` returns a full path, so we don't care *where* on disk the file is, as long as we can read it, and we don't need the intermediate version in the resulting repo which `tmp_target` technically represents.

koji/collection-mash-split.py
```diff
@@ -160,6 +165,16 @@ def createrepo(self, path, comps=None, checksum=None):
...
 
+    def inject_modulemd_yml(self, path, modulemd_yaml):
+        """Generate modular metadata and inject it.
+
+        @param path: path to repository
+        @type path: str
+        @param modulemd_yaml: metadata template
+        @type modulemd_yaml: str
+        """
+        pass
```
evgeni, line 176: obviously a stub. I don't think running `ansible` in here is a good idea, so I'd probably re-implement the playbooks in pure python.

koji/collection-mash-split.py
```diff
@@ -160,6 +167,32 @@ def createrepo(self, path, comps=None, checksum=None):
...
+        # The epoch number in the NEVRA string is mandatory per spec
+        # https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L668
+        # So we can't just use `%{nevra}` here
+        cmd = "rpm --query --package {}/*rpm".format(path)
+        cmd += " --queryformat='%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n'"
+        self.logger.debug("running %s" % cmd)
+        status, output = kobo.shortcuts.run(cmd)
+        with open(modulemd_yaml) as modulemd_file:
+            modules = yaml.safe_load(modulemd_file)
+        modules['data']['artifacts'] = {'rpms': output.split()}
```
ekohl, line 187: Should it split on lines, in case there's a space in a filename somewhere?

koji/collection-mash-split.py
ekohl: It would save the whole `git archive`, though that probably isn't very expensive since it's just retrieving it for a single filename. That said, `git show remotes/origin/%s:%s` and return that would be more efficient. In `handle_extras` it would save writing it to disk altogether and in this case it would avoid writing it to disk once.

koji/collection-mash-split.py
evgeni: strictly speaking, those aren't file names, but NEVRAs, which shouldn't contain spaces.

but better safe than sorry, I guess.

koji/collection-mash-split.py
evgeni: also works with the comps usecase, I think

evgeni: mm, no, it wants an file-like object, or a path here.

StringIO here I come?

evgeni: I'd prefer moving this to a next PR, as an overall improvement of `get_from_git`, mostly because I'd prefer this PR to be as simple as possible ;)

evgeni: #1702 is a draft for that effort

evgeni:
the mash-split part of this is ready, and I considering dropping the ansible variant (well, moving it to an archive where we might pull it out again for the case we don't use mash-split anymore)

koji/collection-mash-split.py
```diff
@@ -160,6 +168,32 @@ def createrepo(self, path, comps=None, checksum=None):
...
+        @type path: str
+        @param modulemd_yaml: metadata template
+        @type modulemd_yaml: str
+        """
+        # The epoch number in the NEVRA string is mandatory per spec
+        # https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L668
+        # So we can't just use `%{nevra}` here
+        cmd = "rpm --query --package {}/*rpm".format(path)
+        cmd += " --queryformat='%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n'"
+        self.logger.debug("running %s" % cmd)
```
ekohl, line 184: If this is a regular Python logger, you can let it do the interpolation for you
```suggestion
        self.logger.debug("running %s", cmd)
```

ekohl (commented):
I'd leave the Ansible out of the PR. Given that you've essentially rewritten the whole PR, perhaps you should submit your own? Then bring this PR back to just Ansible so we can "archive" it that way.

koji/collection-mash-split.py
evgeni: I don't know, I was just copying code from the other functions ;)

but it does look like the normal python logger, yeah,

koji/collection-mash-split.py
```diff
@@ -396,7 +443,11 @@ def main():
         extras = []
 
         mashes = [MashConfig(collection, version, "rhel7-dist", "rhel7", "RHEL/7", extras)]
-        mashes.append(MashConfig(collection, version, "el8", "el8", "RHEL/8"))
+        if collection == 'foreman' and version == 'nightly':
```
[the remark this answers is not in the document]
evgeni: Yeah, I think I'd go for "3.2+" here and not backport it to existing releases. (same for Katello 4.3+)

koji/collection-mash-split.py
```diff
@@ -160,6 +168,32 @@ def createrepo(self, path, comps=None, checksum=None):
...
+        # https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L668
+        # So we can't just use `%{nevra}` here
+        cmd = "rpm --query --package {}/*rpm".format(path)
+        cmd += " --queryformat='%{name}-%{epochnum}:%{version}-%{release}.%{arch}\n'"
+        self.logger.debug("running %s" % cmd)
+        status, output = kobo.shortcuts.run(cmd)
+        with open(modulemd_yaml) as modulemd_file:
+            modules = yaml.safe_load(modulemd_file)
+        modules['data']['artifacts'] = {'rpms': output.splitlines()}
+        # this is still lacking the foreman version, but we don't have it available here
```
ekohl, line 189: Outdated comment now?

koji/collection-mash-split.py
```diff
@@ -440,9 +491,14 @@ def main():
         el8_candlepin = MashConfig(collection, version, "candlepin-el8", "candlepin-el8", "candlepin/el8")
         el8_candlepin.name = 'katello-candlepin-{}-el8'.format(version)
```
ekohl, line 492: This almost makes me think candlepin deserves is own module metadata. Probably overkill now, but it is a separate repo.

evgeni:
https://github.com/theforeman/foreman-infra/pull/1703

koji/collection-mash-split.py
evgeni: I would rather question why it is a separate repo…

ekohl: I'm guessing partly historical. Another is that Candlepin is tagged in Koji, rather than that the source lives in foreman-packaging.

evgeni:
closing for archival reasons, #1703 was merged.

PR #1702, opened 2022-01-25
simplify get_from_git

evgeni:


koji/collection-mash-split.py
```diff
@@ -277,9 +274,8 @@ def handle_extras(self, whole_path, mash_config, arches, extras, extras_path, gi
         for arch in arches:
             extras_cache_path = os.path.join(extras_path, mash_config, arch)
             extras_repo_path = os.path.join(whole_path, mash_config, arch, "os", "Packages")
-            extras_file = self.get_from_git(gitloc, extras_baseloc, extras + '-' + arch, extras_git_path)
-            with open(extras_file) as f:
-                self.download_extras(extras_cache_path, extras_repo_path, [line.rstrip() for line in f.readlines()])
+            extras_file = self.get_from_git(gitloc, extras_baseloc, extras + '-' + arch)
+            self.download_extras(extras_cache_path, extras_repo_path, extras_file.splitlines())
```
ekohl, line 278: This no longer strips lines in the extras file. Is that no longer needed?

ekohl (commented):
I like this, it simplifies things.

koji/collection-mash-split.py
```diff
@@ -83,16 +82,14 @@ def get_from_git(self, gitloc, baseloc, filename, finalloc):
             clone_cmd = "git clone %s %s" % (gitloc, workdir)
             kobo.shortcuts.run(clone_cmd, workdir="/mnt/tmp/gitrepo/", can_fail=True)
         final_fn = os.path.join(os.path.realpath(finalloc), filename)
-        cmd = "git fetch && git archive remotes/origin/%s %s | tar -C %s -x -f -" % (baseloc, filename, finalloc)
+        cmd = "git fetch && git show remotes/origin/%s/%s" % (baseloc, filename)
```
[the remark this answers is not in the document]
evgeni: it is, but baseloc is already `branch:basedir`:
https://github.com/theforeman/foreman-infra/blob/69341723cff2172b8f6ace78ffafa69c8d96d418/koji/collection-mash-split.py#L28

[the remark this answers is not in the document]
ekohl: Ah, that's not obvious but makes sense.

koji/collection-mash-split.py
evgeni: on the other hand, last time we used `extras` was in 2.1, and then we dropped it in b0522c9650645c5daaac60572426a2c7fbb78408, so maybe just nuke all that code?

ekohl: Oh, that's probably even better. We can always `git revert` if we do need it again.

ekohl (commented):
I think we should also clean up all code for versions prior to 2.5 With the modular metadata we implicitly sort of did that.

koji/collection-mash-split.py
```diff
@@ -213,13 +182,13 @@ def list_srpms_from_rpms(self, rpms_path):
...
+        @param comps: comps file content
         @type comps: str
         @return: package name
         @rtype: set
         """
         yum_comps = yum.comps.Comps()
-        comps_file = open(comps, "r")
+        comps_file = StringIO.StringIO(comps)
         yum_comps.add(comps_file)
         comps_file.close()
```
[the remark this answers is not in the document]
evgeni: Well, `StringIO.close` does free up memory, but so does garbage collection once this method call is over and the scope doesn't exist…

koji/collection-mash-split.py
```diff
@@ -498,7 +429,7 @@ def main():
         el8_candlepin = MashConfig(collection, version, "candlepin-el8", "candlepin-el8", "candlepin/el8")
         el8_candlepin.name = 'katello-candlepin-{}-el8'.format(version)
 
-        if version not in ('4.2', '4.1', '4.0'):
+        if version not in ('4.3', '4.2', '4.1'):
```
evgeni, line 432: this was technically a bug ;)

ekohl (approved):

ekohl:
@evgeni I approved but didn't merge because we have no automated way of deploying this.

evgeni:
Oh yeah, I didn't deploy it because I wanted a few runs to see if the old method is stable, but it's a good call, lemme merge that.

PR #1703, opened 2022-01-25
inject modulemd in mash-split script

evgeni:


koji/collection-mash-split.py
```diff
@@ -160,6 +168,31 @@ def createrepo(self, path, comps=None, checksum=None):
         self.logger.debug("running %s" % cmd)
         kobo.shortcuts.run(cmd)
 
+    def inject_modulemd_yml(self, path, modulemd_yaml, modulemd_version=0):
+        """Generate modular metadata and inject it.
+
+        @param path: path to repository
+        @type path: str
+        @param modulemd_yaml: metadata template
+        @type modulemd_yaml: str
```
ekohl, line 177: Should `modulemd_version` also be documented?

ekohl (approved):
Only small nits remain. I'd be OK with deploying this and see how it performs.

koji/collection-mash-split.py
evgeni: yes!

PR #1704, opened 2022-01-31
Switch from mirror to vault for Centos 8 content.

pcreech:
This is an intermittent step to switching our el8 content strategy.

### DIFF
--- a/koji/vars/external_repos.yml
+++ b/koji/vars/external_repos.yml
@@ -20,17 +20,17 @@ external_repos:
     url: http://mirror.centos.org/centos/7/updates/x86_64/
   # EL8 repositories
   - name: centos8-baseos
-    url: http://mirror.centos.org/centos/8/BaseOS/$arch/os/
+    url: http://vault.centos.org/centos/8/BaseOS/$arch/os/
   - name: centos8-appstream
-    url: http://mirror.centos.org/centos/8/AppStream/$arch/os/
+    url: http://vault.centos.org/centos/8/AppStream/$arch/os/
   - name: test-flat-el8
     url: http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001/
   - name: centos8-devel
     url: https://vault.centos.org/8.4.2105/Devel/$arch/os/
   - name: centos8-configmanagement-ansible-29
-    url: http://mirror.centos.org/centos/8/configmanagement/$arch/ansible-29/
+    url: http://vault.centos.org/centos/8/configmanagement/$arch/ansible-29/
   - name: centos8-powertools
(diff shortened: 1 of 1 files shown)


evgeni (approved):

koji/vars/external_repos.yml
```diff
@@ -20,17 +20,17 @@ external_repos:
...
+    url: http://vault.centos.org/centos/8/BaseOS/$arch/os/
   - name: centos8-appstream
-    url: http://mirror.centos.org/centos/8/AppStream/$arch/os/
+    url: http://vault.centos.org/centos/8/AppStream/$arch/os/
   - name: test-flat-el8
     url: http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001/
   - name: centos8-devel
     url: https://vault.centos.org/8.4.2105/Devel/$arch/os/
   - name: centos8-configmanagement-ansible-29
-    url: http://mirror.centos.org/centos/8/configmanagement/$arch/ansible-29/
```
ekohl, line 31: It looks like the SIG isn't archived. Perhaps they asked to be rebuilt against RHEL and kept? (https://lists.centos.org/pipermail/centos-devel/2022-January/120120.html said that's an option).

koji/vars/external_repos.yml
```diff
@@ -20,17 +20,17 @@ external_repos:
...
   - name: centos8-baseos
-    url: http://mirror.centos.org/centos/8/BaseOS/$arch/os/
+    url: http://vault.centos.org/centos/8/BaseOS/$arch/os/
   - name: centos8-appstream
-    url: http://mirror.centos.org/centos/8/AppStream/$arch/os/
+    url: http://vault.centos.org/centos/8/AppStream/$arch/os/
   - name: test-flat-el8
     url: http://koji.katello.org/kojifiles/releases/split/yum/koji-modules/koji/staged/x86_64/RHEL-8-001/
   - name: centos8-devel
     url: https://vault.centos.org/8.4.2105/Devel/$arch/os/
```
ekohl, line 29: Now I wonder if we're even using this. It's quite old.

evgeni:
I'd prefer to get this is as is, and do a cleanup later

koji/vars/external_repos.yml
pcreech: For the time being, this directory exists on vault:

https://vault.centos.org/centos/8/configmanagement/x86_64/ansible-29/

koji/vars/external_repos.yml
pcreech: Based on what I remember at the time, we had a dependency coming from this repo that got removed in 8.5+.  This repo was more of a catch-all of sorts by the Centos team for needed packages.  I don't remember which package at this time.

PR #1705, opened 2022-02-02
Fixes #33444 - drop old (2016) debian signing key

evgeni:


evgeni:
I'd merge this the day after 3.2 goes GA.

ekohl (approved):
Why the day after 3.2 goes GA?

evgeni:
mostly because infra changes aren't merged *together* with the release, and I think the promise was "3.0 will continue being verifyable with the old key as long as it's a supported release", which it won't be the day after 3.2 is out

ekohl:
That makes sense, thanks!

evgeni:
The key (563278F6) expired yesterday, and thus can't be used for signing anymore.

Instead of prolonging it again, I am going to merge this PR one day *before* 3.2 goes GA.

ekohl (approved):
I'm ok with that. It should unblock our nightly packages. It will probably also fix publishing 3.2 once we release it.

Issue #1677, opened 2021-11-05
New Sponsor - Conova


ehelms:
Conova has provided us with a new hardware resource for our use how we want to help the project.

Box:
  - HP 325
  - 24 core x2 (hyper-threading)
  - 64GB RAM
  - 2 1TB SSD
  - In Austria

Action Items:

- [ ] Setup libvirt
- [ ] Configure foreman-libvirt in foreman.theforeman.org
- [ ] Rename to virt01
- [ ] Document libvirt setup
- [ ] Provision VMs
  - [ ] deb-node01 - deb builder
  - [ ] node01 - CI node
- [ ] Update infrastructure docs (https://github.com/theforeman/foreman-infra/blob/master/docs/overview.md)

PR #1707, opened 2022-02-03
add conova to docs overview

evgeni:


ekohl (approved):
Addresses the last item in https://github.com/theforeman/foreman-infra/issues/1677. Thanks!

PR #1714, opened 2022-02-08
Revert "simplify get_from_git" (PR #1702)

evgeni:
this was incomplete and breaks mashing :see_no_evil:


ekohl:
I'd prefer if we can fix it up: https://github.com/theforeman/foreman-infra/pull/1715

PR #1718, opened 2022-02-22
call git fetch in an own call

evgeni:
otherwise the output will be part of the result, messing up parsing of
the same


ekohl (approved):

zjhuntin (approved):
APJ

PR #1719, opened 2022-02-24
Correct Vagrant provisioning override

ekohl:
Without this it tries to run the apt code on all boxes instead of only those particular builders.

Fixes: 5d8a9fe524abf3e5ebb96c951ba7f7fe8aa93705
Fixes: 387719b19ea050b3b7925741ca06dd46b780bb61


ekohl:
Replaced by https://github.com/theforeman/foreman-infra/pull/1720 which includes this and more.

PR #1721, opened 2022-02-25
install a modern version of createrepo_c

evgeni:


puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -61,6 +61,17 @@
     package { 'createrepo':
       ensure => present,
     }
+
+    yumrepo { 'pulpcore-3.16':
```
ekohl, line 65: Should we scope this to EL7 or is it also needed on EL8?

puppet/modules/web/manifests/vhost/yum.pp
```diff
@@ -61,6 +61,17 @@
     package { 'createrepo':
```
ekohl, line 61: Is this still needed if we install `createrepo_c`?

evgeni: probably not, but I wanted to keep a fallback, just to be sure ;)

puppet/modules/web/manifests/vhost/yum.pp
evgeni: no, el8 has a sufficiently new version, so can be el7 only

ekohl: I just merged https://github.com/theforeman/foreman-infra/pull/1720 so let's try to keep it ready for EL8 :)

ekohl (approved):

PR #1725, opened 2022-03-08
don't exclude metadata when syncing repo from koji

evgeni:
we need the metadata as generated on koji for the modules


evgeni:
@ekohl as discussed yesterday :)

ekohl (approved):
I verified this locally. I didn't end up testing the `dnf downgrade` scenario but that still works regardless until we remove `module_hotfixes=1`.

PR #1726, opened 2022-03-11
inject context for modulemd too

evgeni:


ekohl (approved):
Is this to fix https://community.theforeman.org/t/pulpcore-3-16-repo-issue/27685?

evgeni:
Correct. I linked it in the thread, but not here back to the thread ;)

PR #1728, opened 2022-03-25
Add CentOS Stream 9 external repos

pcreech:
Use web urls for now, until mirror internal mirror is set up.

"dnf module list" currently doesn't list any modules, so I'm curious if
we need to flatten it, or if default will suffice for now.


evgeni:
I hear rumors that default should suffice for now

evgeni:
What does "web" mean in this context? Those look like normal mirror urls to me?

pcreech:
> What does "web" mean in this context? Those look like normal mirror urls to me?

As opposed to "file" or "local" or "nfs" or something.  i.e. specifying that it's coming from over the interwebs instead of from the internal storage.

pcreech:
At the time, it felt like some kind of identifier would be nice.  Feel free to bikeshed the actual term, or we can just remove it.

evgeni:
Ah, gotcha. Yeah, I think i wouldn't have added any suffix, but in the end I don't think I care too much.

ekohl:
I'm also leaning to no suffix. For the buildroot it doesn't really matter what the source is IMHO.

PR #1735, opened 2022-04-21
add Ubuntu 22.04 Jammy to pbuilder

evgeni:


ehelms (approved):

ekohl:
"fun" fact: this never deployed properly since our deb node is Debian 10, but debootstrap is too old to know Jammy there. We need to update to Debian 11 and get debootstrap from backports. I opened https://github.com/theforeman/foreman-infra/pull/1773 to revert this until we can upgrade.

PR #1737, opened 2022-04-27
merge metadata instead of keeping it

evgeni:


evgeni:
draft, needs testing

puppet/modules/web/files/deploy-yumrepo.sh
```diff
@@ -46,7 +46,10 @@ do_rsync() {
 			fi
 
 			if [[ $MERGE == true ]] ; then
-				createrepo_c --skip-symlinks --update --keep-all-metadata .
+				zcat repodata/*-modules.yaml.gz > modules.yaml
+				rm repodata/*-modules.yaml.gz
```
ehelms, line 50:
```suggestion
				modifyrepo_c --remove modules repodata/
```

This prevents a warning like seen below and feels cleaner by using the tooling.

```
C_CREATEREPOLIB: Warning: Record with type "modules" already exists in repomd.xml
```

evgeni: Does that also remove unreferenced modules yaml files?

ehelms: Oh you want to clean out all modules.yaml.gz files, so perhaps both are needed to avoid the warning?

evgeni: Yeah. The idea is:
- We have an old yaml
- We rsync a new yaml
- We cat both into one
- We clean up
- We add the new one into the repomd

But yeah, both sounds good

ehelms: And do I understand correctly that the idea is to wind up with:

*modules.yaml.gz
```
---
  <module stuff
...
---
  <next set of module stuff>
...
---
  <next next set of stuff
...
```

ehelms (approved):

PR #1742, opened 2022-05-11
Revert "add Ubuntu 22.04 Jammy to pbuilder"

evgeni:
This reverts commit 60e38cdcb1a3037afba9d1c2f3f25fd1fe6e4d68.

Jammy uses zstd compression, which dpkg in Debian can't handle: https://bugs.debian.org/892664
This means we cannot bootstrap a Jammy chroot on our Debian-based build infrastructure.


evgeni:
might not be needed once we have https://github.com/theforeman/foreman-infra/pull/1741

PR #1747, opened 2022-05-18
don't fail if there is no moduledata to remove

evgeni:
when merging a fresh repo, there is only one modules.yaml.gz present
and that already gets removed by modifyrepo_c in the previous step.


evgeni:
cc @Odilhao this should fix the publishing error we've seen yesterday in katello 4.5

ehelms (approved):

PR #1751, opened 2022-05-25
Add GNU v3 License

ehelms:
Looking through the licenses of the external modules, they are all Apache or GPL v3. According to https://www.apache.org/licenses/GPL-compatibility.html I think that means we need to license the repository as GPL v3.


evgeni:
To the best of my understanding (IANAL!) the license of the individual modules doesn't have to influence our choice of license, as we never distribute a "binary" representation of the modules with our own code (the puppet catalog would be such a thing, but that is only distributed between our, private, systems).

That said, I am fine with whichever license you pick, as long as we can assume in good faith that all previous contributors also do that.

ehelms:
Fair -- I was going by the note we had " Figure out what license we should use based on the software we import (is Apache compatible with GPL? If yes, let's use it)". I would prefer Apache.. but I was unsure.

evgeni (approved):
I am good with this (re-)licensing!

ehelms:
@theforeman/infrastructure-contributors @theforeman/infrastructure-maintainers Please review this addition of a LICENSE to the repository and add a Github approval if you are OK with the change.

melcorr (approved):

ehelms:
I intend to merge one week from today.

https://github.com/orgs/theforeman/teams/infrastructure-contributors https://github.com/orgs/theforeman/teams/infrastructure-maintainers Please review this addition of a LICENSE to the repository and add a Github approval if you are OK with the change. Please reject if you have a concern and raise said concern for discussion.

PR #1752, opened 2022-05-26
Add puppet/postfix to have available in Foreman

ehelms:


ekohl:
This doesn't add the dependencies. I'm going to pick this up to fix on Jenkins, where we need it.

PR #1757, opened 2022-06-15
update jenkins plugin list

evgeni:


puppet/modules/jenkins_master/manifests/init.pp
```diff
@@ -59,6 +61,7 @@
     'create-fingerprint'                 => {},
     'credentials'                        => {},
     'credentials-binding'                => {},
+    'cvs'                                => {},
```
evgeni, line 64: actually, let's drop this one

PR #1761, opened 2022-07-06
Remove Ruby 2.6 RVM config

ekohl:
No job uses this anymore. The Ruby 2.4 config is still used by Kafo.

For now this is untested, but submitting this while I test it out.


ekohl:
It does leave a ruby-2.6.3 wrapper in `/usr/local/rvm/wrappers` but I don't know how to get rid of those. I think this is good enough now.

PR #1777, opened 2022-09-19
Rebuild Foreman infrastructure

ekohl:
This is to address the Big Update where Foreman is updated from version 1.24 to 3.1. 3.1 is chosen because it's the last version that can update the DB from pre-2.0 databases and also runs on EL8.

It also moves to a Hiera-based structure to assign classes rather than the Foreman ENC. This makes it easier to bootstrap a new environment.


ehelms:
> It also moves to a Hiera-based structure to assign classes rather than the Foreman ENC. This makes it easier to bootstrap a new environment.

:fireworks:

puppet/data/common.yaml
```diff
@@ -0,0 +1,33 @@
...
+puppet::server_environments_group: 'deploypuppet'
+puppet::server_puppetserver_telemetry: false
+puppet::show_diff: true
+
+users::users:
+  ekohl:
+    fullname: 'Ewoud Kohl van Wijngaarden'
+
+# TODO:
+# - sysadmins
```
ekohl, line 33: @evgeni I think we should discuss how to set this up. Today it's done via Foreman's ENC. The problem is that you can't bootstrap with the ENC (unlike Hiera). Once you apply Puppet it prevents sudo access (since for each user an individual rule is made, unlike a group).

Including users and their password hashes in a public repo isn't exactly what you'd want. Also the sysadmins email addresses shouldn't be public. I was thinking about our setup and wonder if we can use the wheel group for sudo access.

For email access we could also enhance the user via `foreman_users` in the ENC. It'd mean we would need to set the owner correctly, but that's not a bad thing IMHO.

ekohl: > For email access we could also enhance the user via `foreman_users` in the ENC. It'd mean we would need to set the owner correctly, but that's not a bad thing IMHO.

https://github.com/theforeman/foreman-infra/pull/1789 would be that.

docs/bootstrap.md
```diff
@@ -0,0 +1,61 @@
...
+
+Install a minimal EL8 install. Then:
+
+```sh
+dnf install https://yum.puppet.com/puppet7-release-el-8.noarch.rpm
+dnf install puppetserver
+. /etc/profile.d/puppet-agent.sh
+puppetserver ca setup --ca-name 'Foreman Puppet CA' --certname $HOSTNAME --subject-alt-names puppet.theforeman.org
+puppet config set --section agent server puppet.theforeman.org
+puppet config set --section main dns_alt_names puppet.theforeman.org
```
evgeni, line 15: Shouldn't this one be already there, as the Puppetserver defaults to adding `puppet.$domain`? Or did they stop doing this?

docs/bootstrap.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+With a basic Puppetserver running, deploy the environment. From your local machine:
+
+```
+# Download the latest from https://github.com/xorpaul/g10k/releases and place it in ~/bin
+git clone https://github.com/theforeman/foreman-infra
+cd foreman-infra/puppet
+g10k -cachedir ~/.cache/.g10k -puppetfile
+rsync -av ./* puppet01.conova.theforeman.org:/etc/puppetlabs/code/environments/bootstrap
```
evgeni, line 32: I guess the hostname here is an example? ;)

puppet/data/common.yaml
```diff
@@ -0,0 +1,33 @@
+foreman_servicename: 'foreman.theforeman.org'
+foreman_url: 'https://%{lookup("foreman_servicename")}'
+puppet_servicename: 'puppet.theforeman.org'
+
+foreman::repo::repo: '3.1'
+foreman::config::apache::proxy_no_proxy_uris:
+  - /icons
+  - /server-status
```
evgeni, line 8: we override this, because you want `/server-status` to work?
if so, I think this deserves a comment a la "can be removed once we use `theforeman/foreman` whichever version" or something like that?

puppet/data/common.yaml
evgeni: `wheel` doesn't exist on Debian, but I'd rather go and implement it there (via Puppet) than have those individual files again ;)

I must admit I'm always a bit cautious with using ENC in "important" places (like mail or sudo), as that effectively means that if the ENC goes bonkers for some reason¹, Puppet will re-reploy the config with *no* mail aliases or *no* sudo access, which will make finding and fixing errors harder.

¹: I've seen this happening when rolling out the deploy-ssh-keys to jenkins nodes (or rather, their registration in web01), never with the sudo part, but still.

docs/bootstrap.md
```diff
@@ -0,0 +1,61 @@
...
+
+```sh
+dnf install https://yum.puppet.com/puppet7-release-el-8.noarch.rpm
+dnf install puppetserver
+. /etc/profile.d/puppet-agent.sh
+puppetserver ca setup --ca-name 'Foreman Puppet CA' --certname $HOSTNAME --subject-alt-names puppet.theforeman.org
+puppet config set --section agent server puppet.theforeman.org
+puppet config set --section main dns_alt_names puppet.theforeman.org
+# To allow the foreman node a SAN
+sed -i '/allow-subject-alt-names/ s/false/true/' /etc/puppetlabs/puppetserver/conf.d/ca.conf
```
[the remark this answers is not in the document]
evgeni: (I guess we can enable that, sign the cert for `foreman.tfm.o` and disable it again?)

evgeni:
I left a few comments, but overall the plan sounds good to me!

puppet/data/common.yaml
ekohl: Technically that was a Foreman search, not the ENC. Since 26ff6baf21a92fd6bbe3461f35f67a2d531f9bd0 via the ENC, though that did have a fallback.

docs/bootstrap.md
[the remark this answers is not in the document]
ekohl: You're right about that. The alternative I thought about was using Let's Encrypt for a proper URL. Tricky part there is access to it. Right now I have set up a firewall to only allow port 443 from the Puppetserver and my home network. I guess we could set up rules so that it does IP checks in Apache for everything except `/.well-known` or use DNS validation, but that's a lot trickier.

docs/bootstrap.md
ekohl: Right now the domain is set to `conova.theforeman.org` so I think it wouldn't work well. I also had to add this at runtime because otherwise it would reject requests.

docs/bootstrap.md
ekohl: I thought about the same thing but was lazy and had a copy paste example here. I'll change it to example.com.

puppet/data/common.yaml
ekohl: Will add a note that we need at least 21.1.0

ekohl:
> Then the various machines need to be updated to the new Puppetserver. This will need work since there's a new CA. Additionally, they should be upgraded from Puppet 6 to 7.

At this point we should also remove all Puppet classes from hostgroups so we don't get duplicate definitions.

docs/bootstrap.md
evgeni: Ah right, because the hoster subdomain, yeah ok, I can see that.

evgeni:
> > Then the various machines need to be updated to the new Puppetserver. This will need work since there's a new CA. Additionally, they should be upgraded from Puppet 6 to 7.
> 
> At this point we should also remove all Puppet classes from hostgroups so we don't get duplicate definitions.

Is that required? I thought both HGs and the site.pp you have here end up being `include` and those can come multiple times?

ekohl:
IIRC `site.pp` is evaluated first and then the class is defined. Since it tries to pass parameters it's a concrete class and that fails.

puppet/data/common.yaml
ekohl: For now I've solved it by not including users and utility. This means we should continue setting that via Foreman's ENC for now.

What you would miss in a bootstrap then is the root email alias and the users. However, you have the initial user which is part of the wheel/sudo group and our new sudo rules respect that group. This means we can deal with it after the migration.

ekohl:
I changed the post a bit to have easier copy-paste steps and checkboxes so it's easier to mark progress.

Remaining open discussion points:
* Certificate SAN handling: https://github.com/theforeman/foreman-infra/pull/1777#discussion_r999086000
* Sync service name to puppet profile? So rename `puppet.t.o` to `puppetserver.t.o`

I'll be out until Tuesday so hopefully we can perform the migration next week.

evgeni:
FTR, I added a new host (`mirror01.koji.aws.theforeman.org`) to Puppet, but then realized I could have also used the new puppet setup?

docs/bootstrap.md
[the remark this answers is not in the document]
evgeni: I think I'd go with the easiest option, which to me seems to be "enable, sign, disable, (forget it 5 years)"?

evgeni:
> I changed the post a bit to have easier copy-paste steps and checkboxes so it's easier to mark progress.
> 
> Remaining open discussion points:
> 
>     * Certificate SAN handling: [Rebuild Foreman infrastructure #1777 (comment)](https://github.com/theforeman/foreman-infra/pull/1777#discussion_r999086000)
> 
>     * Sync service name to puppet profile? So rename `puppet.t.o` to `puppetserver.t.o`
> 
> 
> I'll be out until Tuesday so hopefully we can perform the migration next week.

IMHO `puppet` is more descriptive, so I'd stick to that.

evgeni:
For postgresql, I'd use `pg_dump -Fc` and `pg_restore` instead of a plain dump and `psql` to restore.

docs/bootstrap.md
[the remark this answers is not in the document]
ekohl: I'll opt to keep this in bootstrap for now, then when Puppet actually runs it should be disabled again.

ekohl:
So it looks like backups are also sent to the old puppetmaster. For now I've created https://github.com/theforeman/foreman-infra/issues/1805.

PR #1779, opened 2022-09-26
Update mash Pulpcore 3.21

Odilhao:
Making sure that EL9 is added for >= 3.18, and EL7 is only added for release <= 3.18


Odilhao:
@evgeni make sense?

ekohl (approved):
Not @evgeni but it makes sense to me.

PR #1786, opened 2022-10-18
Generate SSH keys in Puppet's vardir

ekohl:
Rather than hardcoding it to /etc/puppet (which doesn't exist on a Puppet AIO install) this uses the vardir setting. This works well for extlib::cache_data. As an additional benefit it can now run this in tests on a non-privileged user.

### DIFF
--- a/puppet/modules/profiles/manifests/web.pp
+++ b/puppet/modules/profiles/manifests/web.pp
@@ -17,16 +17,12 @@
 # @param rsync_max_connections
 #   Maximum connection per rsync target. Using a small value to try and reduce
 #   server load
-#
-# @param setup_receiver
-#   Set up the SSH receiver setup. Mostly turned off for testing.
 class profiles::web (
   String[1] $stable = '3.4',
   String[1] $next = '3.5',
   Hash[String, Hash] $debugs_htpasswds = {},
   Boolean $https = true,
   Integer[0] $rsync_max_connections = 10,
-  Boolean $setup_receiver = true,
 ) {
   contain awstats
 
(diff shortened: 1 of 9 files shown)


puppet/spec/classes/profiles_web_spec.rb
```diff
@@ -4,12 +4,11 @@
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let(:facts) { os_facts }
-      let(:params) { {setup_receiver: false} }
```
ekohl, line 7: Should we drop the whole parameter? It was only used for testing.

evgeni: If it was only for testing, yes!

evgeni:
I guess this will require a manual migration once? Or should we just regenerate the keys?

ekohl:
> I guess this will require a manual migration once? Or should we just regenerate the keys?

For now I made a symlink in `/opt/puppetlabs/server/data/puppetserver` (which is what `Puppet[:vardir]` resolves to on a server).

evgeni:
That works too!

ekohl:
Updated. This should tell us if https://github.com/theforeman/jenkins-jobs/pull/282 actually works.

puppet/modules/ssh/lib/puppet/parser/functions/ssh_keygen.rb
```diff
@@ -30,8 +30,7 @@ module Puppet::Parser::Functions
     # Ensure ecdsa uses keylength 521
     config['size'] = 521  if config['type'] == 'ecdsa'
 
-    # XXX hard coded path!
```
evgeni, line 33: :+1:

puppet/modules/profiles/manifests/web.pp
```diff
@@ -37,14 +33,8 @@
   }
   contain web
 
-  class { 'web::vhost::archivedeb':
-    setup_receiver => $setup_receiver,
-  }
   contain web::vhost::archivedeb
```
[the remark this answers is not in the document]
ekohl: Yes. `contain` is equal to `include` and making sure that if something requires on the caller (the web class in this case) that `web::vhost::archivedeb` must also be done.

evgeni (approved):

PR #1788, opened 2022-10-18
Drop Debian 9 & 10 support in our Puppet infra

ekohl:
Today we only run Debian 11


evgeni:
Did you update discourse01? ;)

ekohl:
I wish. I think that may be easier to do as part of a migration away from Scaleway so for now I'll add Debian 9 back.

PR #1792, opened 2022-10-24
document koji mirror setup

evgeni:


docs/koji-mirror.md
```diff
@@ -0,0 +1,47 @@
+# Koji Mirror
+
+| | mirror01.koji.aws.theforeman.org |
+|-|-|
+| type | AWS EC2 us-east-1 |
+| flavor | t2.medium |
```
evgeni, line 6: @pcreech is that correct? I guessed based on specs :)

puppet/modules/koji/files/mirror_reposync.cron
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
ekohl, line 1: Should this use `-e` so we know when it fails or does cron do that automatically?

puppet/modules/koji/files/mirror_reposync.conf
```diff
@@ -0,0 +1,80 @@
...
+
+[rhel-8-for-x86_64-baseos-rpms]
+name = Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
+baseurl = https://cdn.redhat.com/content/dist/rhel8/8/x86_64/baseos/os
+enabled = 1
+gpgcheck = 1
+gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslverify = 1
+sslcacert = /etc/rhsm/ca/redhat-uep.pem
+sslclientkey = /etc/pki/entitlement/3902935856580506602-key.pem
```
ekohl, line 12: Is this ID always the same or should it be a template with a parameter?

puppet/modules/koji/manifests/mirror.pp
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  apache::vhost { 'mirror.koji.theforeman.org':
+    serveraliases => ['mirror.koji.aws.theforeman.org'],
+    port          => 80,
+    docroot       => '/srv/mirror',
+    directories   => [
+      {
+        'path'    => '/srv/mirror',
+        'require' => 'ip 127.0.0.1',
```
ekohl, line 44: Might be nice to have the name, aliases and IP requirement as class parameters.

puppet/modules/koji/manifests/mirror.pp
```diff
@@ -0,0 +1,48 @@
+# Basic mirror for Koji
+class koji::mirror {
+  ensure_packages(['dnf-plugins-core'])
+
+  file { '/srv/mirror':
```
ekohl, line 5: I'd also make this a class parameter, given the repetition.

puppet/modules/koji/files/mirror_reposync.cron
evgeni: cron does automatically if there is output, we're using `--quiet` to `dnf`, so by default it will be silent, but I'd expect output when there are errors -- will test

puppet/modules/koji/files/mirror_reposync.conf
evgeni: I think it's the same per machine, so parameter it is

puppet/modules/koji/manifests/mirror.pp
evgeni: yeah, this part isn't done yet, I need to figure out the AWS VPC layout for that

puppet/modules/koji/manifests/mirror.pp
evgeni: ack

puppet/modules/koji/files/mirror_reposync.cron
evgeni:
```console
[root@mirror01 ~]# dnf --config reposync.conf reposync --download-path=/srv/mirror --download-metadata --quiet
[root@mirror01 ~]# vim reposync.conf 
[root@mirror01 ~]# dnf --config reposync.conf reposync --download-path=/srv/mirror --download-metadata --quiet
Error: Failed to download metadata for repo 'codeready-builder-for-rhel-9-x86_64-rpms': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
[root@mirror01 ~]# echo $?
1
```

IMHO sufficient?

puppet/modules/koji/templates/mirror_reposync.cron.erb
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
ekohl, line 1: Could we use [epp](https://puppet.com/docs/puppet/7/lang_template_epp.html) for new templates? I like the validation it provides and there's also better handling for the `Sensitive` data type. Not that that part is relevant here, but I'd eventually like to convert all templates.

evgeni:
IMHO this is ready for the first "round" -- as the flattening is missing it won't produce repos that are useable by Koji yet, but it's a start and I intend to add the flattening later.

evgeni:
should this get a `profile::koji::mirror` too?

ekohl:
> should this get a `profile::koji::mirror` too?

Yes please

ehelms (approved):

ekohl:
@evgeni what's the status on this?

evgeni:
I was holding off given the Copr initiative

evgeni:
closing

PR #1796, opened 2022-11-07
use the latest stable, not nightly in the deb header text

evgeni:


ekohl (approved):
I like this.

evgeni:
> I like this.

I hoped you do ;)

PR #1803, opened 2023-01-13
Use chrony on EL7 too

ekohl:
Previously we ran chrony on EL7, but the bulk of the hosts didn't use this base profile yet so it wasn't noticed. After the migration this started to show up.

Because it ran on some hosts, this also includes code to remove ntp from the system.

Fixes: e519403be6959247ab8632599dfc1edaa51c2e90

### DIFF
--- a/puppet/modules/profiles/manifests/base.pp
+++ b/puppet/modules/profiles/manifests/base.pp
@@ -6,7 +6,11 @@ class profiles::base (
   include ssh
   include timezone
   include unattended
-  if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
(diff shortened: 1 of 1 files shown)


ekohl:
I noticed this on master02 where chrony ran, but web01 and redmine01 already ran ntp before this. Still, I think it's best to use chrony. Quoting https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system_administrators_guide/ch-configuring_ntp_using_the_chrony_suite#sect-Choosing_between_NTP_daemon
>  Chrony should be preferred for all systems except for the systems that are managed or monitored by tools that do not support chrony, or the systems that have a hardware reference clock which cannot be used with chrony.

evgeni (approved):

PR #1808, opened 2023-01-17
Include backup configuration

ekohl:
Prior to this we had this information in the Foreman ENC, but now it's included in Hiera. A possibly better solution would be to search for the correct hostnames using a foreman lookup, but since we don't migrate that often this is good enough for now.

### DIFF
--- a/puppet/data/common.yaml
+++ b/puppet/data/common.yaml
@@ -26,3 +26,27 @@ puppet::server_puppetserver_telemetry: false
 puppet::show_diff: true
 
 sudo::wheel_config: password
+
+dirvish::vaults:
+  ci-jenkins:
+    client: master02.rackspace.theforeman.org
+    tree: "/var/lib/jenkins/"
+    expire_default: "+7 days"
+    expire_rules:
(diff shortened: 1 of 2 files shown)


evgeni (approved):

ekohl:
Step 1 fails: no dirvish package is available. Looking on the puppetmaster I see:
```console
$ rpm -qv dirvish
dirvish-1.2.1-4.el5.noarch
```

It's packaged for Fedora, but not EPEL: https://src.fedoraproject.org/rpms/dirvish

For now I'll manually rebuild and install it so I can proceed.

ekohl:
Then:
```
  - nothing provides perl(Time::ParseDate) needed by dirvish-1.2.1-26.el8.noarch
  - nothing provides perl(Time::Period) needed by dirvish-1.2.1-26.el8.noarch
```

After enabling EPEL:
```
  - nothing provides perl(Time::Period) needed by dirvish-1.2.1-26.el8.noarch
```

So now I'm going to revert this and think about a different solution. Trying to get this working while the tool is unmaintained is probably not worth the investment.

PR #1809, opened 2023-01-17
Make our puppetserver the secret storage

ekohl:


evgeni (approved):

ekohl:
Rebased to resolve a conflict due to the dirvish revert.

ekohl:
Ended up reverting the site.pp change and apply it via the ENC. Also created both stores. I'd rather avoid hardcoding the hostname so I've created `secrets.theforeman.org` as a CNAME.

Now I've pushed the repos. Everyone should now update their config:
```console
$ cd ~/.password-store-theforeman-shared
$ git remote set-url origin secrets.theforeman.org:/srv/secretsgit/theforeman-passwords.git
$ cd ~/.password-store-theforeman-releases
$ git remote set-url origin secrets.theforeman.org:/srv/secretsgit/theforeman-release.git
```

I've removed the write bit on the old server so users can no longer push using `chmod g-w -R /srv/secretsgit`.

I'll email everyone who has access the same info.

PR #1812, opened 2023-01-24
Remove rbot module

ekohl:
We haven't ran rbot in ages. At this point it will probably need a serious effort to revive it. Then we can simply revert this commit.


evgeni (approved):
Deleted code is good code

PR #1815, opened 2023-01-26
ensure libyaml-devel is installed

evgeni:
psych stopped vendoring it and this breaks several tests


evgeni:
example of such a failure: https://community.theforeman.org/t/smart-proxy-develop-source-release-169-failed/32008/4

fails on node04.aws, succeeds on node02.aws

ekohl (approved):

PR #1816, opened 2023-02-17
Use user's own cache directory for g10k & shellcheck the script

ekohl:
g10k defaults to /tmp/g10k, which should be considered insecure. This uses one that isn't shared with other users.

Also pleases shellcheck.

### DIFF
--- a/puppet/modules/deploy/files/script.sh
+++ b/puppet/modules/deploy/files/script.sh
@@ -5,7 +5,7 @@ dir=$(mktemp -d)
 environment=production
 trap 'rm -rf ${dir}' EXIT
 git clone --quiet --depth 1 https://github.com/theforeman/foreman-infra "${dir}/"
-g10k -quiet -puppetfile -puppetfilelocation "${dir}"/puppet/Puppetfile -moduledir "${dir}/puppet/external_modules"
+g10k -quiet -cachedir "$HOME/.cache/g10k" -puppetfile -puppetfilelocation "${dir}/puppet/Puppetfile" -moduledir "${dir}/puppet/external_modules"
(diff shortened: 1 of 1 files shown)


puppet/modules/deploy/files/script.sh
```diff
@@ -1,11 +1,11 @@
...
 environment=production
-trap "rm -rf ${dir}" EXIT
-git clone --quiet --depth 1 https://github.com/theforeman/foreman-infra ${dir}/
-g10k -quiet -puppetfile -puppetfilelocation ${dir}/puppet/Puppetfile -moduledir ${dir}/puppet/external_modules
-rsync -aqx --delete-after --exclude=.git ${dir}/puppet/* /etc/puppetlabs/code/environments/$environment/
-puppet generate types --environment $environment --config /etc/puppetlabs/puppet/puppet.conf
-echo "Deploy complete at `date`"
+trap 'rm -rf ${dir}' EXIT
+git clone --quiet --depth 1 https://github.com/theforeman/foreman-infra "${dir}/"
+g10k -quiet -cachedir "$HOME/.local/g10k" -puppetfile -puppetfilelocation "${dir}"/puppet/Puppetfile -moduledir "${dir}/puppet/external_modules"
```
evgeni, line 8:
```suggestion
g10k -quiet -cachedir "$HOME/.local/g10k" -puppetfile -puppetfilelocation "${dir}/puppet/Puppetfile" -moduledir "${dir}/puppet/external_modules"
```

evgeni (approved):

puppet/modules/deploy/files/script.sh
ekohl: I'm not sure why I typed `~/.local` instead of `~/.cache` here. Also fixing.

PR #1819, opened 2023-02-21
Don't show Koji certificate diff & Jenkins profiles tests

ekohl:


ehelms (approved):
Test failures are unrelated

ekohl:
They actually are, since it's the tests I added that are failing. That's the reason I kept it open, but it's a good reminder I should get back to it.

ekohl:
It's now green.

Issue #1676, opened 2021-11-04
Update Foreman to 3.6


ekohl:
Foreman's current version is outdated and needs to be updated to the latest release.

ekohl:
Should be done as part of #1686

evgeni:
- update puppet modules
- stop services
- update foreman
- db stuff

ekohl:
> update puppet modules

https://github.com/theforeman/foreman-infra/pull/1831

> * stop services
> 
> * update foreman
> 
> * db stuff

We are now running version 3.6.1.

PR #1831, opened 2023-04-07
Update modules & Redmine data dir & ssh_keygen refactor

ekohl:
These are preparations I have in my backup branch that can already be merged. See individual commits for details.


ekohl:
This got a bit out of hand, but it also updates the Foreman modules because otherwise it became very hard to update the stdlib etc versions. It's easier to update Foreman (https://github.com/theforeman/foreman-infra/issues/1676) than to make compatible module releases for EOL versions, so I'll move forward with that.

ekohl:
Current failures about duplicate resources are because of the `present` -> `installed` switch with `ensure_packages`.

ehelms:
My approval still holds :smile:

PR #1832, opened 2023-04-10
Add ianballou to Foreman infra access

ianballou:
So I can release 3.5.3 and assist with 3.7.0.


ehelms:
@ekohl letting you handle this as you are working on release process

ianballou:
Actually... @ekohl do I not need this access for "Make sure [test_3_5_stable](https://ci.theforeman.org/job/test_3_5_stable/) and [smart-proxy3.5-stable-test](https://ci.theforeman.org/job/smart-proxy-3.5-stable-test/) are green using [verify_green_jobs](https://github.com/theforeman/theforeman-rel-eng/blob/master/verify_green_jobs)" ?

ekohl:
> @ekohl ah! okay, I assumed I had to play that role too. I'll close this then.

You can if you want, and then we can take care of it. It that did trigger me to extend the release engineering role requirements because it doesn't list Koji yet.

> Actually... @ekohl do I not need this access for "Make sure [test_3_5_stable](https://ci.theforeman.org/job/test_3_5_stable/) and [smart-proxy3.5-stable-test](https://ci.theforeman.org/job/smart-proxy-3.5-stable-test/) are green using [verify_green_jobs](https://github.com/theforeman/theforeman-rel-eng/blob/master/verify_green_jobs)" ?

Jenkins access is separate, and is probably a good idea. I think today we could modify `verify_green_jobs` to use the anonymous API, but right now the code always uses credentials.

ianballou:
> You can if you want, and then we can take care of it. It that did trigger me to extend the release engineering role requirements because it doesn't list Koji yet.

I'll stick to the release owner part :)



> Jenkins access is separate, and is probably a good idea. I think today we could modify `verify_green_jobs` to use the anonymous API, but right now the code always uses credentials.

Alrighty -- let me know if I need to initiate another request for Jenkins access.

PR #1846, opened 2023-04-26
Add postfix module and dependencies

ekohl:
I started to look through the module and see how to use it. One thing that concerns me is the dependency on camptocamp/augeas which appears unmaintained. I opened https://github.com/voxpupuli/puppet-postfix/pull/350 to remove it in most cases, which helps.

Next step is to add it to systems. I'm leaning to adding it to the base profile for EL8+ but still considering.


ekohl:
> Next step is to add it to systems. I'm leaning to adding it to the base profile for EL8+ but still considering.

I now only added it to the Jenkins controller.

PR #1849, opened 2023-06-20
Update what Rackspace is used for

ekohl:


docs/index.md
```diff
@@ -33,7 +33,7 @@ A list of the hosting we have, who provides it, and what capacity it has
 
 ### Rackspace
   * Previously sponsored, [being phased out](https://community.theforeman.org/t/rfc-moving-off-of-rackspace-infrastructure/17932)
-  * Hosts Jenkins and Foreman
+  * Used for Mailgun (to send email from Redmine and Discourse) and CloudFiles (to store Fastly CDN logs)
```
evgeni, line 36: The later is not correct anymore, @ares deleted the logs

evgeni (changes requested):
Should we also update the places where the services are now hosted?

docs/index.md
evgeni: the first one: yes

the second: why?

ekohl: I may be tired and somehow remembered web02, but that was web02.rackspace.

ekohl:
> Should we also update the places where the services are now hosted?

I'm not sure what you mean by this.

Otherwise updated.

evgeni (approved):

evgeni:
> > Should we also update the places where the services are now hosted?
> 
> I'm not sure what you mean by this.

Like: Jenkins server and Puppet server and Foreman are now at Conova, right?

PR #1850, opened 2023-06-21
Update Ruby 3.1 to 3.1.4

ehelms:

PR #1856, opened 2023-07-20
Update Jenkins nodes to Ruby 3.0.5

ehelms:


ekohl (approved):

ekohl:
Looks like this fails. In https://github.com/theforeman/foreman-infra/pull/1850#issuecomment-1644099976 I posted a longer comment but tl;dr: gcc on EL7 is too old.

PR #1860, opened 2023-08-07
add nalfassi SSH keys

evgeni:


ekohl (approved):
@ nofaralfasi can you verify?

evgeni:
(those keys were taken from a trusted source, so I will go with yes for now)

PR #1862, opened 2023-08-13
Add stagingyum vhost to web01 for Copr staging repositories

ehelms:


ehelms:
This would create the vhost but there is the question of how the content gets to the Vhost. In the case of Koji, web01 rsyncs the repositories to it via a special script.

In the case of Copr, where we are generating the repositories on a Jenkins node, I do not think that strategy will work. I think we would need to `scp` the repositories to web01 in that case. I don't think we do anything like this today to model this after exactly. My open questions:

 1. Is SCP the right way to transfer?
 2. How do we lock web01 down to just allow this SCP action when initiated from the Jenkins node?

puppet/modules/web/manifests/vhost/stagingyum.pp
```diff
@@ -0,0 +1,75 @@
+# @summary Set up the yum vhost
+# @api private
+class web::vhost::yum (
```
evgeni, line 3:
```suggestion
class web::vhost::stagingyum (
```

;-)

puppet/modules/web/manifests/vhost/stagingyum.pp
```diff
@@ -0,0 +1,75 @@
...
+    },
+    {
+      path            => 'repomd.xml',
+      provider        => 'files',
+      expires_active  => 'on',
+      expires_default => 'access plus 2 minutes',
+    },
+  ]
+
+  web::vhost { 'yum':
```
evgeni, line 30:
```suggestion
  web::vhost { 'stagingyum':
```

evgeni:
> This would create the vhost but there is the question of how the content gets to the Vhost. In the case of Koji, web01 rsyncs the repositories to it via a special script.
> 
> In the case of Copr, where we are generating the repositories on a Jenkins node, I do not think that strategy will work. I think we would need to `scp` the repositories to web01 in that case. I don't think we do anything like this today to model this after exactly. My open questions:
> 
>     1. Is SCP the right way to transfer?
>     2. How do we lock web01 down to just allow this SCP action when initiated from the Jenkins node?

(if anyone wants to see the "special" script, it's this one: https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/files/deploy-yumrepo.sh )

I think rsync-over-SSH is the right answer, because we already have infrastructure for that in place (Debian uses that to push the built packages from the builders to web01). See https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/manifests/vhost/deb.pp and https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/manifests/vhost/archivedeb.pp

ehelms:
Let's see if I got the hang of this with this update. I've also opened up a draft for the Jenkins job in order to align the changes necessary to pull this workflow off (https://github.com/theforeman/jenkins-jobs/pull/349).

If I understand this correctly, initiating an rsync from Jenkins will actually initiate the `deploy-stagingyum.sh` script, where it will do the rsync to a temporary location and then copy it to the deployed location on web01.

This change will require me to create a user via our Foreman? Or just an SSH key? I assume this gets created via some action in Foreman but then will need to be added to Jenkins.

puppet/modules/web/files/deploy-stagingyum.sh
```diff
@@ -0,0 +1,11 @@
+  # Make sure target dir can be created
+  YUM_PATH=`echo "${SSH_ORIGINAL_COMMAND}" | awk '{ print $NF }'`
+  PROJECT=`echo $YUM_PATH | /bin/cut -f2 -d/`
+  RELEASE=`echo $YUM_PATH | /bin/cut -f3 -d/`
+  mkdir -p <%= @home %>/rsync_cache/$PROJECT/$RELEASE
```
evgeni, line 5: this file needs to be deployed as a template, or this line not to have erb in it ;-)

evgeni:
this is missing something like
```patch
diff --git puppet/modules/profiles/manifests/web.pp puppet/modules/profiles/manifests/web.pp
index 9383587a..8fa6fa8d 100644
--- puppet/modules/profiles/manifests/web.pp
+++ puppet/modules/profiles/manifests/web.pp
@@ -60,4 +60,9 @@ class profiles::web (
     rsync_max_connections => $rsync_max_connections,
   }
   contain web::vhost::yum
+
+  class { 'web::vhost::stagingyum':
+    stable                => $stable,
+  }
+  contain web::vhost::stagingyum
 }
```

to actually load that class

puppet/modules/web/manifests/vhost/stagingyum.pp
```diff
@@ -0,0 +1,81 @@
+# @summary Set up the yum vhost
+# @api private
+class web::vhost::stagingyum (
+  String[1] $stable,
```
evgeni, line 4: I'd argue we don't need none of the "stable", "latest" handling in here, as we always will refer with the "real" release versions to the repos

puppet/modules/profiles/manifests/web.pp
```diff
@@ -60,4 +60,7 @@
     rsync_max_connections => $rsync_max_connections,
   }
   contain web::vhost::yum
+
+  class { 'web::vhost::stagingyum': }
+  contain web::vhost::yum
```
ekohl, line 65: Technically this is redundant. If you don't need parameters, go with the simple variant.
```suggestion
  contain web::vhost::yum
```

puppet/modules/web/templates/deploy-stagingyum.sh.erb
```diff
@@ -0,0 +1,11 @@
...
+  YUM_PATH=`echo "${SSH_ORIGINAL_COMMAND}" | awk '{ print $NF }'`
+  PROJECT=`echo $YUM_PATH | /bin/cut -f2 -d/`
+  RELEASE=`echo $YUM_PATH | /bin/cut -f3 -d/`
+  mkdir -p <%= @home %>/rsync_cache/$PROJECT/$RELEASE
+
+  # Permit transfer
+  $SSH_ORIGINAL_COMMAND
+
+  # Publish the site - stderr/out redirect is required to stop the noninteractive shell from hanging
+  rsync -rvx --delete-after /home/yumrepostage/rsync_cache/$PROJECT/$RELEASE /var/www/vhosts/stagingyum/htdocs/ 2>&1 >/dev/null ;
```
ekohl, line 11: Why doesn't this line use `@home`? And it could also use `yum_directory`
```suggestion
  rsync -rvx --delete-after <%= @home %>/rsync_cache/$PROJECT/$RELEASE <%= @yum_directory %>/ 2>&1 >/dev/null ;
```

puppet/modules/web/manifests/vhost/stagingyum.pp
```diff
@@ -0,0 +1,53 @@
...
+    docroot_group => $user,
+    docroot_mode  => '0755',
+    directories   => $yum_directory_config,
+  }
+
+  ['HEADER.html', 'robots.txt'].each |$filename| {
+    file { "${yum_directory}/${filename}":
+      ensure  => file,
+      owner   => $user,
+      group   => $user,
```
ekohl, line 48: I wouldn't mind making these owned by root so rsync can't overwrite them, even accidentally. Otherwise it may flip back & forth.

puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -114,4 +116,10 @@
   package { ['dnf', 'dnf-plugins-core']:
     ensure => present,
   }
+
+  secure_ssh::rsync::uploader_key { 'yumstage':
```
ekohl, line 120: This means it's present on every host. That may not be a bad thing, but in the receiver setup it's limited to just OSUOSL nodes and that feels inconsistent.

The primary reason we limited it in the past is that we had sponsored machines where others had root. I don't think that's the case today so we can simplify it and allow running it on every node.

For Debian it doesn't matter that much because it's just a single machine.

ehelms: I don't know if I tackled the update for this correctly so please check it.

puppet/modules/web/manifests/vhost/stagingyum.pp
```diff
@@ -0,0 +1,53 @@
...
+      path            => 'repomd.xml',
+      provider        => 'files',
+      expires_active  => 'on',
+      expires_default => 'access plus 2 minutes',
+    },
+  ]
+
+  secure_ssh::receiver_setup { $user:
+    user           => $user,
+    foreman_search => '(name = external_ip4 or name = external_ip6)',
```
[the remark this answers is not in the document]
ehelms: There is only 11 nodes, so I think we are OK and this search works.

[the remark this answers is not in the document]
ekohl: If all have both IPv4 and IPv6 then 11 * 2 = 22, so we'd miss 2 IPs. Today we're OK because not all nodes are dual stack.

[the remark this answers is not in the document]
ehelms: Agreed. And this current definition keeps it in line with the yum vhost as well:

https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/web/manifests/vhost/yum.pp#L31-L35

puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -114,4 +116,10 @@
...
+    user       => $user,
+    dir        => "${workspace}/staging_key",
```
ekohl, line 122: One thing to note: for other things I've also taken the approach of manually adding this key to Jenkins as a real secret and then use the Jenkins built in functionality. Perhaps worth considering here too.

ekohl (approved):
I think this is correct now.

ehelms:
I seem to be getting permission denied from Jenkins to web01:

```
08:26:07  + export 'RSYNC_RSH=ssh -i /home/jenkins/workspace/staging_key/rsync_yumstage_key'
08:26:07  + RSYNC_RSH='ssh -i /home/jenkins/workspace/staging_key/rsync_yumstage_key'
08:26:07  + /usr/bin/rsync -avPx foreman/nightly yumrepostage@web01.osuosl.theforeman.org:rsync_cache/foreman/nightly
08:26:08  Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
```

PR #1868, opened 2023-08-27
Allow and remove old yumstage key

ehelms:
I thought the naming was the issue with why the key in the `authorized_keys` was not matching the key on the Jenkins nodes but not the case. Thus I created an unused key and this allows cleaning it up.

### DIFF
--- a/puppet/modules/slave/manifests/packaging/rpm.pp
+++ b/puppet/modules/slave/manifests/packaging/rpm.pp
@@ -111,4 +111,9 @@ class slave::packaging::rpm (
   }
 
   include rsync
+
+  secure_ssh::rsync::uploader_key { 'yumstage':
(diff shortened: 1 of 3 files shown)


puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -122,4 +122,10 @@
     dir        => "${workspace}/staging_key",
     manage_dir => true,
   }
+
+  secure_ssh::rsync::uploader_key { 'yumstage':
+    ensure => 'absent',
+    user   => $user,
+  }
+
```
ekohl, line 130:
```suggestion
```

ekohl (approved):
Small nit: puppet-lint would flag this if we had it enabled.

PR #1870, opened 2023-08-27
rsync to project directory

ehelms:


ehelms:
@ekohl mind taking a look at this one? I need it to properly rsync over with the right structure and I figured we should reduce storage usage.

ekohl (commented):
Why? The idea behind keeping a cache is that you speed up future executions. Perhaps you need the `--delete` option to remove deleted files?

ehelms:
> Why? The idea behind keeping a cache is that you speed up future executions. Perhaps you need the `--delete` option to remove deleted files?

I also noticed that we do a clean up in freight? Is that for a different reason?

https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/freight/templates/rsync.erb#L13-L14



I assume you are saying the a delete would need to be added here? https://github.com/theforeman/jenkins-jobs/pull/349/files#diff-c12500cef361f017ede0862bcabf9b799c8e4b9287be181a1d74203617208968R482

ehelms:
For now I have isolated this just to the rsync location update to allow me to continue testing and will leave the conversation around cleanup to be separate on whether we do that or not.

ekohl:
Good question. Perhaps that's more a case of "it was always that way"?

PR #1871, opened 2023-08-28
install copr-cli on all nodes that do RPM packaging

evgeni:


evgeni:
otherwise, things fail rather confusingly: https://ci.theforeman.org/job/foreman-packaging-rpm-copr-pr-test/105/console

ekohl (approved):
We no longer need a copr config like we did prior to f1934f33bcdb8d4a553a5a41208386935c14f314?

PR #1876, opened 2023-08-30
add redmine test instance to fastly

evgeni:


ekohl (approved):

evgeni:
this won't be necessary.

PR #1882, opened 2023-09-04
various small changes to the redmine module

evgeni:
see individual commits, everything should be still working fine on the existing el7 setup


puppet/modules/redmine/manifests/init.pp
```diff
@@ -45,6 +45,15 @@
   }
 
   # Needed for bundle install
+  if $facts['os']['release']['major'] == '8' {
+    package { 'ruby dnf module':
```
ekohl, line 48: This be done before `rubygem-bundler` (and why do we list `rubygem-bundler.noarch`?) and `ruby-devel`?

puppet/modules/redmine/manifests/init.pp
```diff
@@ -126,7 +126,7 @@
   }
 
   exec { 'install redmine':
-    command     => 'bundle install',
+    command     => 'bundle install --path ./vendor',
```
ekohl, line 128: Technically `--path` is deprecated and `bundle config --local path ./vendor` + `bundle install` is the right way. Or write down the `.bundle/config` file explicitly.

evgeni: technically, yes. it worked tho, and was the path of lowest resistance ;-)

puppet/modules/redmine/manifests/init.pp
evgeni: it should be, yes. but I didn't know how to tell `ensure_packages` to add a before relationship, and it worked fine w/o

puppet/modules/redmine/manifests/init.pp
evgeni: do you have a preferred way to do it? (I'd guess a `.bundle/config` file is cleaner for a cfgmgmt system)

puppet/modules/redmine/manifests/init.pp
ekohl: You can do a `before` on this package instead of a `requires` on the `ensure_packages`:
```suggestion
    package { 'ruby dnf module':
      before => Package['rubygem-bundler.noarch', 'ruby-devel'],
```

puppet/modules/redmine/manifests/init.pp
ekohl: I'd think `.bundle/config` indeed, but I'd also accept a `TODO` comment here ;)

evgeni: TODO it is!

puppet/modules/redmine/manifests/init.pp
```diff
@@ -100,10 +100,16 @@
     content => template('redmine/secure_config.yaml.erb'),
   }
 
+  file { $app_root:
+    ensure => directory,
+    owner  =>  $username,
```
ekohl, line 117:
```suggestion
    owner  => $username,
```

ekohl (approved):

PR #1883, opened 2023-09-04
allow puma/systemd deployment

evgeni:


puppet/modules/redmine/manifests/init.pp
```diff
@@ -161,10 +162,59 @@
...
+      ensure        => 'present',
+      enable        => true,
+      active        => true,
+      unit_entry    => {
+        'Description'   => 'redmine',
+      },
+      service_entry => {
+        'Type'             => 'notify',
+        'User'             => $username,
+        #'PrivateTmp'      => true,
```
evgeni, line 195: this one needs https://github.com/voxpupuli/puppet-systemd/pull/372

puppet/modules/redmine/manifests/init.pp
```diff
@@ -161,10 +162,59 @@
...
+        'WorkingDirectory' => $app_root,
+        'ExecStart'        => "${app_root}/bin/rails server --environment production",
+      },
+    }
+
+    $apache_backend_config = {
+      'proxy_preserve_host' => true,
+      'proxy_add_headers'   => true,
+      'request_headers'     => ['set X_FORWARDED_PROTO "https"'],
```
ekohl, line 204: In Foreman we also unset SSL client cert parameters. Is that not needed here because Redmine ignores those?

ekohl (commented):
Looks good.

For unix sockets: we did find it was needed to set the hostname: https://github.com/theforeman/puppet-foreman/blob/af5dbe295525aa375c1ecc15225a40e95ef36053/manifests/config/apache.pp#L132

That part is what it supposedly connected to. https://stackoverflow.com/questions/51243332/apache-proxy-pass-to-unix-domain-socket/73741715#73741715 uses `%{HTTP_HOST}` which may actually be better than what we do.

So changing the `proxy_pass` url to `unix:///path/to/socket|http://

puppet/modules/redmine/manifests/init.pp
```diff
@@ -161,10 +162,59 @@
...
+      passenger_start_timeout => 600,
+    }
+  } else {
+    systemd::manage_unit {'redmine.socket':
+      ensure        => 'present',
+      unit_entry    => {
+        'Description' => 'redmine socket',
+      },
+      socket_entry  => {
+        'ListenStream' => 3000,
```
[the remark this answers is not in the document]
ekohl: At least track it as an issue? With Foreman we did see that under load we have connection issues when using a TCP socket on localhost and using unix sockets made those disappear. Given both are Rails apps, it's likely similar.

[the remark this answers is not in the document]
evgeni: Yeah. Maybe using httpd_var_run_t as the context will be enough, will check

puppet/modules/redmine/manifests/init.pp
```diff
@@ -169,17 +169,11 @@
       passenger_start_timeout => 600,
     }
   } else {
-    systemd::manage_unit {'redmine.socket':
+    systemd::unit_file {'redmine.socket':
```
ekohl, line 172: `manage_unit` was not enough?

evgeni: in 6.0.0 it is enough. but the amount of modules I'd need to fix/bump is too high for now.

ekohl: Ah, fair enough.

evgeni: that's what the `make code compatible with systemd 4.2.0` commit is for -- it can be simply reverted once we have 6.0

puppet/modules/redmine/manifests/init.pp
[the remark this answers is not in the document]
evgeni: tracked below as a todo item

puppet/modules/redmine/manifests/init.pp
evgeni: the revert is tracked below as an todo item

ekohl:
> * switch to a unix socket (this will require some selinux mangling, as the socket by default is created as `var_run_t` and Apache can't write to that, so we'll need to make it be created as `httpd_var_run_t`)

There's this context:
```
/var/run/apache.*                                  all files          system_u:object_r:httpd_var_run_t:s0 
```

So would using `/run/apache.redmine.sock` automatically get the correct context? Manual way suggests it would:

```console
# touch /var/run/apache.redmine.sock
# ls -lZ /var/run/apache.redmine.sock
-rw-r--r--. 1 root root unconfined_u:object_r:var_run_t:s0 0 Sep  5 12:41 /var/run/apache.redmine.sock
```

evgeni:
```
type=AVC msg=audit(1693980504.181:4437): avc:  denied  { connectto } for  pid=62541 comm="httpd" path="/run/apache-redmine.sock" scontext=system_u:system_r:httpd_t:s0 tcontext=system_u:system_r:unconfined_service_t:s0 tclass=unix_stream_socket permissive=0
```

Too damn smart thing thing is.

The socket is owned correctly:
```
# ls -alhZ /run/apache-redmine.sock 
srw-------. 1 apache apache system_u:object_r:httpd_var_run_t:s0 0 Sep  6 06:08 /run/apache-redmine.sock
```

But the service is running as `unconfined_service_t` and thus denied by SELinux:
```
[root@redmine ~]# ps auxZ |grep puma
system_u:system_r:unconfined_service_t:s0 redmine 62615 3.3  8.4 555808 170152 ? Ssl  06:10   0:02 puma 6.3.1 (unix:///run/apache-redmine.sock) [redmine]
system_u:system_r:unconfined_service_t:s0 redmine 62621 0.1  8.6 758724 173848 ? Sl   06:10   0:00 puma: cluster worker 0: 62615 [redmine]
system_u:system_r:unconfined_service_t:s0 redmine 62622 0.0  8.1 623492 162284 ? Sl   06:10   0:00 puma: cluster worker 1: 62615 [redmine]
```

ekohl:
I think that's what `SELinuxContext=` is for. Quoting `man systemd.exec`:
```
       SELinuxContext=
           Set the SELinux security context of the executed process. If set, this will override the automated domain transition. However, the policy still needs to authorize the
           transition. This directive is ignored if SELinux is disabled. If prefixed by "-", failing to set the SELinux security context will be ignored, but it's still possible
           that the subsequent execve() may fail if the policy doesn't allow the transition for the non-overridden context. This does not affect commands prefixed with "+". See
           setexeccon(3) for details.
```

The question is then: which domain would make sense.

evgeni:
> The question is then: which domain would make sense.

Yeah, and this is something I'd prefer to think about another day ;)

ekohl:
I'd be good with an issue that summarizes the information about using a unix socket instead of TCP.

evgeni:
Opened #1885 and #1886 for the follow ups

ekohl (approved):

PR #1892, opened 2023-09-09
Allow for Jenkins EL9 node

ehelms:


puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -8,12 +8,13 @@
...
   # TODO: Fix on EL8 and get rid of this
   $is_el8 = $facts['os']['release']['major'] == '8'
+  $is_el9 = $facts['os']['release']['major'] == '9'
 
   package { ['koji', 'rpm-build', 'createrepo', 'copr-cli']:
     ensure => installed,
   }
 
-  unless $is_el8 {
+  unless $is_el8 or $is_el9 {
```
evgeni, line 17: No annex, no pkg builds. While not needed for your task today, we should try using https://git-annex.branchable.com/install/rpm_standalone/ on el8/9.

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -3,6 +3,20 @@
...
   $is_el8 = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8'
+  $is_el9 = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9'
+
+  if $is_el9 {
+    if $facts['os']['name'] == 'CentOS' {
+      yumrepo { 'crb':
+        enabled => '1',
+      }
+    }
+    if $facts['os']['name'] == 'RedHat' {
```
ekohl, line 14: Should this be `case $facts['os']['name']`? And I'll admit that I still often look up https://www.puppet.com/docs/puppet/7/lang_conditional.html for the exact syntax.

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -53,12 +67,12 @@
   ensure_packages([$libxml2_dev, $libxslt1_dev, $libkrb5_dev, $systemd_dev, 'freeipmi', 'ipmitool', $firefox, $libvirt_dev, $libcurl_dev,
   $sqlite3_dev, $libyaml_dev])
 
-  unless $is_el8 {
+  unless $is_el8 or $is_el9 {
     ensure_packages(['python-virtualenv', 'transifex-client'])
```
ekohl, line 71: Slightly unrelated, but `transifex-client` no longer works and can be dropped.

puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -71,7 +72,7 @@
     source => 'puppet:///modules/slave/katello-ca.cert',
   }
 
-  unless $is_el8 {
+  unless $is_el8 or $is_el9 {
     # tito
```
ekohl, line 76: Can't we drop tito by now?

ehelms: Maybe, I think @Odilhao has merged all the Pulpcore PRs to switch away from tito.

puppet/modules/slave/manifests/unittests.pp
ehelms: Should it be? No idea. Could it be? Probably.

puppet/modules/slave/manifests/packaging/rpm.pp
ehelms: There doesn't seem to be any other options I can find other than us packaging it up into our infra repository. Or using Fedora nodes :smile:

puppet/modules/slave/manifests/packaging/rpm.pp
Odilhao: Yes, all PR's are now merged.

ekohl: https://github.com/theforeman/foreman-infra/pull/1895

puppet/modules/slave/manifests/unittests.pp
ekohl: I think it should, because you're essentially doing the same thing. So you could make it determine `$crb_repo_name` and then have a single `yumrepo` statement.

puppet/modules/slave/manifests/packaging/rpm.pp
ehelms: Updated this to use the repo -- take a look

ehelms:
Pulled the git-annex standalone change to https://github.com/theforeman/foreman-infra/pull/1900 in order to allow EL8 based nodes to be used for all workloads similar to EL7 today.

EL9 has a big caveat with NodeJS that complicates things, and while we can/should merge this I'd prefer to pursue moving to EL8 for nodes rather than splitting their usage and wasting resources.

evgeni:
can you elaborate on the caveat? EL8 is EOL next year, so it would be cool if we would be able to deploy EL9 directly *or* at least have a description of the issue we need to solve before that.

ehelms:
> can you elaborate on the caveat? EL8 is EOL next year, so it would be cool if we would be able to deploy EL9 directly _or_ at least have a description of the issue we need to solve before that.
> 
> or is it just the fact that there is only 16+ in the repos?

It's a couple items:

 * NodeJS 16 (default) or NodeJS 18 (via module) are the only NodeJS runtimes available
 * NodeJS 16 does not provide the nodejs-devel package, it's in EPEL
   ** The nodejs and nodejs-devel RPMs can be out of sync due to separate release streams as they are at the time of this writing

evgeni:
This sounds like blockers to deploy Foreman on EL9 too?

ehelms:
Correct, and it is known.

puppet/modules/slave/manifests/packaging/rpm.pp
```diff
@@ -81,7 +82,7 @@
     source => 'puppet:///modules/slave/katello-ca.cert',
   }
 
-  unless $is_el8 {
+  unless $is_el8 or $is_el9 {
```
ekohl, line 85: To keep this PR smaller: https://github.com/theforeman/foreman-infra/pull/1906

ekohl (commented):
I took your PR and created https://github.com/theforeman/foreman-infra/pull/1929 as an alternative which I think is a bit cleaner.

PR #1893, opened 2023-09-11
run the db backup via restic, not as a separate cron job

evgeni:


puppet/modules/profiles/manifests/redmine.pp
```diff
@@ -4,12 +4,47 @@
   # Redmine is already included via ENC
   # TODO include redmine
   $backup_path = '/var/lib/redmine'
+  $backup_db_path = "${backup_path}/db"
```
evgeni, line 7: `restic` can't access `/var/lib/redmine` right now, so we would need to either place the dump somewhere else, or allow restic access, feels odd, both.

puppet/modules/profiles/manifests/redmine.pp
```diff
@@ -4,12 +4,48 @@
...
 
   include profiles::backup::sender
+  include postgresql::server
+  include redmine
+
+  postgresql::server::role { 'restic':
+    password_hash => postgresql::postgresql_password('restic', 'mypasswd'),
```
ekohl, line 14: I wonder if it even needs a password. It's probably going to use ident auth.

puppet/modules/profiles/manifests/redmine.pp
ekohl: We give it CAP_DAC: doesn't that mean it can override it?

puppet/modules/profiles/manifests/redmine.pp
evgeni: it does use ident, yeah, can be dropped

puppet/modules/profiles/manifests/redmine.pp
evgeni: indeed :see_no_evil: 

I was testing this outside systemd, and thus had no CAP mangling

puppet/modules/profiles/manifests/redmine.pp
```diff
@@ -4,12 +4,47 @@
...
+    ensure => directory,
+    owner  => 'restic',
+    mode   => '0600',
+  }
 
   restic::repository { 'redmine':
     backup_cap_dac_read_search => true,
     backup_path                => $backup_path,
     backup_flags               => ['--exclude', "${backup_path}/git"],
+    backup_pre_cmd             => ["pg_dump --file=${backup_db_path}/redmine.sql $redmine::db_name"],
```
ekohl, line 48: Do we need a `backup_post_cmd` to remove it again?

evgeni: nah, it doesn't really hurt being around. IMHO

ekohl, line 48: Lint is unhappy with you
```suggestion
    backup_pre_cmd             => ["pg_dump --file=${backup_db_path}/redmine.sql ${redmine::db_name}"],
```

ekohl (approved):

evgeni:
this has one subtle oddity one should be aware of: the `GRANT … ALL TABLES` is not a permanent thing (as in "it stores a `*` permission"), but is resolved at execution time. this means that if tables (or sequences) are created after this grant command has been executed, the backup won't work until puppet corrects it again on next run.

this is not super bad as such, but can confuse users who try a backup directly after a `db:migrate` has run.

ekohl:
Good to note. It's not going to be a big thing, given Puppet runs twice an hour so this only affects us if the update happens right before the backup runs. Perhaps we should adopt a policy where we backup just before we update.

PR #1894, opened 2023-09-11
run the db backup via restic, not as a separate cron job

evgeni:


puppet/modules/profiles/manifests/redmine.pp
```diff
@@ -6,10 +6,17 @@
...
+  include sudo
+  sudo::conf { "sudo-puppet-backup-redmine":
+    content => "restic ALL=($redmine::username) NOPASSWD: /usr/bin/pg_dump",
+  }
 
   restic::repository { 'redmine':
     backup_cap_dac_read_search => true,
     backup_path                => $backup_path,
     backup_flags               => ['--exclude', "${backup_path}/git"],
+    backup_pre_cmd             => ["/usr/bin/sudo -u $redmine::username pg_dump --file=${backup_path}/redmine.backup.sql $redmine::db_name"],
```
evgeni, line 20: this doesn't work as systemd can't call sudo… :tableflip:

evgeni:
closed in favor of #1893

PR #1898, opened 2023-09-11
migrate redmine params to hiera from foreman

evgeni:


puppet/data/common.yaml
```diff
@@ -209,3 +209,7 @@ puppet::server_puppetserver_telemetry: false
 puppet::show_diff: true
 
 sudo::wheel_config: password
+
+redmine::app_root: /usr/share/redmine
+redmine::data_dir: /var/lib/redmine
```
evgeni, line 214: technically, those two are defaults, but they are overriden today for some reason, and I didn't dare to look further

ekohl: I think that's a bit of a historic thing. We can safely drop them since they do match the defaults.

puppet/data/osfamily/RedHat-9.yaml
```diff
@@ -2,3 +2,5 @@
 redmine::repo_branch: 'new-redmine'
 redmine::deployment: 'puma'
 redmine::servername: 'projects-test.theforeman.org'
+redmine::db_name: redmine
+redmine::username: redmine
```
ekohl, line 6: These also match the defaults, so we can drop them.

evgeni:
updated

puppet/modules/profiles/manifests/redmine.pp
```diff
@@ -1,8 +1,6 @@
 # @summary A Redmine server
 class profiles::redmine (
 ) {
-  # Redmine is already included via ENC
-  # TODO include redmine
   $backup_path = '/var/lib/redmine'
```
ekohl, line 4: My original was:
```puppet
include redmine
$backup_path = $redmine::data_dir
```
Perhaps I should have added that second line as TODO as well, but can you do it now?

ekohl (changes requested):

ekohl (approved):

PR #1902, opened 2023-09-12
post redmine move cleanup

evgeni:


puppet/data/osfamily/RedHat-9.yaml
```diff
@@ -1,3 +1,2 @@
 ---
 redmine::repo_branch: 'new-redmine'
```
ekohl, line 2: While you're at it: remove this and keep the old deployment on some branch like `old-redmine`?

evgeni: yeah…

evgeni: https://github.com/theforeman/redmine/pull/32

ekohl (approved):

PR #1903, opened 2023-09-12
Update restic

evgeni:


ekohl (commented):
Does it make sense to do this for the whole Red Hat OS family?

ekohl (approved):

PR #1907, opened 2023-09-12
add puppetserver to backup

evgeni:


ekohl (commented):
The current backup host is the same machine.

puppet/modules/profiles/manifests/puppetserver.pp
```diff
@@ -14,4 +14,9 @@
   class { 'deploy':
     user => $puppet::server_environments_owner,
   }
+
+  restic::repository { 'puppetserver':
+    backup_cap_dac_read_search => true,
+    backup_path                => '/etc/puppetlabs',
```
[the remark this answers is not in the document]
evgeni: added the four paths that seem useful (doesn't seem to be excessive in size either way)

puppet/data/common.yaml
```diff
@@ -20,6 +20,7 @@ profiles::backup::receiver::targets:
   - redmine01
   - master02
```
ekohl, line 21: Slightly related question: shouldn't this be `controller01` now? I may have messed that up in the Jenkins migration.

evgeni: Most probably it should

PR #1922, opened 2023-09-22
add el9 to spec helper

evgeni:


ekohl:
```
       Evaluation Error: Error while evaluating a Function Call, The nodejs module is not supported on CentOS 9. (file: /home/runner/work/foreman-infra/foreman-infra/puppet/external_modules/nodejs/manifests/params.pp, line: 115, column: 9) on node fv-az389-73.hbj5htltzyauvpykp5onz33k1d.gx.internal.cloudapp.net
```

evgeni:
https://github.com/voxpupuli/puppet-nodejs/pull/479 should fix that (well, and an update here)

evgeni:
:green_apple:

PR #1924, opened 2023-09-22
Change the default Vagrant box to centos/stream9

ekohl:
This is what we aim for in our production and more and more boxes are migrated.


Vagrantfile
```diff
@@ -5,7 +5,7 @@ CENTOS_8_BOX_URL = "https://cloud.centos.org/centos/8-stream/x86_64/images/CentO
 CENTOS_9_BOX_URL = "https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-Vagrant-9-latest.x86_64.vagrant-libvirt.box"
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "centos/7"
+  config.vm.box = "centos/stream9"
```
ekohl, line 8: How to deal with the box override? https://app.vagrantup.com/centos/boxes/stream9 is still empty now.

evgeni (approved):

Vagrantfile
evgeni: if "cleanly" errors if there is no override, so I'd leave it as is for now

PR #1925, opened 2023-09-22
update the key of the backup system to be backup01.netways

evgeni:


evgeni:
the old TTL in DNS was 10800, so we gotta wait a bit, I've just changed it to 300.

evgeni:
`backup01` has currently an ssh key from `root@puppet01` deployed, so I could sync things:
```
rsync -aAHXv --delete /srv/backup/ 193.142.157.176:/srv/backup/
```

evgeni:
(draft so that it won't get accidentally merged w/o DNS update)

evgeni:
DNS was flipped

PR #1929, opened 2023-09-26
Allow for Jenkins EL9 node

ekohl:


puppet/modules/slave/manifests/unittests.pp
```diff
@@ -2,7 +2,21 @@
...
-  $is_el8 = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8'
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9' {
+    case $facts['os']['name'] {
+      'CentOS': {
+        yumrepo { 'crb':
+          enabled => '1',
+        }
+      }
+      'RedHat': {
+        yumrepo { 'codeready-builder-for-rhel-9-x86_64-rpms':
```
ehelms, line 13: One thing to note, this is untested, and probably does not work on AWS because of RHUI. However, I figure we can address that when we actually try to move those nodes.

ehelms (approved):

PR #1930, opened 2023-09-27
cleanup backup::receiver on puppet01

evgeni:


puppet/manifests/site.pp
```diff
@@ -30,7 +30,9 @@
 node /^puppet\d+\.[a-z]+\.theforeman\.org$/ {
   include profiles::base
   include profiles::puppetserver
-  include profiles::backup::receiver
+  class {'profiles::backup::receiver':
+    ensure => absent
```
ekohl, line 34:
```suggestion
    ensure => absent,
```

puppet/modules/profiles/manifests/backup/receiver.pp
```diff
@@ -11,17 +11,20 @@
...
+  Enum['present', 'absent'] $ensure = present,
 ) {
   group { $group:
-    ensure => present,
+    ensure => $ensure,
   }
 
   file { $directory:
-    ensure => directory,
+    ensure => bool2str($ensure == present, 'directory', 'absent'),
```
ekohl, line 27: You need the purge & recurse attributes to clean directories: https://www.puppet.com/docs/puppet/8/type.html#file-attribute-purge

ekohl (changes requested):

puppet/modules/profiles/manifests/backup/receiver.pp
evgeni: Mh, this will make Puppet try to manage the directory, which potentially contains a lot of files.
I'd prefer to make this step manually probably?

ekohl: I also thought about that and think this may be the best way:
```puppet
if $ensure == 'present' {
  file { $directory:
    ensure => directory,
    # current options here
  }
} else {
  file { $directory:
    ensure  => absent,
    recurse => true,
    purge   => true,
  }
}
```

evgeni: and also `force=>true`… the more I think about this the uneasier I find it to allow puppet to purge all our backups

ekohl: Yes, I'd also be fine with leaving the "else" part out and only managing on the "true" part. We rarely remove backups, so manual cleanups are fine.

PR #1931, opened 2023-09-27
puppet module updates

evgeni:


evgeni:
most of the major bumps are dropping Puppet 6 or Debian 9 and somesuch.

Once discourse01 is gone, this should be safe.

evgeni:
> Once discourse01 is gone, this should be safe.

it is gone.

ehelms (approved):
Looks sane to me, as I do not see any failures.

Issue #1844, opened 2023-04-25
Optimize Discourse backup


ekohl:
Right now the Discourse backup contains a large (1G) gzipped tarball with all the image files. I doubt restic can deduplicate those individual files, even though the uploaded files won't change. This will blow up our required storage for backups.

evgeni:
not a problem right now, we can reconsider later

PR #1932, opened 2023-09-28
trigger a backup as pre command instead on relying on discourse cron

evgeni:


ekohl (approved):
I like this. Should we then disable the cron in Discourse itself?

Another thought for a follow up: scan we add extraction of the backup to allow for deduplication (https://github.com/theforeman/foreman-infra/issues/1844)?

evgeni:
> I like this. Should we then disable the cron in Discourse itself?

That was my idea, yes.

> Another thought for a follow up: scan we add extraction of the backup to allow for deduplication (#1844)?

I'd like to defer thinking about this to another day :)

PR #1944, opened 2023-10-02
cleanup buster and bionic references from setup

evgeni:


puppet/modules/slave/manifests/packaging/debian.pp
```diff
@@ -25,24 +25,12 @@
   }
 
   slave::pbuilder_setup {
-    'bionic64':
```
evgeni, line 28: I'll manually purge these from the disk.

ekohl (approved):

PR #1952, opened 2023-10-25
update restic to 0.16.1

evgeni:


ekohl (approved):
Is there a specific reason to upgrade now, or just keeping up to date?

evgeni:
> Is there a specific reason to upgrade now, or just keeping up to date?

Keeping shit updated.
Nothing from the changelog stood out as affecting us.

PR #1958, opened 2023-11-02
Use checksum to detect RPM changes and drop the use of --archive

ehelms:
The --archive option includes --times implicitly that preserves modification time and results in all RPMs being thought of as changed. Instead we explicitly define the options that are included in --archive and drop the use of --times.

We need to apply some of the same rsync logic here to ensure we do not re-copy the RPMs each and every time. As you can see from this test (https://ci.theforeman.org/blue/organizations/jenkins/foreman-nightly-rpm-pipeline/detail/foreman-nightly-rpm-pipeline/2076/pipeline/) pipeline all the files were copied over from stagingyum.

### DIFF
--- a/puppet/modules/web/files/deploy-yumrepo.sh
+++ b/puppet/modules/web/files/deploy-yumrepo.sh
@@ -17,7 +17,7 @@ prepcache() {
 }
 
 do_rsync() {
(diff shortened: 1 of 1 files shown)


ekohl:
With the work that @evgeni did, is this still needed?

evgeni:
No, I think it's not needed anymore

PR #1959, opened 2023-11-06
add renovate

evgeni:


evgeni:
You can see this in action at https://github.com/evgeni/foreman-infra/pulls :)

evgeni:
Instead of using a github action, we could also use the Renovate app (https://github.com/marketplace/renovate) which would not have the token issue, I am happy with either, but I'd like the feature of $thing bumping our deps.

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,20 @@
...
+    - cron: '4 3 * * *'
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Self-hosted Renovate
+        uses: renovatebot/github-action@v39.1.1
```
ekohl, line 15: This will generate a lot of noise. Not too happy with that.

ekohl (changes requested):
Because of the noise in updating the GH action version I'm leaning to the app.

.github/workflows/renovate.yml
evgeni: yepp, they don't have a "rolling" branch :(

evgeni:
Starting setting up the app in https://github.com/theforeman/foreman-infra/pull/1961

evgeni:
closing in favor of https://github.com/theforeman/foreman-infra/pull/1961

PR #1962, opened 2023-11-09
set apache::default_vhost: false globally

evgeni:


puppet/modules/web/manifests/base.pp
```diff
@@ -9,9 +9,7 @@
     include web::letsencrypt
   }
 
-  class { 'apache':
-    default_vhost => false,
-  }
+  class { 'apache': }
```
ekohl, line 12: I'd prefer `include` since that doesn't conflict if there was already a `class { 'apache': }` somewhere.
```suggestion
  include apache
```

ekohl (changes requested):
This will also affect foreman01, but I think that's actually a good thing since the default vhost isn't used there.

ekohl (approved):

PR #1986, opened 2023-11-16
Update yum header to list more URLs

ekohl:
Often I want to quickly get the URL for the nightly foreman-release to test out some changes. This makes it easier. It also documents Katello's release repos. To make it a bit easier to see, the Foreman section is shortened.

### DIFF
--- a/puppet/modules/web/files/yum/HEADER.html
+++ b/puppet/modules/web/files/yum/HEADER.html
@@ -1,11 +1,11 @@
 <h1>yum.theforeman.org</h1>
 
-<h3>Stable Foreman releases</h3>
+<h3>Foreman</h3>
 
-<p>Releases are available under <code>/releases/VERSION/DIST/ARCH</code>, e.g.</p>
+<p>Foreman is available under <code>/releases/VERSION/DIST/ARCH</code>, e.g.</p>
 
 <ul>
-  <li>/releases/3.4/el8/x86_64</li>
+  <li>/releases/3.8/el8/x86_64</li>
   <li>/releases/nightly/el8/x86_64</li>
 </ul>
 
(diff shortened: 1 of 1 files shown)


ehelms (approved):
Pending CI passing

ekohl:
CI is broken. Context is that https://github.com/theforeman/foreman-infra/pull/1985 made CI fail on broken deps, but that should have been there in the first place before we started to merge updates.

@evgeni and I decided that it's easier to roll forward and make sure we update all deps to the latest versions. https://github.com/theforeman/foreman-installer/pull/898 now passes and I also created releases, so we should be ready. We may need to update Foreman from 3.6 to 3.8, though I doubt it since we didn't really make changes.

puppet/modules/web/files/yum/HEADER.html
```diff
@@ -1,29 +1,26 @@
...
 
-<p>A symlink is available at <code>/releases/latest</code> which always points to the latest stable release.  Please be careful when using this, as release upgrades often require some manual intervention (see release notes).</p>
-
-<h3>Nightly builds and release candidates</h3>
-
-<p>Nightly builds of Foreman are available under <code>/nightly/DIST/ARCH</code> and are refreshed a few times a day, but are not GPG signed.</p>
+<p>Nightly builds of Foreman are available under <code>/releases/nightly/DIST/ARCH</code> and are refreshed a few times a day, but are not GPG signed.</p>
 
-<p>Release candidates are in the versioned directories under <code>/releases</code>.  These are GPG-signed like stable packages.</p>
+<p>A symlink is available at <code>/releases/latest</code> which always points to the latest stable release.  Please be careful when using this, as release upgrades often require some manual intervention (see release notes).</p>
```
evgeni, line 23: can we drop this line (and the symlink)? :)

evgeni (approved):

puppet/modules/web/files/yum/HEADER.html
ekohl: Let's push that off to a separate discussion. I just wanted easy links to quickly install and verify https://github.com/theforeman/puppet-katello/pull/478 was no longer needed.

Issue #1989, opened 2023-11-19
redmine:production.log is not rotated and grows indefinetly


evgeni:
```
type=AVC msg=audit(1700092800.224:574348): avc:  denied  { write } for  pid=2124627 comm="logrotate" name="production.log" dev="dm-0" ino=34805887 scontext=system_u:system_r:logrotate_t:s0 tcontext=system_u:object_r:usr_t:s0 tclass=file permissive=0
```

`audit2allow` creates:
```
allow logrotate_t usr_t:file write;
```

but I guess that file just shouldn't be `usr_t` but `var_log_t` or similar.

evgeni:
probably rsyslog can just be dropped given we have journald.

evgeni:
```
[root@redmine01 ~]# systemctl disable --now rsyslog
Removed "/etc/systemd/system/multi-user.target.wants/rsyslog.service".
```

evgeni:
no "no space left on device" events since - closing

PR #1990, opened 2023-11-20
fix SELinux context of Redmine logs so that logrotate works

evgeni:
Fixes: #1989


evgeni:
This turned out not to work as expected anyway :(

evgeni:
Yes, doing so in https://github.com/theforeman/foreman-infra/pull/2067

PR #1993, opened 2023-11-20
Update collection mash script to use 3.9 and 4.11

zjhuntin:


ekohl (commented):
We'll use COPR for 3.9, but https://github.com/theforeman/theforeman-rel-eng/pull/285 hasn't been merged yet so the instructions are a bit out of date. I don't think this is needed.

PR #1995, opened 2023-11-22
Use trailing slash for stagingyum rsync from rsync_cache

ehelms:


Odilhao (approved):

ehelms:
For context, I thought this was another trailing slash issue, but it turned out to be a permission issue that was causing rsync to again for some circumstances. A folder had been created with `root` as the owner and that has been fixed.

PR #2005, opened 2023-12-21
Drop exim module

### LINKED ISSUE #1692: Replace exim Puppet module (or exim)
We currently have our own exim puppet module that could break in the future with newer operating systems. Moving to a shared puppet module may have a longer and better support life.

ekohl:
This module was appled on the old redmine.theforeman.org box, but the new redmine01.conova.theforeman.org no longer uses exim which means it can be dropped.

The question is if we should have some (minimal) postfix config that just ensures it's installed and running. We did see that on Jenkins where that now has been done manually.

Fixes #1692


ekohl:
Rebased now that CI should be green again.

ekohl:
> The question is if we should have some (minimal) postfix config that just ensures it's installed and running. We did see that on Jenkins where that now has been done manually.

That would be https://github.com/theforeman/foreman-infra/pull/1846

ekohl:
I noticed we have exim on web01.

ekohl:
And web01 is gone

evgeni (approved):

puppet/Rakefile
```diff
@@ -11,7 +11,6 @@ PuppetSyntax.exclude_paths = ["vendor/**/*"]
 require 'puppet-lint/tasks/puppet-lint'
 PuppetLint::RakeTask.new :lint do |config|
   config.pattern = PuppetSyntax.manifests_paths
-  config.ignore_paths = ['modules/exim/**/*.pp']
```
ekohl, line 14: Rebased and updated this line to fully get rid of it. Waiting for CI now to see if it's still happy.

PR #2008, opened 2024-01-04
Update latest stable to 3.8

ekohl:


ehelms (approved):

evgeni (approved):

evgeni:
We should just get rid of that symlink

ekohl:
It also updates https://theforeman.org/manuals/latest so we'd need to make sure that's also not referenced anymore.

PR #2013, opened 2024-01-10
Use --force with puppet generate types

ekohl:
When installing a new module it maintains the file timestamps from the original tarball. puppet generate types uses the timestamp of when it ran.

If you update a module to a release that was created before it ran puppet types generate, it will never update the types. This can happen with a module downgrade, or just when you regenerated types manually. By using --force the timestamps are ignored.


evgeni:
I could swear I had already submitted such a PR with exactly that explanation 🙈

evgeni (approved):

ekohl:
https://github.com/puppetlabs/puppet/pull/9206 would be a proper fix in Puppet, but until we can rely on that I think this is better.

ekohl:
I do recall you complaining about it, but I didn't see any PR.

PR #2015, opened 2024-01-16
Add instructions for Jenkins node on EL9

ehelms:


docs/jenkins.md
```diff
@@ -54,6 +54,20 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
 
 ### Configuring a new node
 
+For Enterprise Linux 9:
```
ekohl, line 57: Feels like a lot of repetition. Perhaps we need 2 lines on how to install Puppet and then a common set of instructions to run Puppet.

ehelms: We could, I suspect we will clean this file up real soon and remove all but EL 9.

ekohl: Debian will also remain, so there would at least be 2.

docs/jenkins.md
```diff
@@ -54,6 +54,20 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
...
+
+* Ensure yum.puppet.com is configured: [puppet7-release](https://yum.puppet.com/puppet7-release-el-9.noarch.rpm)
+* `yum -y install puppet-agent`
+* `echo "server = puppet.theforeman.org" >> /etc/puppetlabs/puppet/puppet.conf`
+* ensure hostname is set node0X.jenkins.<provider\>.theforeman.org where <provider\> is osuosl or aws for example and that the record is in DNS
+* Make the `puppet` command available: `source /etc/profile.d/puppet-agent.sh`
+* `puppet ssl bootstrap`
+* Sign the certificate on the puppetmaster or via Foreman
+* `puppet agent --test`
+* Set the host group to "Builders" in Foreman
```
evgeni, line 67: is the "builders" group really needed these days? or should most things come from the `profile::jenkins::node` class?

evgeni: it's needed for the koji certs and user accounts

ehelms: > it's needed for the koji certs and user accounts

Is that all? That would imply we can put an issue to track dropping this step once we drop Koji.

evgeni: that was the only thing I've seen erroring out

ehelms:
Updated to only include EL 9 instructions going forward.

docs/jenkins.md
```diff
@@ -54,35 +54,23 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
...
-* Ensure EPEL is configured: [epel-release](https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm)
-* Ensure yum.puppet.com is configured: [puppet7-release](https://yum.puppet.com/puppet7-release-el-7.noarch.rpm)
-* `yum -y install puppet-agent`
+* `dnf -y install puppet-agent`
 * `echo "server = puppet.theforeman.org" >> /etc/puppetlabs/puppet/puppet.conf`
 * ensure hostname is set node0X.jenkins.<provider\>.theforeman.org where <provider\> is osuosl or aws for example and that the record is in DNS
 * Make the `puppet` command available: `source /etc/profile.d/puppet-agent.sh`
 * `puppet ssl bootstrap`
-* Sign the certificate on the puppetserver or via Foreman
+* Sign the certificate on the puppetmaster or via Foreman
```
ekohl, line 69:
```suggestion
* Sign the certificate on the puppetserver or via Foreman
```

docs/jenkins.md
```diff
@@ -54,35 +54,23 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
...
-* Run `puppet agent --test` twice (second run is important, due to the rvm module behaviour)
+For Enterprise Linux 9, ensure the correct Puppet repository is configured:
 
+* Ensure yum.puppet.com is configured: [puppet7-release](https://yum.puppet.com/puppet7-release-el-9.noarch.rpm)
+* `dnf -y install https://yum.puppet.com/puppet7-release-el-9.noarch.rpm`
 
-For Enterprise Linux 7:
+For Enterprise Linux:
 
-* Ensure EPEL is configured: [epel-release](https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm)
```
ekohl, line 73: Are you sure we no longer need EPEL? Pretty sure we had some dependencies that we relied on. Though I think we should configure EPEL via Puppet if we need it, rather than making it something to bootstrap from.

ehelms: Nothing failed and I did not configure this.

docs/jenkins.md
```diff
@@ -54,25 +54,14 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
...
-* Make the `puppet` command available: `source /etc/profile.d/puppet-agent.sh`
-* `puppet ssl bootstrap`
-* Sign the certificate on the puppetmaster or via Foreman
-* `puppet agent --test`
-* Set the host group to "Builders" in Foreman
-* Run `puppet agent --test` twice (second run is important, due to the rvm module behaviour)
+* Ensure yum.puppet.com is configured: [puppet7-release](https://yum.puppet.com/puppet7-release-el-9.noarch.rpm)
+* `dnf -y install https://yum.puppet.com/puppet7-release-el-9.noarch.rpm`
 
+For Enterprise Linux:
```
evgeni, line 62: This duplicates the heading in line 57

docs/jenkins.md
```diff
@@ -54,25 +54,14 @@ https://github.com/theforeman/foreman-infra/tree/master/puppet/modules contains
 
 ### Configuring a new node
 
-For Enterprise Linux 8:
+For Enterprise Linux 9, ensure the correct Puppet repository is configured:
```
evgeni, line 57: I think I'd drop the version and the "ensure" as that's already in the first step

evgeni (changes requested):

evgeni:
Besides the headings, this seems fine. I used it to setup node03.conova

PR #2016, opened 2024-01-16
Install openssl 1.0 on EL9 Jenkins nodes through RVM

ehelms:
I realize this won't work for our jobs that perform unit tests and still need Ruby 2.7.

```
please read /usr/local/rvm/log/1705408021_ruby-2.5.1/make.log

There has been an error while running make. Halting the installation.
Error: /Stage[main]/Slave::Rvm/Slave::Rvm_config[ruby-2.5]/Rvm_system_ruby[ruby-2.5.1]/ensure: change from 'absent' to 'present' failed: Execution of '/usr/local/rvm/bin/rvm install ruby-2.5.1 ' returned 2: ruby-2.5.1 - #removing src/ruby-2.5.1..
Searching for binary rubies, this might take some time.
No binary rubies available for: centos/9/x86_64/ruby-2.5.1.
Continuing with compilation. Please read 'rvm help mount' to get more information on binary rubies.
Checking requirements for centos.
Requirements installation successful.
Installing Ruby from source to: /usr/local/rvm/rubies/ruby-2.5.1, this may take a while depending on your cpu(s)...
ruby-2.5.1 - #downloading ruby-2.5.1, this may take a while depending on your connection...
ruby-2.5.1 - #extracting ruby-2.5.1 to /usr/local/rvm/src/ruby-2.5.1.....
ruby-2.5.1 - #applying patch /usr/local/rvm/patches/ruby/2.5.1/libressl_2_7.patch.
ruby-2.5.1 - #configuring...................................................................
ruby-2.5.1 - #post-configuration..
ruby-2.5.1 - #compiling.......................................................................................................
Error running '__rvm_make -j8',
please read /usr/local/rvm/log/1705408021_ruby-2.5.1/make.log

There has been an error while running make. Halting the installation.
Notice: /Stage[main]/Slave::Rvm/Slave::Rvm_config[ruby-2.5]/Rvm_alias[ruby-2.5]: Dependency Rvm_system_ruby[ruby-2.5.1] has failures: true
Warning: /Stage[main]/Slave::Rvm/Slave::Rvm_config[ruby-2.5]/Rvm_alias[ruby-2.5]: Skipping because of failed dependencies
Warning: /Stage[main]/Slave::Rvm/Slave::Rvm_config[ruby-2.5]/Exec[ruby-2.5.1/update_rubygems]: Skipping because of failed dependencies
Error: Execution of '/usr/local/rvm/bin/rvm install ruby-2.7.4 ' returned 2: ruby-2.7.4 - #removing src/ruby-2.7.4..
Searching for binary rubies, this might take some time.
No binary rubies available for: centos/9/x86_64/ruby-2.7.4.
Continuing with compilation. Please read 'rvm help mount' to get more information on binary rubies.
Checking requirements for centos.
Requirements installation successful.
Installing Ruby from source to: /usr/local/rvm/rubies/ruby-2.7.4, this may take a while depending on your cpu(s)...
ruby-2.7.4 - #downloading ruby-2.7.4, this may take a while depending on your connection...
ruby-2.7.4 - #extracting ruby-2.7.4 to /usr/local/rvm/src/ruby-2.7.4.....
ruby-2.7.4 - #configuring........................................................................
ruby-2.7.4 - #post-configuration..
ruby-2.7.4 - #compiling......................................................................................................-
Error running '__rvm_make -j8',
please read /usr/local/rvm/log/1705408149_ruby-2.7.4/make.log

There has been an error while running make. Halting the installation.
Error: /Stage[main]/Slave::Rvm/Slave::Rvm_config[ruby-2.7]/Rvm_system_ruby[ruby-2.7.4]/ensure: change from 'absent' to 'present' failed: Execution of '/usr/local/rvm/bin/rvm install ruby-2.7.4 ' returned 2: ruby-2.7.4 - #removing src/ruby-2.7.4..
Searching for binary rubies, this might take some time.
No binary rubies available for: centos/9/x86_64/ruby-2.7.4.
Continuing with compilation. Please read 'rvm help mount' to get more information on binary rubies.
Checking requirements for centos.
Requirements installation successful.
Installing Ruby from source to: /usr/local/rvm/rubies/ruby-2.7.4, this may take a while depending on your cpu(s)...
ruby-2.7.4 - #downloading ruby-2.7.4, this may take a while depending on your connection...
ruby-2.7.4 - #extracting ruby-2.7.4 to /usr/local/rvm/src/ruby-2.7.4.....
ruby-2.7.4 - #configuring........................................................................
ruby-2.7.4 - #post-configuration..
ruby-2.7.4 - #compiling......................................................................................................-
Error running '__rvm_make -j8',
please read /usr/local/rvm/log/1705408149_ruby-2.7.4/make.log

There has been an error while running make. Halting the installation.
```

### DIFF
--- a/puppet/modules/slave/manifests/rvm.pp
+++ b/puppet/modules/slave/manifests/rvm.pp
@@ -17,32 +17,52 @@ class slave::rvm {
   Class['Rvm::System']
 
   if $facts['rvm_installed'] {
+    if $facts['os']['release']['major'] == '9' {
+      exec { 'install openssl 1.1 through rvm':
+        command  => 'rvm pkg install openssl',
+        creates  => '/usr/local/rvm/usr/lib/libssl.so',
+        path     => '/usr/local/rvm/bin:/usr/bin:/bin',
+        provider => 'shell',
+        before   => Slave::Rvm_config['ruby-2.7', 'ruby-3.0'],
+      }
+
+      $build_opts = '--with-openssl-dir=/usr/local/rvm/usr'
+    } else {
+      $build_opts = undef
+    }
+
     rvm::system_user { 'jenkins':
       create  => false,
       require => User['jenkins'],
     }
 
     slave::rvm_config { 'ruby-2.4':
-      version => 'ruby-2.4.3',
+      version    => 'ruby-2.4.3',
+      build_opts => $build_opts,
     }
     slave::rvm_config { 'ruby-2.5':
-      version => 'ruby-2.5.1',
+      version    => 'ruby-2.5.1',
+      build_opts => $build_opts,
     }
     slave::rvm_config { 'ruby-2.6':
-      ensure  => absent,
-      version => 'ruby-2.6.3',
+      ensure     => absent,
+      version    => 'ruby-2.6.3',
+      build_opts => $build_opts,
     }
     slave::rvm_config { 'ruby-2.7':
       version          => 'ruby-2.7.4',
       rubygems_version => '3.1.6',
+      build_opts       => $build_opts,
     }
     slave::rvm_config { 'ruby-3.0':
       version          => 'ruby-3.0.4',
       rubygems_version => '3.2.3',
+      build_opts       => $build_opts,
     }
     slave::rvm_config { 'ruby-3.1':
       version          => 'ruby-3.1.2',
       rubygems_version => '3.3.3',
+      build_opts       => $build_opts,
     }
 
     # Cleanup log dirs
--- a/puppet/modules/slave/manifests/rvm_config.pp
+++ b/puppet/modules/slave/manifests/rvm_config.pp
@@ -3,11 +3,13 @@ define slave::rvm_config (
   String[1] $version,
   Enum['present', 'absent'] $ensure = 'present',
   String[1] $rubygems_version = '3.0.6',
+  Optional[String[1], Undef] $build_opts = undef,
 ) {
   $alias = $title
 
   rvm_system_ruby { $version:
-    ensure => $ensure,
+    ensure     => $ensure,
+    build_opts => $build_opts,
   }
 
   rvm_alias { $alias:


ekohl (changes requested):
Any idea why Ruby 2.7 fails to build? Whats in `/usr/local/rvm/log/1705408149_ruby-2.7.4/make.log`.

ehelms:
And then it gets used like this for installs:

```
/usr/local/rvm/bin/rvm install ruby-3 --with-openssl-dir=/usr/local/rvm/usr
```

Where we are not replacing system openssl, just providing 1.1 for compilation by RVM. The alternatives from what I can tell would be:

 * Stay on EL8 only til this gets resolved for our nodes
 * Swap to running inside a container for our tests in Jenkins

ekohl:
https://bugs.ruby-lang.org/issues/18658 suggests you could install the openssl v3 gem on Ruby 2.7. Interestingly enough, it also mentions Ruby 3.0 not getting the backport. That makes me wonder if RVM with Ruby 3.0 would have problems on EL9 hosts.

evgeni:
Having `compat-openssl11` is not sufficient, right? As it doesn't have `-devel`?

ehelms:
> https://bugs.ruby-lang.org/issues/18658 suggests you could install the openssl v3 gem on Ruby 2.7. Interestingly enough, it also mentions Ruby 3.0 not getting the backport. That makes me wonder if RVM with Ruby 3.0 would have problems on EL9 hosts.

When I tested this on my own, I did hit the same problem with Ruby 3.0. On a Jenkins node, Ruby 3 and 3.1 did not throw any errors.

ehelms:
> Having `compat-openssl11` is not sufficient, right? As it doesn't have `-devel`?

I was not able to use it to success and as you say, couldn't find a devel.

ehelms:
I have updated the code and intent of this PR to install openssl through `rvm pkg`. While this is a bit ugly, I cannot find another viable solution. And while `rvm pkg` is deprecated, the replacement `autolibs` does not solve the problem nor does it provide a replacement method to solve the problem.

evgeni:
I think long-term we should get rid of `rvm` (either by using `rbenv`, which knows that it needs to build OpenSSL for those old Rubies, or by running those tests in dedicated throw-away containers).
But for now I think this is a good enough band aid.

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,6 +17,15 @@
   Class['Rvm::System']
 
   if $facts['rvm_installed'] {
+    if $facts['os']['release']['major'] == '9' {
+      exec { 'install openssl 1.1 through rvm':
+        command  => "rvm pkg install openssl",
+        creates  => "/usr/local/rvm/usr/",
```
evgeni, line 23: can this be a more specific path, something that includes openssl? so that if we ever get into the corner needing another `rvm pkg install`, it won't clash?

ekohl (commented):
I think I saw somewhere that you can provide the openssl directory as a parameter. Do we need to do that for Ruby < 3.1 to use this?

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,6 +17,15 @@
   Class['Rvm::System']
 
   if $facts['rvm_installed'] {
+    if $facts['os']['release']['major'] == '9' {
+      exec { 'install openssl 1.1 through rvm':
```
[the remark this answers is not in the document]
evgeni: > I am worried about this overriding the system openssl, resulting in a platform that doesn't represent a real EL9 system.

It won't represent it either way as EL9 has a patched Ruby/Ruby-OpenSSL to work with OpenSSL 3.0, which you'll only get by using the system Ruby, not something self-compiled.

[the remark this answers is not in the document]
ekohl: That's a good point. Ubuntu 22.04 also ships Ruby 3.0 with OpenSSL 3.0, just like EL9. Testing Ruby 3.0 with OpenSSL 1.1 isn't a real target for us, so isn't upgrading ruby-openssl an option?

Makes me wonder what GHA does.

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,32 +17,52 @@
...
+        command  => "rvm pkg install openssl",
+        creates  => "/usr/local/rvm/usr/lib/libssl.so",
+        path     => '/usr/local/rvm/bin:/usr/bin:/bin',
+        provider => 'shell',
+        before => Slave::Rvm_config['ruby-2.7', 'ruby-3.0']
+      }
+
+      $build_opts = '--with-openssl-dir=/usr/local/rvm/usr'
```
ehelms, line 29: @ekohl This is how you pass along the where it should look for openssl

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,32 +17,52 @@
   Class['Rvm::System']
 
   if $facts['rvm_installed'] {
+    if $facts['os']['release']['major'] == '9' {
+      exec { 'install openssl 1.1 through rvm':
+        command  => "rvm pkg install openssl",
+        creates  => "/usr/local/rvm/usr/lib/libssl.so",
```
ehelms, line 23: @evgeni How's this for being more specific?

evgeni: splendid!

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,32 +17,52 @@
...
-      version => 'ruby-2.5.1',
+      version    => 'ruby-2.5.1',
+      build_opts => $build_opts,
     }
     slave::rvm_config { 'ruby-2.6':
-      ensure  => absent,
-      version => 'ruby-2.6.3',
+      ensure     => absent,
+      version    => 'ruby-2.6.3',
+      build_opts => $build_opts,
```
evgeni, line 50: no need to pass opts when it's absent :)

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,32 +17,52 @@
...
 
     slave::rvm_config { 'ruby-2.4':
-      version => 'ruby-2.4.3',
+      version    => 'ruby-2.4.3',
+      build_opts => $build_opts,
     }
     slave::rvm_config { 'ruby-2.5':
-      version => 'ruby-2.5.1',
+      version    => 'ruby-2.5.1',
+      build_opts => $build_opts,
```
evgeni, line 45: I'd rather remove these instead of trying to build them: https://github.com/theforeman/foreman-infra/pull/2018

evgeni: https://github.com/theforeman/foreman-infra/pull/2019 cleans those up too

puppet/modules/slave/manifests/rvm.pp
```diff
@@ -17,32 +17,52 @@
   Class['Rvm::System']
 
   if $facts['rvm_installed'] {
+    if $facts['os']['release']['major'] == '9' {
+      exec { 'install openssl 1.1 through rvm':
```
evgeni, line 21:
```suggestion
      exec { 'install openssl 1.0 through rvm':
```

evgeni:
@ehelms did you try to run our testsuite on a node built with this? Reading https://github.com/rvm/rvm/issues/5209 it seems that a Ruby built with OpenSSL 1.0 will segfault when trying to use a rubygem-pq that is linked against system OpenSSL (so 3.0) via libpq. I'd prefer not to have also compile PostgreSQL.

Can we spin up a few (RH)EL8 nodes and use those in the meantime?

ekohl:
Perhaps we need to start using labels for available Ruby versions

ehelms:
> Perhaps we need to start using labels for available Ruby versions

I get the thought there, but upon further testing even Ruby 3 via RVM does not want to work reliably on EL 9. That is why I am saying we at least undertake updates to EL 8 and then consider a swap to rbenv instead of RVM.

ekohl:
I have no love for RVM so I don't object to anything else. Even using system Ruby would be fine by me if we can dind a good way to isolate tests (`GEM_HOME` might be an option for that) but the downside is that we currently don't have parallel installable Rubies on RPM and EL9 wouldn't have 2.7 either. Rbenv works well for me but the concept of gemsets that RVM has looks like a separate gem that was questionable in my first impression. Which is why I didn't dare the migration myself yet

On the Puppet side the RVM module (and especially its GPG dependency) is on life support so there I also considered rbenv already

evgeni:
Do we really need gemsets when we use Bundler?

gemsets create an empty environment whenever we run tests, which means we gotta download and compile *all the things* every time.
but what is later used in the process is managed by bundler, which already does enough isolation for the running process?
or is the worry that things get overwritten/corrupted when multiple tests in parallel try to write to ~/.cache/gems etc?

ekohl:
I think that's a worry, but I have no idea how real that worry is. Another is that rubygems < 3.5 doesn't use user installs by default, so if the Ruby is system wide it wants to do nasty things like use sudo etc. Rubygems 3.5+ detects the directory is not writable and uses user mode installs. I wonder if you can teach older bundler versions this.

With the move to GitHub Actions I think we should ask ourselves: how are we going to continue testing in source build pipelines? Do we even need to do that anymore? Can we avoid Ruby testing in Jenkins altogether somehow?

evgeni:
I think setting BUNDLE_PATH (or gem home/path) should fix that.

We only really need the gem testing in Jenkins as part of the nightly source generation/store. We could use the GitHub gem repository for that, but not sure I like that.

ekohl:
Yes, the bundle paths is afaik also what gha does so we can certainly try that

ehelms:
Closed in favor of https://github.com/theforeman/foreman-infra/pull/2022

PR #2019, opened 2024-01-23
drop 2.4 and 2.5 Ruby from RVM, now that it has been cleaned up

evgeni:


evgeni:
(draft until all nodes have performed the cleanup)

ehelms (approved):

PR #2022, opened 2024-01-30
Add rbenv to Jenkins nodes

ehelms:


puppet/modules/slave/manifests/rbenv.pp
```diff
@@ -0,0 +1,24 @@
+# rbenv Configuration
+# @api private
+class slave::rbenv {
+  class { 'rbenv':
+    #owner       => 'jenkins',
+    #group       => 'jenkins',
+    #install_dir => '/home/jenkins/.rbenv',
+    #require     => User['jenkins'],
```
ehelms, line 8: Commented out, as I think this is what we want to do rather than doing it for the whole system as root? The only problem is the error I get right now:

```
==> jenkins-node-el9: Notice: /Stage[main]/Slave::Rbenv/Rbenv::Build[3.1.0]/Exec[git-pull-rubybuild-3.1.0]/returns: fatal: detected dubious ownership in repository at '/home/jenkins/.rbenv/plugins/ruby-build'
==> jenkins-node-el9: Notice: /Stage[main]/Slave::Rbenv/Rbenv::Build[3.1.0]/Exec[git-pull-rubybuild-3.1.0]/returns: To add an exception for this directory, call:
==> jenkins-node-el9: Notice: /Stage[main]/Slave::Rbenv/Rbenv::Build[3.1.0]/Exec[git-pull-rubybuild-3.1.0]/returns: 
==> jenkins-node-el9: Notice: /Stage[main]/Slave::Rbenv/Rbenv::Build[3.1.0]/Exec[git-pull-rubybuild-3.1.0]/returns: 	git config --global --add safe.directory /home/jenkins/.rbenv/plugins/ruby-build
==> jenkins-node-el9: Error: 'git reset --hard HEAD && git pull' returned 128 instead of one of [0]
==> jenkins-node-el9: Error: /Stage[main]/Slave::Rbenv/Rbenv::Build[3.1.0]/Exec[git-pull-rubybuild-3.1.0]/returns: change from 'notrun' to ['0'] failed: 'git reset --hard HEAD && git pull' returned 128 instead of one of [0]
```

I think this is related to how it performs things as root and then uses chown to update owner and group.

puppet/modules/slave/manifests/rbenv.pp
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  ensure_packages(['perl'])
+
+  rbenv::plugin { 'rbenv/ruby-build':
+  }
+  rbenv::build { '3.1.0':
+    env => ['RUBY_CONFIGURE_OPTS=--without-openssl --with-openssl-dir=/usr'],
```
ekohl, line 16: This sounds conflicting: without openssl, but with an openssl option.

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -168,4 +168,9 @@
 
   # RVM
   include slave::rvm
```
ekohl, line 170: Perhaps put this in the `else` below so you either get rvm or rbenv? Perhaps even go further and introduce a parameter that's`$rbenv = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9'`

ekohl (commented):
Can this use `rbenv` and `ruby-build-rbenv` from EPEL? I generally feel more comfortable with packages than git based installations.

I'm tempted to write our own module that only uses packages. https://github.com/theforeman/foreman-infra/pull/2023 is probably that.

puppet/modules/slave/manifests/rbenv.pp
ehelms: I believe it is saying "do not build openssl yourself and use it from over there instead"

puppet/modules/slave/manifests/rbenv.pp
```diff
@@ -0,0 +1,21 @@
...
+  class { 'rbenv':
+    install_dir => '/home/jenkins/.rbenv',
+    require     => [File['/home/jenkins/.rbenv'], Package['gcc-c++']],
+  }
+
+  ensure_packages(['gcc-c++'])
+
+  rbenv::build { '3.1.0': }
+  rbenv::build { '3.0.4': }
+  rbenv::build { '2.7.6': }
```
ehelms, line 20: These are not the most up to date versions of Ruby but our choices are dictated by what's available from ruby-build.

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -168,4 +168,12 @@
...
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '9' {
+    include slave::rvm
+  }
+
+  # rbenv
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '7' {
+    include slave::rbenv
```
ehelms, line 177: Intent is to add this to all our EL8 nodes, switch our code to it and then we can drop RVM and then update to EL9.

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -168,4 +168,12 @@
 
   # RVM
   include slave::rvm
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '9' {
+    include slave::rvm
+  }
```
evgeni, line 173:
```suggestion
  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '9' {
    include slave::rvm
  }
```

Otherwise you always include `rvm` ;)

puppet/modules/rbenv/manifests/build.pp
```diff
@@ -0,0 +1,90 @@
...
+#
+# @param bundler_version
+#   This is used to set a specific version of bundler.
+#
+# @example Install Ruby 2.7.8 as global
+#   rbenv::build { '2.7.8':
+#     global => true,
+#   }
+#
+define rbenv::build (
```
evgeni, line 26: I must admit I find the name of this define odd.
Yes, it's using `ruby-build` under the hood, but the `rbenv` command is `rbenv install` and to be even more precise, I'd probably not use the "action" (build/install) but the "object" (ruby) that we're managing here and call it `rbenv::ruby`.

That all said, this is based on the work in jdowning/rbenv and I have no idea how much we want to diverge from it, naming wise.

puppet/modules/slave/manifests/rbenv.pp
evgeni: "from ruby-build in EPEL", which we can influence, if we want.

Looking at https://canidepend.com/#ruby, we'd want 2.7.8, 3.0.4, 3.1.2 to match EL?

puppet/modules/slave/manifests/unittests.pp
ehelms: Touche

puppet/modules/rbenv/manifests/build.pp
ehelms: No real preference from me

puppet/modules/slave/manifests/unittests.pp
```diff
@@ -167,5 +167,12 @@
   include slave::postgresql
 
   # RVM
-  include slave::rvm
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '9' {
```
evgeni, line 170: This means Debian nodes don't get neither rvm nor rbenv, which is fine as we do not run unittests there, but needs tests fixed

evgeni (approved):
Ack once tests are green

PR #2023, opened 2024-01-30
Add rbenv module

ekohl:
This is based on jdowning/rbenv but reduced to use system packages. Currently a draft since it's untested. I also didn't implement rbenv::gem or a metadata.json file.


ehelms:
> This is based on jdowning/rbenv but reduced to use system packages. Currently a draft since it's untested. I also didn't implement rbenv::gem or a metadata.json file.

I will take this for a test drive.

ehelms:
I used this over here https://github.com/theforeman/foreman-infra/pull/2022

PR #2026, opened 2024-02-19
Jenkins nodes on CentOS should install postgresql-evr

ehelms:


puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -2,59 +2,24 @@
...
       }
+    }
 
-      file { '/usr/bin/pg_config':
-        ensure => 'link',
-        target => '/opt/rh/rh-postgresql12/root/usr/bin/pg_config',
-      }
+    yumrepo { 'foreman-infra-el8':
+      descr    => 'foreman-infra-el8',
+      baseurl  => 'https://yum.theforeman.org/infra/el8/',
```
ehelms, line 15: I am going to have to create this.

ehelms: Actually, I think creating a Copr repository for this purpose makes more sense and is easier?

evgeni: Why not just using the katello repos?

ehelms: I had the same thought and then... why didn't we do that before? I assume we had a "good" reason

ekohl: Because they're in the Pulp repos. This is what we do elsehwere: https://github.com/theforeman/puppet-katello/blob/5585d1114bcd9ad5e04f3bde074d7fcef6498696/spec/setup_acceptance_node.pp#L19-L26

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -2,59 +2,24 @@
...
-        service_name         => 'postgresql',
-        datadir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
-        confdir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
-        bindir               => '/usr/bin',
-      }
-
-      file { '/etc/profile.d/enable_postgresql12_scl.sh':
-        ensure  => absent,
+    ['postgresql-server', 'postgresql-devel', 'postgresql-client', 'postgresql'].each |$pkg| {
+      package { "${pkg}":
```
ekohl, line 6: You can write this as:
```suggestion
      package { ['postgresql-server', 'postgresql-devel', 'postgresql-client', 'postgresql']:
```

But why do we want PostgreSQL to be absent on EL7? Can't we just leave the nodes untouched now?

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -2,59 +2,23 @@
...
-        datadir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
-        confdir              => '/var/opt/rh/rh-postgresql12/lib/pgsql/data',
-        bindir               => '/usr/bin',
-      }
-
-      file { '/etc/profile.d/enable_postgresql12_scl.sh':
-        ensure  => absent,
-      }
+    package { ['postgresql-server', 'postgresql-devel', 'postgresql']:
+      ensure => absent,
```
ekohl, line 6: These are the package names on EL8 so I don't think you can ensure them to be absent. Which is also why I think the test are failing.

ekohl (changes requested):

puppet/modules/slave/manifests/postgresql.pp
ehelms: :facepalm: I failed at reading code

puppet/modules/slave/manifests/postgresql.pp
```diff
@@ -2,59 +2,18 @@
...
-        ensure  => absent,
-      }
-
-      file { '/usr/bin/pg_config':
-        ensure => 'link',
-        target => '/opt/rh/rh-postgresql12/root/usr/bin/pg_config',
-      }
+    # Necessary for PostgreSQL EVR extension
+    yumrepo { 'pulpcore':
+      baseurl  => 'http://yum.theforeman.org/pulpcore/3.39/el\$releasever/x86_64/',
```
ekohl, line 7: I think you don't need to escape it if you use single quotes. If you already use `$releasever, why not also use `$basearch`. Linting often ends up suggesting:
```suggestion
      baseurl  => "http://yum.theforeman.org/pulpcore/3.39/el\$releasever/\$basearch/",
```

ekohl (approved):

PR #2034, opened 2024-02-28
disable zstd compression on Ubuntu Jammy

evgeni:
thanks to Myon and Unit193 from Debian for the previous art in [1]

[1] https://git.postgresql.org/gitweb/?p=pgapt.git;a=blob;f=jenkins/sbuild-update.sh;h=881af551b001a480ba4e18e425c315b88c2a6c3a;hb=HEAD#l133


ehelms (approved):

evgeni:
@ehelms @ekohl should we merge this with the puppet 7.29 hack so that CI can be kept green while puppetlabs figures out a fix?

ehelms:
> @ehelms @ekohl should we merge this with the puppet 7.29 hack so that CI can be kept green while puppetlabs figures out a fix?

Sounds fine to me.. what would be the downside?

evgeni:
"useless entry in the gemfile once the fix is released"

so not too bad (and revertable) :)

PR #2035, opened 2024-02-29
rename slave to jenkins_node

evgeni:


evgeni:
I checked, there are no references to this class (or its parameters) in Foreman itself, everything is in hiera/profiles, which was updated here in the PR.

ehelms (approved):
I've been wanting to see this happen, cheers to you tackling it!

puppet/modules/jenkins_node/manifests/init.pp
```diff
@@ -71,14 +71,14 @@
     ensure => file,
     owner  => 'jenkins',
     group  => 'jenkins',
-    source => 'puppet:///modules/slave/gitconfig',
+    source => 'puppet:///modules/jenkins_node/gitconfig',
```
ekohl, line 74: I guess this is why in the past many people used `${module_name}` instead of hardcoding it. Sometimes people do rename a module.

evgeni: Wait, `${module_name}` is a thing?! TIL

evgeni: https://www.puppet.com/docs/puppet/7/lang_facts_builtin_variables#lang_facts_builtin_variables-compiler-variables

PR #2036, opened 2024-02-29
use nodesource nodistro repos for NodeJS 18 on Jammy

evgeni:


puppet/modules/slave/manifests/packaging/debian.pp
```diff
@@ -36,7 +36,8 @@
       arch       => 'amd64',
       release    => 'bookworm',
       apturl     => $debian_mirror,
-      aptcontent => "deb ${debian_mirror} bookworm main non-free contrib\ndeb-src ${debian_mirror} bullseye main non-free contrib\n";
+      aptcontent => "deb ${debian_mirror} bookworm main non-free contrib\ndeb-src ${debian_mirror} bullseye main non-free contrib\n",
+      nodesource => false;
```
ekohl, line 40: This is because NodeJS on Debian 12 is new enough so we can use the distro version?

ekohl (approved):

puppet/modules/slave/manifests/packaging/debian.pp
evgeni: commit message says "disable nodesource for bookworm, it ships with NodeJS 18 already", so yes :)

PR #2059, opened 2024-05-06
drop stagingyum from rsync config

evgeni:


ekohl (commented):
This keeps the existing server running. Is that intentional and will you drop it when it's migrated to a new server?

evgeni:
> This keeps the existing server running. Is that intentional and will you drop it when it's migrated to a new server?

Yes, because the "server" also serves other paths.

But I read https://github.com/puppetlabs/puppetlabs-rsync/blob/main/manifests/server/module.pp in a way that the concat fragment is now gone and thus it gets removed from rsyncd.conf

puppet/modules/profiles/manifests/web.pp
```diff
@@ -62,7 +62,6 @@
   contain web::vhost::yum
 
   class { 'web::vhost::stagingyum':
```
ekohl, line 64: If you don't set parameters, the `contain` below is sufficient because it implies including the class. Note we do the same with `web::vhost::stagingdeb`.

ekohl (approved):

PR #2066, opened 2024-05-08
drop rsync from web

evgeni:


evgeni:
I've stopped xinetd and disabled puppet agent on web01 until this is merged.

https://community.theforeman.org/t/sunsetting-rsync-theforeman-org/37884/2

puppet/modules/profiles/manifests/web.pp
```diff
@@ -52,7 +42,6 @@
 
   class { 'web::vhost::yum':
     stable                => $stable,
```
ekohl, line 44: Perhaps align it again?
```suggestion
    stable => $stable,
```

puppet/modules/jenkins_node/manifests/packaging/rpm.pp
```diff
@@ -47,8 +47,6 @@
     manage_dir => true,
   }
 
-  include rsync
```
ekohl, line 50: Don't we still need to ensure the `rsync` package is present? While we can count on the utility, I wouldn't mind it being explicit if we rely on it.

ekohl (approved):
2 minor comments, but :+1: overall.

puppet/modules/jenkins_node/manifests/packaging/rpm.pp
evgeni: I was under the impression https://github.com/theforeman/foreman-infra/pull/2066/files#diff-c9be9d7252cac9998dce62c020c7a0b220cf5e1c6d95c511ca9484127b3df492 would handle that.

puppet/modules/profiles/manifests/web.pp
evgeni: qouting reviewer 2: "muss das sein!?"

(ok, I'll do it)

puppet/modules/jenkins_node/manifests/packaging/rpm.pp
evgeni: done

PR #2069, opened 2024-05-13
Update next version for 3.12

Griffin-Sullivan:
Now that we are branching we need to update $next.


evgeni:
Shall we instead get https://github.com/theforeman/foreman/pull/10155 done? One step less during branching :)

Griffin-Sullivan:
Whatever works for me :man_shrugging:

evgeni:
This got merged. And so did https://github.com/theforeman/foreman-infra/pull/2082

No more `$next`! :tada: 

Can you update the release procedure and drop this step too?

PR #2070, opened 2024-05-14
drop rvm module

evgeni:


puppet/Puppetfile
```diff
@@ -2,7 +2,6 @@ moduledir 'external_modules'
 
 mod 'citrin/htpasswd', '1.0.2'
 mod 'claranet/timezone', '2.0.1'
-mod 'golja/gnupg', '1.2.3'
```
evgeni, line 5: :fire:

puppet/modules/jenkins_node/manifests/unittests.pp
```diff
@@ -164,11 +164,6 @@
   # Databases
   include jenkins_node::postgresql
 
-  # RVM
-  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '9' {
-    include jenkins_node::rvm
-  }
-
   # rbenv
   if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '7' {
```
ekohl, line 168: Unrelated, but this can be simplified too now. But I see that approach can be taken wider so let's not include it here.

ekohl (approved):

puppet/modules/jenkins_node/manifests/unittests.pp
evgeni: I think we should do a sweeping cleanup once we migrated the CI nodes to EL9, then we can remove all el7 and 8 specialcases.

PR #2071, opened 2024-05-14
use NodeJS 18.x on EL9 CI nodes

evgeni:
Requires
- [x] https://github.com/theforeman/jenkins-jobs/pull/463
- [ ] Foreman 3.9 to be EOL


ehelms:
Do we know for certain the tests will fail with 18? We could try it :)

evgeni:
Webpack 3 doesn't like the new Node afaik

evgeni:
@ekohl @ehelms need a few brains here…

* We have two kinds of Jenkins nodes: those labeled `fast` and those not labeled `fast`
* Only `fast` ones get the heavy NodeJS involved jobs
* We do not differentiate in Puppet, so both kinds get NodeJS installed
* My idea was to keep the fast ones on EL8 until 3.9 is ded, and upgrade the "slow" ones already.
* This can be done using this PR, but…

We do not need NodeSource anymore. CentOS has new enough NodeJS. And someone even made `puppet-nodejs` support DNF modules in https://github.com/voxpupuli/puppet-nodejs/pull/488 (which is in 11.x) ;)
However, I am sure that flipping the *existing* setup to DNF modules would be a disaster.

How crazy would be to just stop managing NodeJS on EL8 completely, and instead use modules on EL9?
The existing Jenkins nodes would be unaffected and new ones on EL9 "cleaner".

ehelms:
> * Only `fast` ones get the heavy NodeJS involved jobs

I wonder if the reasoning behind this still holds true.



> We do not need NodeSource anymore. CentOS has new enough NodeJS. And someone even made `puppet-nodejs` support DNF modules in [voxpupuli/puppet-nodejs#488](https://github.com/voxpupuli/puppet-nodejs/pull/488) (which is in 11.x) ;) However, I am sure that flipping the _existing_ setup to DNF modules would be a disaster.
> 
> How crazy would be to just stop managing NodeJS on EL8 completely, and instead use modules on EL9? The existing Jenkins nodes would be unaffected and new ones on EL9 "cleaner".

Fine by me.

evgeni:
> > * Only `fast` ones get the heavy NodeJS involved jobs
> 
> I wonder if the reasoning behind this still holds true.

It does. OSUOSL nodes are significantly slower than Conova/Netways.

ehelms:
> It does. OSUOSL nodes are significantly slower than Conova/Netways.

I more meant to question whether our NodeJS testing has gotten faster as we use newer node and we've re-architected how it all gets built.

evgeni:
it's even an upgrade of the Node version, as the old repo is not updated anymore:
```
[vagrant@jenkins-node-el9 ~]$ node --version
v18.17.1

[vagrant@jenkins-node-el9 ~]$ node --version
v18.19.1
```

evgeni:
> > It does. OSUOSL nodes are significantly slower than Conova/Netways.
> 
> I more meant to question whether our NodeJS testing has gotten faster as we use newer node and we've re-architected how it all gets built.

Ah, well, it was not sure. I recall the original split was because of Ruby, not NodeJS and all tests run on the same jenkins node once they start?

puppet/modules/jenkins_node/manifests/unittests.pp
```diff
@@ -70,9 +70,10 @@
   ensure_packages(['python3-virtualenv'])
 
   # nodejs/npm for JavaScript tests
-  if $facts['os']['family'] == 'RedHat' {
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9' {
     class { 'nodejs':
-      repo_url_suffix       => '14.x',
+      repo_version          => '18',
+      repo_class            => 'nodejs::repo::dnfmodule',
       nodejs_package_ensure => latest,
```
ekohl, line 77: If we rely on RPM packages, can we change this to `present` and instead rely on the `dnf.timer` to update?

ekohl (approved):
3.9 is about to go EOL, but I'd also wait for all 3.9 jobs to be dropped from jenkins.jobs. Other than that, :+1:

evgeni:
> 3.9 is about to go EOL, but I'd also wait for all 3.9 jobs to be dropped from jenkins.jobs. Other than that, 👍

This is scoped to EL9 only, so it won't affect *current* nodes anyway, but yeah.

puppet/modules/jenkins_node/manifests/unittests.pp
ekohl: The reason I say this is that `latest` will purge the dnf cache on every run and that's putting a higher load on mirrors than needed.

evgeni: Fair, I can update that on Monday

ekohl: On a related note: I recall this because it used to be an extremely easy way to overload Satellite.

evgeni:
We declared 3.9 dead, so I am merging this

PR #2078, opened 2024-05-17
Remove EL7 compatibility from jenkins_node module

ekohl:


puppet/modules/jenkins_node/manifests/packaging/rpm.pp
```diff
@@ -69,20 +62,6 @@
     ensure => present,
```
evgeni, line 62: the line above contains `scl-utils`, which we don't need on el8+ either.

evgeni (approved):

puppet/modules/jenkins_node/manifests/packaging/rpm.pp
ekohl: I also thought about that, but wasn't sure if we should ensure it's absent first.

evgeni: We're about to re-deploy all nodes with el9, so I wouldn't care too much.

PR #2083, opened 2024-05-17
drop jjb from Puppet, it's completely inside jenkins these days

evgeni:


puppet/data/common.yaml
```diff
@@ -13,9 +13,6 @@ foreman_proxy::trusted_hosts:
   - "foreman01.conova.theforeman.org"
 foreman_proxy::foreman_base_url: '%{alias("foreman_url")}'
 
-# Foreman's jenkins-jobs is incompatible with JJB 5
-jenkins_job_builder::ensure: '4.3.0'
```
evgeni, line 17: this has been broken, ups :)

puppet/modules/jenkins_job_builder/manifests/install.pp
```diff
@@ -1,18 +0,0 @@
...
-  $yaml = if $facts['os']['release']['major'] == '7' { 'PyYAML' } else { 'python3-pyyaml' }
-  ensure_packages(['python-pip', $yaml])
-
-  package { 'jenkins-job-builder':
-    ensure   => $jenkins_job_builder::ensure,
-    provider => 'pip',
-    require  => Package['python-pip', $yaml],
-  }
-
-  file { '/etc/jenkins_jobs':
```
evgeni, line 15: given it's just one system, I'd clean this up manually

puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -1,44 +0,0 @@
...
-    ensure   => latest,
-    provider => git,
-    source   => $git_repo,
-    revision => $git_ref,
-    notify   => Exec["jenkins-jobs-update-${config_name}"],
-    require  => Package[$git],
-  }
-
-  exec { "jenkins-jobs-update-${config_name}":
-    command     => "jenkins-jobs --conf ${inifile} update ${directory}/${git_project_name}/${config_name} ${command_arguments}",
```
evgeni, line 30: technically this is useful for bootstrapping, but on the other hand when bootstrapping there is no user to perform this so… chicken

ekohl (commented):
We had this to bootstrap jobs. Is there a better way to achieve this?

evgeni:
That's what I thought at first too. But there is nothing that creates a user in Jenkins that could do the job deployment?

ekohl:
Last time I tried in the Puppet module for Jenkins created an instance that doesn't require a user and auth, so it worked anonymous

PR #2092, opened 2024-06-02
drop debugs

evgeni:


evgeni:
includes #2066

puppet/modules/web/manifests/vhost/debugs.pp
```diff
@@ -1,20 +0,0 @@
...
-  web::vhost { 'debugs':
-    docroot_owner => 'nobody',
-    docroot_group => 'nobody',
-    docroot_mode  => '0755',
-    attrs         => {
-      'custom_fragment' => template('web/debugs.conf.erb'),
-    },
-  }
-
-  $htpasswds.each |$username, $options| {
```
ekohl, line 15: Since this is removed, `web::htpasswd` is unused and `puppet/modules/web/manifests/htpasswd.pp` can be dropped. It also means we can drop the related bits (`citrin/htpasswd` in `Puppetfile` and `puppet/spec/defines/web_htpasswd_spec.rb`).

puppet/modules/awstats/manifests/init.pp
ekohl, line 1: If we don't host awstats anymore, is the class still of any use?

evgeni: it is not :)

ekohl (approved):
We had these RFCs in the past:
https://community.theforeman.org/t/rfc-deprecate-foreman-debug/13361
https://community.theforeman.org/t/rfc-deprecate-foreman-debug-in-favor-of-sosreport/12465

I can't find a concrete deprecation planning and would like to see such  post, tying this all back. In addition to that, we can drop the `debug` subpackage in `foreman.spec` and probably the Debian equivalent.

ehelms (approved):

PR #2093, opened 2024-06-03
Use Python 3.12 for Ansible

evgeni:


ekohl (approved):

evgeni:
(I've manually installed the pkg from koji, as CS8 mirrors won't give it to me anymore)

PR #2098, opened 2024-06-28
Upgrade to Foreman 3.11

evgeni:


ekohl (approved):
I think 3.6 -> 3.11 will work fine. If db:migrate works (and I don't see why it wouldn't) then I'd trust it. Just take a backup before hand.

PR #2106, opened 2024-07-09
drop EL8 support from our codebase

evgeni:


puppet/modules/utility/manifests/init.pp
```diff
@@ -7,9 +7,7 @@
 
   stdlib::ensure_packages([$vim])
 
-  unless $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8' {
-    stdlib::ensure_packages(['htop', 'iftop', 'screen'])
-  }
+  stdlib::ensure_packages(['htop', 'iftop', 'screen'])
```
ekohl, line 10: `screen` is back in EL9?

puppet/data/osfamily/RedHat-8.yaml
```diff
@@ -1,11 +0,0 @@
----
-apache::default_mods:
-  - status
-apache::protocols:
-  - 'h2'
-  - 'h2c'
-  - 'http/1.1'
-apache::mpm_module: event
```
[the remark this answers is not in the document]
evgeni: oh, I see

puppet/modules/utility/manifests/init.pp
evgeni: No. Which makes me wonder if the `utility` class is properly applied, as the EL9 boxen do not drop any errors but that should have been applied there already?

ekohl: There's a very good chance I somehow missed this. It's supposed to be applied via the Foreman ENC because we don't have the admins in Hiera (for privacy reasons). https://github.com/theforeman/foreman-infra/pull/1789 was intended to solve that but as you can see, that's been open for a while.

puppet/data/osfamily/RedHat-8.yaml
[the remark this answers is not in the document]
ekohl: https://github.com/theforeman/foreman-infra/pull/2109

puppet/modules/utility/manifests/init.pp
evgeni: We could add `utility` to the `base` config group, but that'd mean we need EPEL on all (EL) systems, which we today don't do.

ekohl (approved):

PR #2108, opened 2024-07-19
add dedicated debian repo profile

evgeni:


puppet/modules/freight/manifests/init.pp
```diff
@@ -1,13 +1,12 @@
...
-  if $facts['os']['family'] == 'Debian' {
-    apt::source { 'freight':
-      location => 'http://build.openvpn.net/debian/freight_team',
-      repos    => 'main',
-      key      => {
-        id     => '30EBF4E73CCE63EEE124DD278E6DA8B4E158C569',
-        source => 'https://swupdate.openvpn.net/repos/repo-public.gpg',
-      },
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9' {
+    yumrepo { 'evgeni_freight':
```
ekohl, line 4: Is there something to request it in EPEL 9 proper?

evgeni: I did not follow https://docs.fedoraproject.org/en-US/epel/epel-package-request/ yet to request it.

evgeni: https://bugzilla.redhat.com/show_bug.cgi?id=2299182

puppet/modules/profiles/manifests/repo/deb.pp
```diff
@@ -0,0 +1,24 @@
+# @summary A profile for the debian repo machines
+#
+# @param https
+#   Whether to enable HTTPS. This is typically wanted but can only be enabled
+#   in a 2 pass setup. First Apache needs to run for Letsencrypt to function.
+#   Then Letsencrypt can be enabled. Also useful to turn off in test setups.
+class profiles::repo::deb (
+  Boolean $https = true,
+  String[1] $stable = '3.11',
```
evgeni, line 9: this duplicates the data in profiles::web, not sure how to best externalize this information

ekohl: We have the same problem with RPM repos. In think the best solution is to introduce a key in Hiera and then use aliases. For example, here I introduced some generic keys (that don't map to any class) and then use them in other entries:
https://github.com/theforeman/foreman-infra/blob/7b13ffba6ffef521b9871803dfca5ae13aed8e0a/puppet/data/common.yaml#L2-L14

evgeni: https://github.com/theforeman/foreman-infra/pull/2119

puppet/modules/freight/manifests/init.pp
```diff
@@ -1,15 +1,8 @@
...
-      key      => {
-        id     => '30EBF4E73CCE63EEE124DD278E6DA8B4E158C569',
-        source => 'https://swupdate.openvpn.net/repos/repo-public.gpg',
-      },
-      before   => Package['freight'],
-    }
+  class {'epel':
+    epel_testing_enabled => true,
+    before               => Package['freight'],
   }
```
evgeni, line 6: What's our policy on EPEL, should modules explicitly add `include epel` (or whatever) if they use packages from there, or can we assume it to be present?

(In the past the freight module assumed it)

puppet/modules/web/manifests/base.pp
```diff
@@ -10,6 +10,9 @@
   }
 
   include apache
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9' {
+    include apache::mod::expires
+  }
```
evgeni, line 15: Can be dropped once we have https://github.com/puppetlabs/puppetlabs-apache/pull/2559

evgeni:
marked this as ready as it spins up a working el9-freight host for me and should not influence anything of the current web01 deployment.

puppet/modules/freight/manifests/init.pp
```diff
@@ -1,14 +1,9 @@
...
-      location => 'http://build.openvpn.net/debian/freight_team',
-      repos    => 'main',
-      key      => {
-        id     => '30EBF4E73CCE63EEE124DD278E6DA8B4E158C569',
-        source => 'https://swupdate.openvpn.net/repos/repo-public.gpg',
-      },
-      before   => Package['freight'],
+  if $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '9' {
+    class {'epel':
+      epel_testing_enabled => true,
```
ekohl, line 5: Note to self: once https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2024-87076f97c3 is in then we can drop this line.

puppet/modules/freight/manifests/user.pp
```diff
@@ -73,7 +73,7 @@
     owner   => 'root',
     group   => 'root',
     mode    => '0644',
-    content => epp("${module_name}/${vhost}-HEADER.html.epp", { 'stable' => $profiles::web::stable }),
+    content => epp("${module_name}/${vhost}-HEADER.html.epp", { 'stable' => $web::vhost::deb::stable }),
```
ekohl, line 76: It feels a bit better to make it an explicit parameter rather than refer to some random other class variable.

puppet/modules/web/manifests/base.pp
ekohl: On line 61 in puppet/modules/freight/manifests/user.pp it mentions the headers module, but I think that should have been `apache::mod::expires`.

Also, why is this limited to EL9? I'd think this is needed everywhere.

puppet/modules/freight/manifests/init.pp
ekohl: If it depends on EPEL, it should include it. Preferably like this so it can be done in all modules that use EPEL.
```puppet
include epel
Class['epel'] -> Package['freight']
```

puppet/modules/web/manifests/base.pp
evgeni: Needed? Yes. But on EL7 we have default mods enabled, and then you get duplicate definitions blah blah

evgeni: https://github.com/theforeman/foreman-infra/actions/runs/10596140550/job/29363435733

evgeni: Ah, probably because e2a182031f9f448f01f7fa262915936815d45108 changed `Header` to `use mod_expire`!

puppet/modules/freight/manifests/user.pp
```diff
@@ -57,8 +58,12 @@
...
   ]
 
-  # locations doesn't autorequire the headers module
-  include apache::mod::headers
+  # vhosts don't autorequire the expires module
+  # https://github.com/puppetlabs/puppetlabs-apache/pull/2559
+  # limit to not EL7 as there we use apache::default_mods
+  if $facts['os']['family'] != 'RedHat' or $facts['os']['release']['major'] != '7' {
+    include apache::mod::expires
+  }
```
evgeni, line 66: Or we move that to
https://github.com/theforeman/foreman-infra/blob/78d1281d0ee9c1326d6939d1cea993ef74e3e970/puppet/data/osfamily/RedHat-9.yaml#L2-L3

puppet/modules/freight/manifests/init.pp
ekohl: It should go stable in 2 days. Given all the releases that are ongoing I'd prefer to wait for that and avoid introducing this line.

puppet/modules/freight/manifests/user.pp
```diff
@@ -5,6 +5,7 @@
   Stdlib::Absolutepath $stagedir,
   String $vhost,
   Variant[String, Array[String]] $cron_matches,
+  Optional[String[1]] $stable = undef,
```
evgeni, line 8: I moved this to https://github.com/theforeman/foreman-infra/pull/2120 as it's not directly related to this PR

puppet/modules/freight/manifests/init.pp
evgeni: absolutely. I just have it here so that my local env works.

puppet/modules/freight/manifests/user.pp
```diff
@@ -58,8 +58,16 @@
...
+  # vhosts don't autorequire the expires module
+  # https://github.com/puppetlabs/puppetlabs-apache/pull/2559
+  # limit to not EL7 as there we use apache::default_mods
+  if $facts['os']['family'] != 'RedHat' or $facts['os']['release']['major'] != '7' {
+    include apache::mod::expires
+  }
+  include apache::mod::alias
+  include apache::mod::autoindex
+  include apache::mod::dir
+  include apache::mod::mime
```
evgeni, line 70: I wonder if any of these should (or should not) go to our definition of `apache::default_mods`

evgeni:
@ehelms @ekohl there are cosmetic questions about `apache::mod::*` in here, but otherwise this is working and we should move on :)

ehelms (approved):
I'm good with this - let's do this.

PR #2119, opened 2024-08-29
use a dedicated data entry for stable release

evgeni:


ekohl (approved):
We'll also need to update https://github.com/theforeman/theforeman-rel-eng/blob/8805662ac8310acc7c8fb146b82de91d49ee78d2/procedures/foreman/release.md.erb#L154 to point here.

evgeni:
https://github.com/theforeman/theforeman-rel-eng/pull/418

PR #2124, opened 2024-09-12
add backend serveraliases to debian vhosts

evgeni:
deb.tfm.o (and friends) do not point to our host these days, but to a
CDN, thus, using deb.tfm.o as the vhost name is a tad wrong.

this change introduces two aliases for the vhost: deb-backend.tfm.o and
deb.${fqdn}, which in my test environment results in:

    ServerName deb.theforeman.org
    ServerAlias deb-backend.theforeman.org
    ServerAlias deb.repo-deb.tanso.example.com

This has the benefit that the right vhost is reachable without any
tricks, and will allow us to switch the CDN config to a "more correct"
naming scheme, later dropping deb.tfm.o from the vhost here totally.

(This has the side-benefit that deb.tfm LE requests go via the CDN and
only then hit our box, which is confusing to say the least and that
would stop)

### DIFF
--- a/puppet/modules/freight/manifests/user.pp
+++ b/puppet/modules/freight/manifests/user.pp
@@ -70,6 +70,8 @@ define freight::user (
   include apache::mod::mime
 
   web::vhost { $vhost:
+    servername    => "${vhost}-backend.${facts['networking']['fqdn']}",
+    serveraliases => [ "${vhost}.theforeman.org" ],
     docroot       => $webdir,
     docroot_owner => $user,
     docroot_group => $user,


puppet/modules/freight/manifests/user.pp
```diff
@@ -70,6 +70,7 @@
   include apache::mod::mime
 
   web::vhost { $vhost:
+    serveraliases => [ "${vhost}-backend.theforeman.org", "${vhost}.${facts['networking']['fqdn']}" ],
```
evgeni, line 73: gotta add the right DNS entries before deploying this, or LE will freak out

ehelms:
I am guessing the same will need to happen for the rpm serving and thus I should copy this model?

evgeni:
Need? No.
Would I like to have the same layout there? Yes :)

puppet/modules/freight/manifests/user.pp
ekohl: Can we make the servername `deb-backend`? I don't mind adding the `$vhost.$fqdn` as an alias.

evgeni: no, because then stagingdeb and archivedeb will also have this alias, which is not what I wanted :)

ekohl: So we can't make `$vhost` `deb-backend` instead?

My reasoning is that I'd like to use `mod_md` to get the certificate (based on the `servername`) instead of the current letsencrypt module we use, but we don't have to pull that in scope here.

evgeni: We have 3 instances of `freight::user`: `main`, `archive` and `staging` which create `deb`, `archivedeb` and `stagingdeb` vhosts.
We should switch those to produce `deb-backend`, `archivedeb-backend` and `stagingdeb-backend` at some point, yes.
But I first wanted to introduce the alias, have time to play with the CDN and *then* drop the non-`backend` vhosts.

evgeni, line 73: I thought a bit more about this, and here is a tiny braindump:
* Today we configure the CDN to talk to `web01` and set the `Host` header to `deb.theforeman.org`, etc to request the "right" content.
* I *want* the CDN to talk to `deb.<the backend machine>` (and `stagingdeb.…`, etc) as the end goal, as this allows me to
  * select the backend at the CDN level (no DNS involved, as today)
  * drop `deb.tfm.o` etc DNS names from the cert `<the backend machine>` has (as those do not point at that machine directly)
* This means that as the end result, all I need are `${archive}.${backend}` vhosts (so `deb.repo-deb01.osuosl.theforeman.org`, etc)
* To ease the transition, it's beneficial to keep `deb.tfm.o` and similar aliases until the CDN is reconfigured and verified.

TL;DR:

Step 1:
```puppet
 web::vhost { $vhost:
  servername => "${vhost}.${facts['networking']['fqdn']}",
  serveraliases => [ "${vhost}.theforeman.org" ],
  …
}
```

reconfigure CDN

Step 2:
```puppet
 web::vhost { $vhost:
  servername => "${vhost}.${facts['networking']['fqdn']}",
  …
}
```

(need to set `servername` explicitly as `web::vhost` otherwise uses `"${title}.theforeman.org"`)

ehelms: According to your step 1, you should be setting `servername` ?

evgeni: > According to your step 1, you should be setting `servername` ?

In this PR? Yes.

ehelms: I think I misunderstood -- your proposal looks good to me. I think you can update the PR to reflect that.

puppet/modules/freight/manifests/user.pp
```diff
@@ -70,6 +70,8 @@
   include apache::mod::mime
 
   web::vhost { $vhost:
+    servername    => "${vhost}.${facts['networking']['fqdn']}",
```
evgeni, line 73: I have created `*.web01.osuosl` and `*.repo-deb01.osuosl` in DNS.

ekohl: I'd prefer `-backend`, even though you're right it probably doesn't matter.

Now a question is: should it be the servername or an alias.  I have a vague recollection that the servername is used in built in error pages. https://httpd.apache.org/docs/2.4/mod/core.html#servername at least suggests that it should be the name end users connect to.

evgeni: But the long term idea is not to ever have the name the user connects to in this config.

And we (should) use ServwrSignature Off (at least right now we do), so there is no footer and anything

ekohl: https://www.fastly.com/documentation/guides/integrations/backends/#overriding-the-host-header states:

> If you use a hostname (rather than an IP address) to define your backend, Fastly will only use the hostname to look up the IP address of the server, not to set the [Host](https://www.fastly.com/documentation/reference/http/http-headers/Host/) header or negotiate a secure connection. By default the [Host](https://www.fastly.com/documentation/reference/http/http-headers/Host/) header on backend requests is copied from the client request. For example, if you own www.example.com and point it to Fastly, and create a second domain of origin.example.com that points to your origin, then the web server running on your origin and serving origin.example.com must also be able to serve requests that have a Host: www.example.com header. Fastly will also use the client-forwarded hostname to establish a secure connection using [Server Name Indication](https://en.wikipedia.org/wiki/Server_Name_Indication).

So we need to explicitly configure it in a different way. Are we doing that?

evgeni: Today, we set `address`:
https://github.com/theforeman/foreman-infra/blob/85293f48404e0ddcafbad0e982d9c5a8a49f24c7/ansible/fastly.yml#L17
And `ssl_*`:
https://github.com/theforeman/foreman-infra/blob/85293f48404e0ddcafbad0e982d9c5a8a49f24c7/ansible/fastly.yml#L20-L21
So we'd have to set `override_host` too.

evgeni: https://github.com/theforeman/foreman-infra/pull/2153

ekohl (approved):

PR #2126, opened 2024-09-17
Split out letsencrypt for web

ehelms:
We need to split out letsencrypt certificate setup since we are splitting up our web content across multiple servers. In the current format, this will fail to setup rpm and debian content servers. I am aiming to introduce a reusable defines to solve this but it does present a conundrum still. In theory the various vhosts are independent, but the certificate that gets generated is tied to the machine so that we have a single entity but requires knowledge of the vhosts.

Should I instead be doing this at the profile level? Is that leaking internals to the profile?

### DIFF
--- a/puppet/modules/profiles/manifests/repo/rpm.pp
+++ b/puppet/modules/profiles/manifests/repo/rpm.pp
@@ -12,7 +12,8 @@ class profiles::repo::rpm (
   Boolean $https = true,
 ) {
   class { 'web':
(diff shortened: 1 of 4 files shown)


evgeni:
How about https://github.com/theforeman/foreman-infra/pull/2129

ehelms:
https://github.com/theforeman/foreman-infra/pull/2129 looks good to me!

PR #2129, opened 2024-09-18
create ssl cert per vhost, not one monster

evgeni:


puppet/modules/web/manifests/init.pp
```diff
@@ -15,37 +15,6 @@
     letsencrypt => $https,
```
ekohl, line 15: I'd prefer to drop this parameter and move the inclusion into the vhost:
https://github.com/theforeman/foreman-infra/blob/cc1300e15cbb07e3b94ded6aaa53f0fbe77132fe/puppet/modules/web/manifests/base.pp#L8-L10

puppet/modules/web/manifests/vhost.pp
```diff
@@ -52,6 +52,12 @@
   }
 
   if $web::https {
```
ekohl, line 54: As noted above, let's include the prerequisite here
```suggestion
  if $web::https {
    include web::letsencrypt
```

ekohl (changes requested):
tl;dr: :+1: on the concept, some implementation notes inline.

I recently read there's some benefit in having all names on a certificate so browsers can reuse connections, but given we're mostly behind a CDN I think the benefit of that is really limited and the operational simplicity of one cert per vhost is way more important.

In the installer we use [datacat](https://github.com/richardc/puppet-datacat) to gather all values so you could do it that way per host, but it's complex. Also, the module is unmaintained and [collections](https://github.com/voxpupuli/puppet-collections) is a maintained alternative.

puppet/modules/web/manifests/init.pp
evgeni: That will break using `web::base` with HTTPS directly, like it's done by `redmine` and `profiles::jenkins::controller`, as those do not use `web::vhost`.

ehelms:
This solution works for me as an approach. Thanks for taking a look.

puppet/modules/web/manifests/init.pp
ekohl: Can we modify those to also include the base? I think b31f6de8158d629c1ba7751a84a07b2db5650789 should work and makes it all more consistent.

puppet/modules/web/manifests/init.pp
```diff
@@ -15,37 +15,6 @@
...
   }
 
-  if $https {
-    $letsencypt_domain = 'theforeman.org'
```
ekohl, line 19: Looks like there was a typo here, but at least it was used consistently.

evgeni:
finally updated!

ehelms (approved):
LGTM, but @ekohl should give the final stamp of approval

puppet/modules/web/manifests/jenkins.pp
ekohl, line 1: Looks like we could migrate this to using `web::vhost` as well after this change.

https://webmasters.stackexchange.com/questions/97005/setting-x-forwarded-proto-under-apache-2-4 suggests you can also use `RequestHeader set X-Forwarded-Proto expr=%{REQUEST_SCHEME}` to make it dynamic, meaning it's going to be the same directive for HTTP and HTTPS.

I can submit a follow up PR to do so.

evgeni: please :)

ekohl: https://github.com/theforeman/foreman-infra/pull/2144

PR #2118, opened 2024-08-16
Add rpm repo profile

ehelms:
This is to provide the code to match the structure -- https://github.com/theforeman/foreman-infra/issues/1937

PR #2130, opened 2024-09-18
Split common puppet data into common and internal

ehelms:
This implements a discussion: https://github.com/theforeman/foreman-infra/pull/2118#discussion_r1758826044

The intent is to split what was in `common.yaml` into a file of data that changes rarely and data that tends to change more often.


ehelms:
@ekohl when you get a chance, here is the split we discussed

ekohl (approved):
Note to self: https://github.com/theforeman/theforeman-rel-eng/blob/63830d794ec7f1dca482648fbe645fb210074eef/procedures/foreman/release.md.erb#L149C43-L149C55 remains correct.

PR #2132, opened 2024-09-19
explicitly resolve A/AAAA when pulling external IP from icanhazip

evgeni:
ipv4.icanhazip.com has an AAAA record, and ipv6 has an A record, which
leads to both facts returning v6 addresses.

while this is a missconfiguration on the side of icanhazip.com, we can
prevent it from happening by forcing things at the resolver level.

### DIFF
--- a/puppet/modules/profiles/lib/facter/external_ips.rb
+++ b/puppet/modules/profiles/lib/facter/external_ips.rb
@@ -1,9 +1,15 @@
-require 'open-uri'
+require 'net/http'
+require 'resolv'
 
 Facter.add(:external_ip4) do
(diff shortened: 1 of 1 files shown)


ekohl (approved):
> ipv4.icanhazip.com has an AAAA record

That sounds like it's against the entire point of the service

evgeni:
> > ipv4.icanhazip.com has an AAAA record
> 
> That sounds like it's against the entire point of the service

Yepp. It's a miss-config on the Cloudflare side. I've poked @major (he used to run this service) as I don't know who to poke at CF (who runs the service now).

PR #2141, opened 2024-10-23
add bookworm to deb.tfm.o header

evgeni:


puppet/modules/freight/templates/deb-HEADER.html.epp
```diff
@@ -21,7 +23,7 @@ deb http://deb.theforeman.org/ plugins nightly
 
 <pre>
 wget https://deb.theforeman.org/foreman.asc -O /etc/apt/trusted.gpg.d/foreman.asc
-echo "deb http://deb.theforeman.org/ bullseye <%= $stable %>" > /etc/apt/sources.list.d/foreman.list
+echo "deb http://deb.theforeman.org/ bookworm <%= $stable %>" > /etc/apt/sources.list.d/foreman.list
```
ekohl, line 26: Stable is now 3.12 so we're officially going with it?

ekohl (approved):

puppet/modules/freight/templates/deb-HEADER.html.epp
evgeni: I'm absolutely cool saying 3.12 officially supports Bookworm.
I'd just not retrospectively deprecate Bullseye in the same statement.

ekohl: I was thinking about adding a deprecation that it will be dropped with 3.14 so users can plan ahead.

PR #2144, opened 2024-10-29
Use web::vhost in web::jenkins

ekohl:
Now that `web::vhost` handles the certificate properly, `web::jenkins` can take advantage of it. It needs some special handling to make the HTTP vhost redirect if HTTPS is enabled and that's why `http_attrs` and `https_attrs` are introduced in web::vhost.

The `X-Forwarded-Proto` header is rewritten to use the `REQUEST_SCHEME` variable to automatically get the correct value.


puppet/modules/web/manifests/jenkins.pp
```diff
@@ -1,25 +1,17 @@
...
-      webroot_paths => [$webroot],
-    }
+  $proxy_attrs = {
+    'allow_encoded_slashes' => 'nodecode',
+    'proxy_pass' = {
+      'path'          => '/',
+      'url'           => 'http://localhost:8080/',
+      'keywords'      => ['nocanon'],
+      'no_proxy_uris' => ['/.well-known'],
+    },
```
evgeni, line 14:
```
Could not parse for environment *root*: Syntax error at ','
```

puppet/modules/web/manifests/jenkins.pp
```diff
@@ -1,25 +1,17 @@
-class web::jenkins(
+class web::jenkins (
   Stdlib::Fqdn $hostname = 'ci.theforeman.org',
-  Stdlib::Absolutepath $webroot = '/var/www/vhosts/jenkins/htdocs',
   Boolean $https = false,
 ) {
   include web::base
```
evgeni, line 5: `web::vhost` requires `web` which includes `web::base`

ekohl: I kept it in because we use `$apache::user` and `$apache::group` as docroot owners. I'm not sure we actually need that though.

puppet/modules/web/manifests/jenkins.pp
```diff
@@ -29,50 +21,26 @@
...
   }
 
-  file { dirname($webroot):
-    ensure => directory,
-    owner  => 'root',
-    group  => 'root',
-    mode   => '0755',
-  }
-
   if $https {
```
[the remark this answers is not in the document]
ekohl: That would be more consistent with `web::vhost::web`. Further looking at that, we can probably also align those 2 on redirecting HTTP to HTTPS. I'll look a bit further.

puppet/modules/web/manifests/jenkins.pp
ekohl: I still miss something in my vim config from when I moved to my new laptop.

puppet/modules/web/manifests/jenkins.pp
ekohl: Dropped the docroot ownership.

PR #2149, opened 2024-11-15
clean web profile

evgeni:
- **add backend servername to web vhost**
- **add backend name to downloads vhost**
- **add dedicated website profile for web + downloads**

### DIFF
--- a/puppet/modules/web/manifests/vhost/web.pp
+++ b/puppet/modules/web/manifests/vhost/web.pp
@@ -66,8 +66,8 @@ class web::vhost::web(
   ]
 
   web::vhost { 'web':
-    servername    => 'theforeman.org',
-    serveraliases => ['www.theforeman.org'],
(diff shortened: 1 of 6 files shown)


puppet/modules/web/manifests/vhost/web.pp
```diff
@@ -66,8 +66,8 @@
   ]
 
   web::vhost { 'web':
-    servername    => 'theforeman.org',
-    serveraliases => ['www.theforeman.org'],
+    servername    => "web-backend.${facts['networking']['fqdn']}",
+    serveraliases => ['theforeman.org', 'www.theforeman.org'],
```
evgeni, line 70: this is the same logic I did in #2124, with the long-term hope to move this thing to the CDN and not have those aliases pointing at the actual web machine anymore.

ekohl (approved):

PR #2156, opened 2024-11-19
add web service to fastly setup

evgeni:
- **allow setting domains manually**
- **allow configuring healthcheck path**
- **add web service to fastly setup**

### DIFF
--- a/ansible/fastly.yml
+++ b/ansible/fastly.yml
@@ -19,11 +19,11 @@
             ssl_cert_hostname: "{{ item.backend }}.theforeman.org"
             ssl_sni_hostname: "{{ item.backend }}.theforeman.org"
             override_host: "{{ item.backend }}.theforeman.org"
-            healthcheck: HEADER.html
+            healthcheck: "{{ item.healthcheck | default('HEADER.html') }}"
         healthchecks:
-          - name: HEADER.html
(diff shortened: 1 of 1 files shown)


ansible/fastly.yml
```diff
@@ -10,8 +10,7 @@
       fastly_service:
         name: "{{ item.service }}.theforeman.org"
         fastly_api_key: "{{ fastly_api_key }}"
-        domains:
-          - name: "{{ item.service }}.theforeman.org"
+        domains: "{{ item.domains | default([{'name': item.service ~ '.theforeman.org'}]) }}"
```
ekohl, line 13: What's `~` here?

evgeni: String concat.

evgeni: https://jinja.palletsprojects.com/en/stable/templates/#other-operators

ekohl: I guess that and tried to look for that but somehow I couldn't find it.

ekohl (approved):

ansible/fastly.yml
evgeni: documenting things properly is complicated, as you know :)

PR #2158, opened 2024-11-26
fix up Location headers to use the frontend name

evgeni:


evgeni:
For web this is live already, the others are less affected as there we have fewer places (=none) that issue redirects

PR #2162, opened 2024-11-28
flip web to website01

evgeni:


evgeni:
needs to happen together with https://github.com/theforeman/jenkins-jobs/pull/510

ekohl:
> needs to happen together with [theforeman/jenkins-jobs#510](https://github.com/theforeman/jenkins-jobs/pull/510)

We can modify the deploy job to deploy to both places for some time.

evgeni:
> > needs to happen together with [theforeman/jenkins-jobs#510](https://github.com/theforeman/jenkins-jobs/pull/510)
> 
> We can modify the deploy job to deploy to both places for some time.

Sure. But do we have to? :)

ekohl:
I don't know how long the migration will be. We don't have that many changes to the website and we're also changing the hostname so there's no DNS TTLs so it's probably OK not to.

PR #2166, opened 2024-11-28
deploy fastly-purge scripts everywhere where needed

evgeni:


puppet/modules/fastly_purge/manifests/init.pp
```diff
@@ -0,0 +1,20 @@
+# @summary scripts to manage fastly CDN purging
+#
+class fastly_purge(
+) {
```
ekohl, line 4: I tend to simplify it if there are no parameters
```suggestion
class fastly_purge {
```

puppet/modules/fastly_purge/manifests/init.pp
```diff
@@ -0,0 +1,20 @@
...
+  file { '/usr/local/bin/fastly-purge':
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0755',
+    content => file('web/fastly-purge.sh'),
```
ekohl, line 10: CI is unhappy about this and the other file. You can make it dynamic:
```suggestion
    content => file("${module_name}/fastly-purge.sh"),
```

Or hardcoded:
```suggestion
    content => file('fastly_purge/fastly-purge.sh'),
```

ekohl (changes requested):

puppet/modules/fastly_purge/manifests/init.pp
evgeni: right, that's what you get when looking at "wth did I break between meetings" :) (fixed)

ekohl (approved):

PR #2168, opened 2024-12-06
move deb to repo-deb01 for fastly

evgeni:


ekohl (approved):
I logged into repo-deb01 and noticed it had fail2ban set up, but no other firewall rules. Do we need to?

evgeni:
Oh yeah, I think the default OSUOSL image has it installed.

I personally like f2b, but don't think we absolutely need it.

PR #2172, opened 2024-12-09
use backend naming for rpm/stagingrpm

evgeni:


ekohl (commented):
Does it need a `serveralias` or will you immediately change Fastly too?

evgeni:
There is no fastly setup yet, so nothing to change

ekohl (approved):
Ah yes, `rpm` vs `yum`

PR #2190, opened 2024-12-19
Update NodeJS to 18.x on Ubuntu with Foreman 3.14+

ekohl:
As part of [dropping NodeJS 14](https://community.theforeman.org/t/drop-debian-11-ruby-2-7-and-nodejs-14-support-in-foreman-3-14/40503) this updates nodesource to NodeJS 18. That's also the version included in Ubuntu 24.04.

It drops the conditionals for Foreman < 3.9 since those are gone now.

### DIFF
--- a/puppet/modules/jenkins_node/files/pbuilder_F66-add-nodesource-repos
+++ b/puppet/modules/jenkins_node/files/pbuilder_F66-add-nodesource-repos
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # Nodesource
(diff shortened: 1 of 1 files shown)


evgeni (approved):

ekohl:
Do we need to rebuild all Ubuntu packages after merging this?

evgeni:
No, this shouldn't be necessary. The built assets do not differ based on the Node version they were built with (or at least should not!).

PR #2194, opened 2025-01-06
fix /etc/localtime to be a symlink

evgeni:


jameerpathan111 (approved):

evgeni:
actually, this was only needed on a couple of hosts, fixed manually

PR #2195, opened 2025-01-06
switch Jenkins to use Java 17

evgeni:


evgeni:
Jenkins upstream would support Java 21, but there is no 21 in Debian (old)stable.

puppet/modules/jenkins_node/manifests/init.pp
```diff
@@ -23,12 +23,19 @@
   include fastly_purge
 
   if $facts['os']['family'] == 'RedHat' {
-    $java_package = 'java-11-openjdk-headless'
+    $java_package = 'java-17-openjdk-headless'
 
-    package { ['java-1.8.0-openjdk', 'java-1.8.0-openjdk-headless', 'java-1.8.0-openjdk-devel']:
+    package { ['java-11-openjdk', 'java-11-openjdk-headless', 'java-11-openjdk-devel']:
```
evgeni, line 28: I was wondering if we should have `jenkins_java_version = 17` and `jenkins_java_cleanup = 11` somewhere, and just reuse that through the manifests.

puppet/modules/unattended/manifests/init.pp
```diff
@@ -4,14 +4,10 @@
       auto      => { 'reboot' => false },
       blacklist => [
         'docker-ce',
```
ekohl, line 6: Unrelated, but we can get rid of this by now. We never really adopted Docker on our CI.

ekohl (approved):

puppet/modules/unattended/manifests/init.pp
evgeni: Especially, we do not have Debian-based (general purpose) CI nodes ;)

PR #2239, opened 2025-05-07
add monitoring to discourse, foreman and puppet

evgeni:


ekohl (commented):
At which point do we add it to `profiles::base`?

evgeni:
> At which point do we add it to `profiles::base`?

A profile containing a profile? :)

But yes, I wanted to fold all that into base *after* the rollout, mostly to first being able to roll out in stages (which soon is over)

ekohl (approved):

PR #2240, opened 2025-05-07
add monitoring to jenkins controller and nodes

evgeni:


ekohl (approved):
At this point we will have monitoring on all hosts, right? So is this the point where we move it to the base profile? :)

evgeni:
not the ones in the "Base" group (virt01), but yeah, I merge this and then do the sweep tomorrow

PR #2241, opened 2025-05-08
correct the hostname match for jenkins controller

evgeni:


ekohl (approved):

evgeni:
heh, had to unassing the "controller" profile in Foreman, otherwise it would clash (and that's why we didn't see it being a problem)

PR #2242, opened 2025-05-08
move monitoring setup to base profile

evgeni:


puppet/modules/profiles/manifests/base.pp
```diff
@@ -20,4 +24,41 @@
   class { 'foreman_proxy::plugin::remote_execution::ssh_user':
     manage_user => true,
   }
+
+  # Configure Grafana-based monitoring
```
ekohl, line 28: I wonder if it makes sense to keep this in a private class to keep the base profile size limited

evgeni: you mean something like `profile::base::monitoring` or a totally dedicated `monitoring_client` module that gets included here?

ekohl: I was thinking `profiles::base::monitoring` because I'm lazy and don't want to introduce another module. If it gets larger we can always still do that.

evgeni: like this?

ekohl:
Probably

puppet/manifests/site.pp
```diff
@@ -1,4 +1,4 @@
-node default {
+node /.*\.theforeman\.org$/ {
```
evgeni, line 1: > Important: Make sure all of your node definition name regexes match non-overlapping sets of node names. If a node’s name matches more than one regex, Puppet makes no guarantee about which matching definition it will get.

https://www.puppet.com/docs/puppet/7/lang_node_definitions.html

:crying_cat_face:

evgeni:
mhh, it's intentional… https://github.com/theforeman/foreman-infra/blob/c6d3daa9ad00f8a5ab6bb986c28ddb7036539854/puppet/spec/spec_helper.rb#L27

puppet/manifests/site.pp
```diff
@@ -1,63 +1,52 @@
 node default {
-  include profiles::base
```
ekohl, line 2: If we remove the last one here, should we at least create an entry for `virt\d+` to keep it managed?

evgeni: oh absolutely, this was more meant as a test on whether it will fix the failures on all the other tests.

I do wonder tho whether we should feed the tests with a dedicated (semi-empty) `site.pp`

evgeni: okay, I dropped the manifest usage in tests and it works, so I added the base profile back to the default node

puppet/modules/profiles/manifests/base/monitoring.pp
```diff
@@ -1,4 +1,4 @@
-class profiles::monitoring::client (
+class profiles::base::monitoring (
   String[1] $prometheus_username,
   String[1] $prometheus_password,
   Stdlib::HTTPUrl $prometheus_url,
```
evgeni, line 4: these are non-optional (as otherwise there is no endpoint to report to), but are undefined as I kept them "secret", which makes the tests fail now.
1. I could provide dummy values in `internal.yaml`, but that seems like a hack.
1. I could introduce a dummy hiera file for rspec only
1. I could make these actually optional (and not deploy the `remote_write` config if they are unset)

Preferences?

evgeni: I went with 3

ekohl (approved):

PR #2244, opened 2025-05-08
drop EL7 support

evgeni:


puppet/modules/jenkins_job_builder/manifests/config.pp
```diff
@@ -14,16 +14,15 @@
   $directory = '/etc/jenkins_jobs'
   $inifile = "${directory}/jenkins_jobs_${config_name}.ini"
 
-  $git = if $facts['os']['release']['major'] == '7' { 'git' } else { 'git-core' }
-  stdlib::ensure_packages([$git])
+  stdlib::ensure_packages(['git-core'])
```
ekohl, line 17: There is also `vcsrepo::manage::git` but there's no autorequire so this is fine.

ekohl (approved):

PR #2253, opened 2025-05-12
correct permissions on /var/lib/prometheus/node-exporter

evgeni:
Fixes: a4cda8f34f5b9a8e00de5a6f32f96fe6981f86fd


ekohl (commented):
Should the directory itself have an owner and a group set instead and not rely on world readable files?

evgeni:
I can state the `prometheus` ownership explicitly here?

ekohl:
Given it's writable by the group I like that. Thanks for the additional context

ekohl (approved):

PR #2255, opened 2025-05-13
update to Puppet 8

evgeni:


evgeni:
There is currently nothing that manages the puppet/vox repos on our nodes.

Should we add something that purges `puppet7-release` from them and installs `openvox8-release` to Puppet?

PR #2259, opened 2025-05-19
don't schedule restic forget while restic backup runs

evgeni:
we run backups "daily", which means "*-*-* 00:00:00"
when we run forget "weekly", this ends up being "Mon *-*-* 00:00:00" and
two writing operations to one repository does not end up well.

instead, let's schedule forgets two hours later.

(setting RandomizedDelaySec on the timers would probably also help a
bit, but would still have a potential for a race condition, so let's
avoid it)

### DIFF
--- a/puppet/data/internal.yaml
+++ b/puppet/data/internal.yaml
@@ -35,7 +35,7 @@ restic::enable_forget: true
 restic::forget:
   keep-daily: 14
   keep-weekly: 4
-restic::forget_timer: 'weekly'
+restic::forget_timer: 'Mon *-*-* 02:00:00'
 restic::prune: true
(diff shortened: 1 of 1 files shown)


evgeni:
(found by monitoring! :) )

ekohl:
I wonder if systemd has some native functionality here ut this is a decent workaround

ekohl (approved):

evgeni:
> I wonder if systemd has some native functionality here ut this is a decent workaround

I couldn't find any from a very quick glance, and then went with this.

ekohl:
https://serverfault.com/questions/1128450/delay-systemd-timer-until-conflicting-service-has-finished also suggests there isn't such a thing. We could use `Conflicts` to prevent the conflict, but then nothing guarantees it still runs later.

PR #2260, opened 2025-05-27
auto-reboot EL-based jenkins nodes when needed

evgeni:


puppet/modules/jenkins_node/manifests/init.pp
```diff
@@ -137,4 +137,22 @@
   file { '/etc/cron.daily/jenkins_cleaner':
     ensure => absent,
   }
+
+  file { '/usr/local/sbin/reboot-jenkins-node':
+    ensure => file,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0755',
+    source => 'puppet:///modules/jenkins_node/reboot-jenkins-node',
```
ekohl, line 146: I tend to inline the content to avoid the fileserver. For small scripts this is faster and in theory could could use a cached catalog to reapply, even if we really don't use this. The biggest win is really that it will fail on catalog compilation if the file doesn't exist so your unit tests can capture it.
```suggestion
    content => file("${module_name}/reboot-jenkins-node"),
```

puppet/modules/jenkins_node/manifests/init.pp
```diff
@@ -137,4 +137,22 @@
   file { '/etc/cron.daily/jenkins_cleaner':
     ensure => absent,
   }
+
+  file { '/usr/local/sbin/reboot-jenkins-node':
```
ekohl, line 141: Do you intend to use this on Debian too or should this be inside the `if` block?

ekohl (approved):
I like the idea. The alternative would be to have some systemd timer/cron that checks if a reboot is needed and if so if the node is also idle. Then there's some retry mechanism. But this is simple enough.

puppet/modules/jenkins_node/manifests/init.pp
evgeni: I certainly want to have a look at that, but probably not today.

PR #2266, opened 2025-06-13
Add pulpcore-release.rpm symlink

ehelms:


ekohl (changes requested):
It's `pulpcore-repo`, singular: https://github.com/theforeman/pulpcore-packaging/tree/rpm/develop/packages/pulpcore-repo.

I'd have preferred it the original package was called `pulpcore-release` since that's the convention. `katello-repos` always bothered me.

@Odilhao can we perhaps correct this before we start relying on it?

ekohl (changes requested):
I didn't notice this in the first review.

ekohl (approved):
Thanks!

PR #2268, opened 2025-07-10
Route Redmine through Anubis

evgeni:


puppet/modules/redmine/manifests/init.pp
```diff
@@ -166,16 +166,24 @@
     content => template('redmine/redmine.service.erb'),
   }
 
+  anubis::instance { 'redmine':
```
evgeni, line 169: technically, this should be in the remine *profile* and the httpd config should also be there, but I didn't plan on a big refactoring today.

evgeni: or maybe, this IS the redmine profile? :)

evgeni:
https://anubis.techaro.lol/docs/design/how-anubis-works - if there is no "Mozilla" in the User-Agent, it will let things through.

evgeni:
https://projects.theforeman.org/issues/38565 was properly updated by PRProcessor after this was rolled out

PR #2262, opened 2025-05-30
Update to node 22

MariaAga:
https://github.com/theforeman/foreman/pull/10556

PR #2271, opened 2025-07-18
Update Debian/Ubuntu to Node 22 for 3.16+

adamruzicka:
https://github.com/theforeman/foreman-infra/pull/2262 with review comments addressed

### DIFF
--- a/puppet/modules/jenkins_node/manifests/packaging/debian.pp
+++ b/puppet/modules/jenkins_node/manifests/packaging/debian.pp
@@ -31,7 +31,7 @@ class jenkins_node::packaging::debian(
       release    => 'bookworm',
       apturl     => $debian_mirror,
       aptcontent => "deb ${debian_mirror} bookworm main non-free contrib\ndeb-src ${debian_mirror} bookworm main non-free contrib\n",
-      nodesource => false;
+      nodesource => true;
     'jammy64':
       ensure     => present,
       arch       => 'amd64',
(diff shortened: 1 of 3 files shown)


puppet/spec/classes/jenkins_node_spec.rb
```diff
@@ -31,7 +31,7 @@
           it { is_expected.to contain_file('/etc/pbuilder/bullseye64/hooks/F66-add-nodesource-nodistro-repos').with_ensure('absent') }
           it { is_expected.to contain_file('/etc/pbuilder/jammy64/hooks/F66-add-nodesource-repos').with_ensure('absent') }
           it { is_expected.to contain_file('/etc/pbuilder/jammy64/hooks/F66-add-nodesource-nodistro-repos').with_ensure('present') }
-          it { is_expected.to contain_file('/etc/pbuilder/bookworm64/hooks/F66-add-nodesource-repos').with_ensure('absent') }
+          it { is_expected.to contain_file('/etc/pbuilder/bookworm64/hooks/F66-add-nodesource-repos').with_ensure('present') }
           it { is_expected.to contain_file('/etc/pbuilder/bookworm64/hooks/F66-add-nodesource-nodistro-repos').with_ensure('absent') }
```
evgeni, line 35: I would have expected the "nodistro" one to be present now.
Looking at https://github.com/theforeman/foreman-infra/blob/1485c8265e29408081a9d69272550d7cec0cd7de/puppet/modules/jenkins_node/manifests/pbuilder_setup.pp#L44-L45 we need to adjust the "release" part there too :/
(the other repo is not updated anymore and we shouldn't be using it at all for current releases -- we should clean that up)

evgeni:
Should be merged *after* https://github.com/theforeman/foreman/pull/10556

puppet/spec/classes/jenkins_node_spec.rb
evgeni: the Debian cleanup: https://github.com/theforeman/foreman-infra/pull/2275

puppet/modules/jenkins_node/manifests/packaging/debian.pp
```diff
@@ -37,7 +37,7 @@
       release    => 'bookworm',
       apturl     => $debian_mirror,
       aptcontent => "deb ${debian_mirror} bookworm main non-free contrib\ndeb-src ${debian_mirror} bullseye main non-free contrib\n",
-      nodesource => false;
+      nodesource => true;
```
evgeni, line 40: technically, it's `true` by default, so we could even drop that line, on the other hand… better explicit?

adamruzicka: I'd rather be explicit here

PR #2275, opened 2025-07-21
Debian cleanup

evgeni:


evgeni:
Focal was last present in 3.12,
Bullseye was last present in 3.13.
Both are EOL by now.

puppet/modules/jenkins_node/manifests/packaging/debian.pp
```diff
@@ -26,7 +26,7 @@
 
   jenkins_node::pbuilder_setup {
     'bullseye64':
-      ensure     => present,
+      ensure     => absent,
```
adamruzicka, line 29: Why?

PR #2305, opened 2025-09-17
add module to configure openvox repo

evgeni:
also purges puppet repo

### DIFF
--- a/puppet/modules/openvox_repo/manifests/apt.pp
+++ b/puppet/modules/openvox_repo/manifests/apt.pp
@@ -1,0 +1,17 @@
+class openvox_repo::apt() {
+  include apt
+
+  $os_name = downcase($facts['os']['name'])
+  $release = "${os_name}${facts['os']['release']['major']}"
+
+  apt::source { 'openvox':
+    comment  => "OpenVox ${openvox_repo::release} ${release} Repository",
+    location => 'https://apt.voxpupuli.org',
+    release  => $release,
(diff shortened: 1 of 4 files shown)


puppet/modules/openvox_repo/manifests/apt.pp
```diff
@@ -0,0 +1,17 @@
+class openvox_repo::apt() {
+  include apt
+
+  $os_name = downcase($facts['os']['name'])
+  $release = "${os_name}${facts['os']['release']['major']}"
+
+  apt::source { "openvox":
+    comment  => "OpenVox ${openvox_repo::release} ${release} Repository",
+    location => 'https://apt.voxpupuli.org',
```
bastelfreak, line 9: I think the apt module will log a warning to your server log if that's a string and not an array (don't ask me why)

puppet/modules/openvox_repo/manifests/purge_puppet.pp
```diff
@@ -0,0 +1,5 @@
+class openvox_repo::purge_puppet() {
+  package {['puppet7-release', 'puppet8-release']:
```
bastelfreak, line 2: any chance you also had `puppet-release` and want to purge that as well?

evgeni: I am rather sure we have not, but it's cheap to purge that too.

PR #2312, opened 2025-10-06
Switch to voxpupuli-puppet-lint-plugins

bastelfreak:
Vox Pupuli provides a curated list of puppet-lint plugins with proper versioning. All thoses plugins are pulled in by a meta gem, voxpupuli-puppet-lint-plugins. They are identical with the plugin list Foreman used previously.


bastelfreak:
I think CI fails because it's pulling in the puppet 7 gem and openvox 8 :thinking:

PR #2318, opened 2025-10-08
Update documentation to use Openvox not Puppet

evgeni:
Fixes: #2308


Vagrantfile
```diff
@@ -12,8 +12,8 @@ Vagrant.configure("2") do |config|
   config.vm.provision "install puppet", type: "shell", inline: <<-SHELL
```
evgeni, line 12:
```suggestion
  config.vm.provision "install openvox-agent", type: "shell", inline: <<-SHELL
```

Vagrantfile
```diff
@@ -52,10 +52,10 @@ Vagrant.configure("2") do |config|
     end
     override.vm.provision "install puppet", type: "shell", inline: <<-SHELL
```
evgeni, line 53:
```suggestion
    override.vm.provision "install openvox-agent", type: "shell", inline: <<-SHELL
```

PR #2320, opened 2025-10-08
Move core types from git submodules->Puppetfile

bastelfreak:
mailalias_core and mount_providers were already in the Puppetfile and IMO it makes sense to have all of them in the same place.


ekohl (changes requested):
:-1: on this because it means we'll ship modules to our clients. `mailalias_core` is in there because it's no longer part of the AIO builds. I don't think `mount_providers` ever was in the AIO builds.

PR #2406, opened 2026-04-29
set debbuildopts to an empty string to fix building on trixie

evgeni:
the default is -b, which means "binary only", which worked fine until trixie, but now fails as dpkg-genchanges fails to find the .dsc file which is only generated when doing a sourceful build. doing so is not too expensive and I don't want to debug pdebuild any longer.


devendra104 (approved):
ack.

PR #2431, opened 2026-06-10
Switch Debian builders to use OpenVox instead of Puppet

evgeni:


puppet/modules/jenkins_node/files/pbuilder_F67-add-openvox-repos
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+echo "deb https://apt.voxpupuli.org/ ${DISTRIBUTION} openvox8" >> /etc/apt/sources.list
```
evgeni: This won't work, as Vox doesn't use names for their repos, just numbers.

evgeni:
This will require https://github.com/theforeman/foreman-packaging/pull/11839 to work properly. But our "setup repos" scripts are Foreman version agnostic, and would also apply when building 3.19 and older.

Question: Do we want to backport the packaging change (should be safe, but… you know how things go) or introduce a if/else in the hook that sets up Puppet for 3.19 and older and OpenVox for 5.0+ (we did similar things for NodeJS in the past)?

ekohl:
> Question: Do we want to backport the packaging change (should be safe, but… you know how things go) or introduce a if/else in the hook that sets up Puppet for 3.19 and older and OpenVox for 5.0+ (we did similar things for NodeJS in the past)?

I consider it an oversight for 3.19 so I'd be OK with backporting. Perhaps aim for consistency with RPM builds?

evgeni:
As we're still building 3.18.z, it would also need to go there to keep builds working.
But https://github.com/theforeman/foreman-packaging/pull/12533 landed in 3.17, so you could say backporting that to 3.18 is also fair game.

ogajduse:
OpenVox was backported to `deb/3.18` and `deb/3.19` are we ready to undraft this PR?

- deb/3.19: https://github.com/theforeman/foreman-packaging/pull/13618
- deb/3.18: https://github.com/theforeman/foreman-packaging/pull/13619

evgeni:
*Ruby Rhod voice* Super ready!

ogajduse (approved):

PR #2447, opened 2026-07-08
Remove the excludepkg for openssl as it prevents downgrade

ehelms:
As the exclude happens before the downgrade, it appears to exclude the "current" install version from analysis by dnf and thus the downgrade doesn't happen. Since we manage our nodes with puppet, the ensure of a specific version should be enough.


ehelms:
merging to unblock CI and get this applied

PR #2474, opened 2026-09-10
Update puppetlabs/java to 12.1.0

evgeni:
12.0.0 contains the Debian 13 PR we were waiting on


evgeni:
Bleh, puppet/jenkins needs to support that :/

